PR #26, opened 2018-11-09
Switch to NPM packages for some deps

bryphe:


bryphe:
This was addressed with #48

PR #46, opened 2018-11-15
update gitignore

joprice:


bryphe:
Thanks @joprice ! Looks great! Build failures are unrelated to this change, so I'll bring it in now -

PR #55, opened 2018-11-18
fix crash in example with empty string

joprice:
When the input string is empty and backspace is hit, you get the error: 

`Fatal error: exception Invalid_argument("String.sub / Bytes.sub")`

This is due negative index passed to substring.


bryphe:
Excellent! Thanks for fixing this @joprice .

PR #58, opened 2018-11-18
[WIP] Feature/add flex styles

Akin909:
Add an example using to show usage of various flex box arrangements


src/UI/Style.re
```diff
@@ -5,67 +5,75 @@ open Revery_Core;
 type fontFamily = string;
 
 type t = {
-    backgroundColor: Color.t,
```
Akin909: got `ocaml-reasonl-lsp` working (none of the other plugins work) but it's using `reasonfmt` resulting in the unintended whitespace changes

examples/Flexbox.re
```diff
@@ -0,0 +1,142 @@
...
+  let parentColumnStyle = parentStyles(~direction=LayoutTypes.Column);
+  let headerStyle =
+    Style.make(
+      ~fontSize=60,
+      ~fontFamily="Roboto-Regular.ttf",
+      ~color=Colors.white,
+      (),
+    );
+
+  let horizontalStyles =
```
bryphe: @Akin909 - this is awesome! Great to see all these cases working 💯 Thanks for putting this together.

src/UI/Style.re
```diff
@@ -5,67 +5,75 @@ open Revery_Core;
...
-    right: int,
-    fontFamily: fontFamily,
-    fontSize: int,
+  backgroundColor: Color.t,
+  color: Color.t,
+  width: int,
+  height: int,
+  position: LayoutTypes.positionType,
+  flexDirection: LayoutTypes.flexDirection,
+  alignItems: LayoutTypes.align,
```
bryphe: Thanks for adding these new styles 👍

Akin909:
@bryphe I was hoping to add some more rules like `order`etc. but still having a few issues with my tool/environment re lsp and such so thought rather than sit on this you could use what I've got so far and I can revisit this when I'm a bit better setup. In the mean time I've tidied it up a bit

Akin909:
@bryphe I actually seem to have gotten `ocaml-language-server` to work with terminal nvim although the steps I took were a bit random, I installed the `reason-cli@latest-linux` then also installed `opam` and `ocaml` separately. It didn't work without all those bits (on my system anyway, arch-based linux distro)

PR #83, opened 2018-12-01
Add Clock demo

OhadRau:
I worked to fine-tune this example a little bit after I switched to using <view>s. I still think it would be cleaner to have a real anchor point/pivot thing supported in Revery (right now I'm doing some weird math and it's still probably not perfect), so I'm gonna try to design an API for that (probably will submit that PR in a few days).


OhadRau:
Oh, didn't realize you already have it written lol

PR #88, opened 2018-12-05
Update esy.lock and remove duplicated dependency

jchavarri:
- Removes one instance of `reason-fontkit` from package.json (there was one pointing to 1.1.0 and another to 1.0.0)
- Updates esy.lock to the latest version format


bryphe:
> Removes one instance of reason-fontkit from package.json (there was one pointing to 1.1.0 and another to 1.0.0)

Ah nice catch! Thanks for cleaning this up @jchavarri , looks great 👍

PR #100, opened 2018-12-07
Test Runner: Switch to 'rely'

bryphe:


test/dune
```diff
@@ -1,7 +1,6 @@
 (executable
     (name TestRunner)
```
kyldvs: This is the exact dune file we use for Rely in our internal repo. I'm not sure if any of these `ocamlopt_flags` will somehow fix the warning 9?

(all the memcpy stuff should be unrelated to this, but I pasted the dune file verbatim just in case)

```
(executable
  (package www)
  (name TestsExe)
  (public_name TestsExe.exe)
  (ocamlopt_flags (-g -ccopt -Wl,--wrap=memcpy -ccopt -Wl,--strip-debug -ccopt -Wl,memcpy-fix/memcpy.o))
  (libraries
    MemcpyFixAllowsPrebuiltBinariesInWWW
    Tests
  )
)
```

jordwalke: And the `--strip-debug` flag is one of the things required to get reproducible binaries IIRC (only useful if checking in binaries into the repo).

Et7f3:
Merged in #602.

Issue #103, opened 2018-12-07
API - Colors: Implement full set of colors available from HTML


bryphe:
We have a very small subset of Colors defined in our `Colors` module - like `Colors.red`, `Colors.blue`, etc.

It'd be great to have parity with the default set in HTML - I think this is a comprehensive list here:
https://www.w3schools.com/colors/colors_names.asp

Might be easier with #102 tackled - but it would be awesome to have this full set of Colors in `revery`, too!

Good place to start looking: 
https://github.com/bryphe/revery/blob/master/src/Core/Colors.re

bgoscinski:
I'm on it :)

bryphe:
Sweet, thank you @bgoscinski 😄  Let me know if you need any help getting started!

PR #106, opened 2018-12-07
feat: implement full set of named CSS colors

bgoscinski:
implements #103. I have some problems with `esy install` so I could not test this :(


bryphe:
Woah! Awesome, thank you @bgoscinski !

> I have some problems with esy install so I could not test this :(

CI will do this for us - but let me know if you want help troubleshooting the `esy install` issues. Feel free to open an issue here w/ the error and I'll see if I can help 👍

bryphe:
Thanks for the details @bgoscinski ! Cool to see another Windows developer using Reason - I'm also on Win10.

I opened an issue on the `esy` repo (it does seem more general than `revery`).

> In process manager I found out that esyCommand.exe spawns curl from C:\Windows\System32. Am I right that it should spawn the one which comes with cygwin?

Nice catch! I suspect this is the issue. Indeed when `esy` runs Curl, it's expecting the cygwin version, which has different semantics than the bundled Windows one.

bgoscinski:
woah! You're fast :D

bryphe:
Haha there's actually been a few other issues like this - we had the same problem with `git`, for example: https://github.com/esy/esy/pull/556/files

Thanks for the details / investigation!

PR #107, opened 2018-12-08
Add box shadow property

Akin909:
boxshadow

Updated:

box-shadow

Outstanding
===

- [x] spread radius
- [x] blur radius

### DIFF
--- a/src/UI/GradientShader.re
+++ b/src/UI/GradientShader.re
@@ -1,7 +1,7 @@
 /**
  * Gradient Shader
  *
- * shader which allows for specificying a gradient.
+ * shader used for specificying a gradient.
  * the gradient is used to draw a box shadow
  */
 open Revery_Shaders;
--- a/src/UI/GradientShader.re
+++ b/src/UI/GradientShader.re
@@ -16,12 +16,22 @@ let attributes: list(ShaderAttribute.t) =
     },
   ];
 
-let uniform: list(ShaderUniform.t) =
+let uniforms: list(ShaderUniform.t) =
   SolidShader.uniform
   @ [
     {
       dataType: ShaderDataType.Sampler2D,
       name: "uSampler",
+      usage: FragmentShader,
+    },
+    {
+      dataType: ShaderDataType.Vector3,
+      name: "uShadowColor",
+      usage: FragmentShader,
+    },
+    {
+      dataType: ShaderDataType.Vector2,
+      name: "uShadowAmount",
       usage: FragmentShader,
     },
   ];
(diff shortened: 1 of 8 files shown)


bryphe:
Thanks for the screenshot, @Akin909 ! Awesome progress! 💯

bryphe:
@Akin909 - awesome! Updated screenshot looks so cool 👍 Just what we need for oni v2 😄

src/UI/GradientShader.re
```diff
@@ -0,0 +1,93 @@
...
+|};
+
+let create = (~blur, ~height, ~width, ~color) => {
+  let shader =
+    Shader.create(
+      ~attributes,
+      ~varying,
+      ~uniforms=uniform,
+      ~vertexShader=vsShader,
+      ~fragmentShader=createFragmentShader(~blur, ~height, ~width, ~color),
```
bryphe: I'd recommend here, instead of creating a new shader when parameters change, for us to add a `uniform` shader value that we can set. Shader compilation is actually pretty expensive!

We could do the following:

__Step 1:__
Add to the `uniform` block here:
```
let uniform: list(ShaderUniform.t) =
  SolidShader.uniform
  @ [
    {
      dataType: ShaderDataType.Sampler2D,
      name: "uSampler",
      usage: FragmentShader,
    },
    {
      dataType: ShaderDataType.Vector3,
      name: "uShadowColor",
      usage: FragmentShader,
    },
    {
      dataType: ShaderDataType.Vector2,
      name: "uShadowAmount",
      usage: FragmentShader,
    },
  ];
```

These `uniform` values can be then set externally, without needing to recompile the shader.

src/UI/GradientShader.re
```diff
@@ -0,0 +1,93 @@
...
+    ++ "float amountY = "
+    ++ amountY
+    ++ ";\n";
+  amountStr ++ colorStr;
+};
+
+let createFragmentShader = (~blur, ~height, ~width, ~color) =>
+  updateShader(~blur, ~height, ~width, ~color)
+  ++ {|
+  float leftEdgeAmount = smoothstep(0.0, amountX, vTexCoord.x);
```
bryphe: If we go the `uniform` route described in the other comment - we'd just need to change the `amountX`/`amountY` to `uShadowAmount.x` and `uShadowAmount.y`

src/UI/ViewNode.re
```diff
@@ -218,6 +218,48 @@ let renderBorders =
...
+    gradientShader,
+    "uProjection",
+    m,
+  );
+
+  Shaders.CompiledShader.setUniformMatrix4fv(
+    gradientShader,
+    "uWorld",
+    shadowWorldTransform,
+  );
```
bryphe: Lastly, if go the uniform route, we just need to set the uniforms here!

- `Shaders.CompiledShader.setUniform3f(gradientShader, "uShadowColor", Color.toVec3(color))`
- `Shaders.CompiledShader.setUniform2f(gradientShader, "uShadowAmount", Vec2.create(width, height))`

Akin909: One difference is that `uShadowAmount` isn't the same as `amountX` as that is `blurRadius/width` (although I tweaked that and it didn't help)

bryphe: First step I'd do is, in the shader, just use hardcoded values for `uShadowColor` and `uShadowAmount` - verify the logic is working as expected. If those hardcoded values work - it must be something with the interop, or something wrong with the values passed in.

bryphe: If you push up what you have - I can help you take a look too

> didn't realise the cost to doing it this way, I've tried tweaking it and now nothing renders though, must be a silly mistake but since there's no compilation error its hard to narrow it down.

I found out the hard way! The first builds of Revery were super slow because I was recompiling them every frame 😆

Akin909: Fixed it :+1: (it was silly :man_facepalming:, I had the `Shader.use` line too low)

bryphe: Awesome! Glad you found the issue 👍

PR #111, opened 2018-12-15
Wrap Glfw.Key

jchavarri:
I saw [this todo](https://github.com/bryphe/revery/blob/a6f795bcbc7ba741cc7df3b6f775c60f32e63a5c/examples/Autocomplete.re#L135) and thought about this proposal to alias the `Key` events from glfw from revery `Core`. Maybe `Key` should be moved to a different module like `Ui` 🤔 .

I explored an approach for the converter using `%identity` but it'd lose type safety, so finally went with the manual approach. I think the compiler should optimize the `convert` function, BuckleScript for example compiles a converter of this kind down [to the identity function](https://reasonml.github.io/en/try?rrjsx=true&ocaml=C4TwDgpgBMULwCgpQD5QPIDsJNVAKgO4D2ua+AFgE4Q6iRQCu8ZUAqpqwCISMAerfFWIBLAM4IANhFgBjYpgBuEKrGhwoAWwCGwWRSjRCI4AbRZoUALQA+dpjxFiyW1B79H1Wi7tDRYoA).

@bryphe Let me know what you think! 🙂


src/Core/Key.re
```diff
@@ -0,0 +1,249 @@
...
+  | KEY_LEFT_CONTROL
+  | KEY_LEFT_ALT
+  | KEY_LEFT_SUPER
+  | KEY_RIGHT_SHIFT
+  | KEY_RIGHT_CONTROL
+  | KEY_RIGHT_ALT
+  | KEY_RIGHT_SUPER
+  | KEY_MENU;
+
+let convert = glEvent =>
```
bryphe: WOW! Thank you so much for doing this.

src/Core/Key.re
```diff
@@ -0,0 +1,249 @@
+open Reglfw.Glfw.Key;
+
+type t =
```
bryphe: This looks great to me. Exactly where I saw us going with the API. Sets us up well to start supporting first-class input events that can be leveraged via the JSX / UI components 👍

Issue #120, opened 2018-12-19
Geometry: Implement 'triangle'


bryphe:
Today, we don't have a simple way to create a `Triangle` primitive to use in rendering the UI.

We should create a wrapper similiar to the `Quad` primitive: https://github.com/bryphe/revery/blob/master/src/Geometry/Quad.re

...but for creating Triangles. This is important for implementing border-styling in #119

Akin909:
I think this was done in #123 but the issue didn't close so closing cc @bryphe @OhadRau (in case I'm wrong)

PR #123, opened 2018-12-20
[WIP] Add border styles & triangle primitives

### LINKED ISSUE #119: UI Styles: Implement 'border' styling
To have a useful UI - we should have parity with what the browser supports for the `border` styles: <https://developer.mozilla.org/en-US/docs/Web/CSS/border>

We should decide on our API surface. Perhaps something like this?
- 2-pixel red border around entire element: `<view style={Style.make(~border=Border(Colors.red, 2), ())} />`  
- 1-pixel blue border on the left and right: `<view style={Style.make(~borderHorizontal=Border(Colors.blue, 1)} />`

For the initial implementation, only implementing the `solid` style seems reasonable. We can always add other styles down the road - supporting the `border-width` and `border-color` are important, though!

There are a few things we'll have to do to support this

__Part 1: Style properties__
- `Style.re` - Add the `border` properties to our style type: https://github.com/bryphe/revery/blob/711a90a0b19af2f39dd984cb1570a1231e745365/src/UI/Style.re#L25
- `Style.re` - Add the properties to our `make` function: https://github.com/bryphe/revery/blob/711a90a0b19af2f39dd984cb1570a1231e745365/src/UI/Style.re#L36
- `Style.re` - Add the sizes to our `toLayoutNode` function, so that layout handles it correctly: https://github.com/bryphe/revery/blob/711a90a0b19af2f39dd984cb1570a1231e745365/src/UI/Style.re#L103

__Part 2: Rendering quads__

For the core border rendering, we need to render a quad that spans the width (in the case of border-top/border-bottom) or the height (in the case of border-left or border-right), and positioned correctly in relationship to the node's layout.

This logic can live in `ViewNode`, and it's quite similiar to the logic we have today for drawing the background - the only difference really is the _color_ and _dimensions_ of the quad.

https://github.com/bryphe/revery/blob/711a90a0b19af2f39dd984cb1570a1231e745365/src/UI/ViewNode.re#L22

__Part 3: Rendering Junctions__

The trickiest piece of this work that will need to be handled with the rendering is where borders meet up. For example, if I have a top-border and a left-border, we need to render triangles potentially where these meet. This is dependent on us having a triangle primitive in #120 

For example, if I had a div like this: 
```
<div style="border-left: 10px solid red; border-top: 20px solid yellow; width: 100px; height:100px;">
```
It would look something like this:
image

Once we get Part 3 completed, though, we'll have a pretty usable `border` story 💯

OhadRau:
Will fix #119 & #120

image

### To-Do:
- [x] Fix child nodes to correctly respond to the translated world when using borders (https://github.com/jordwalke/flex/blob/master/src/lib/LayoutSupport.re#L453-L461) -- in case anyone else has a similar issue, make sure you're using `Encoding.cssUndefined` as the default value instead of `0.`
- [x] Optimize the math for triangle texture coordinates (at the very least remove the list/fold operations to find the max & unfold that to pattern matching)
- [x] Improve readability/reduce codesize for border rendering function
- [x] Make sure that it's okay to multiply to apply the translation
- [x] Disable junctions when there aren't borders being connected

### DIFF
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -23,7 +23,31 @@ let init = app => {
       (),
     );
 
-  let render = () => <view style=defaultStyle></view>;
+  let innerStyle =
+    Style.make(
+      ~backgroundColor=Colors.yellow,
+      ~position=LayoutTypes.Relative,
+      ~left=0,
+      ~top=0,
+      ~width=30,
+      ~height=30,
+      ~border=Style.Border.make(~color=Colors.black, ~width=3, ()),
+      (),
+    );
+
+  let textStyle =
+    Style.make(
+      ~fontSize=20,
+      ~fontFamily="Roboto-Regular.ttf",
+      ~color=Colors.black,
+      (),
+    );
+
+  let render = () =>
+    <view style=defaultStyle>
+      <text style=textStyle> "WELCOME TO REVERY!" </text>
+      <view style=innerStyle> </view>
+    </view>;
 
   UI.start(w, render);
 };
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -1,0 +1,31 @@
+open Revery;
+open Revery.UI;
+open Revery.Core;
+
+let init = app => {
+  let w =
+    App.createWindow(
+      app,
+      "border-example",
+    );
+
+  let defaultStyle =
+    Style.make(
+      ~backgroundColor=Colors.white,
+      ~position=LayoutTypes.Relative,
+      ~left=100,
+      ~top=100,
+      ~width=200,
+      ~height=200,
+      ~border=Style.Border.make(~width=15, ~color=Colors.red, ()),
+      ~borderHorizontal=Style.Border.make(~width=8, ~color=Colors.blue, ()),
+      ~borderLeft=Style.Border.make(~width=30, ~color=Colors.green, ()),
+      (),
+    );
+
+  let render = () => <view style=defaultStyle></view>;
+
+  UI.start(w, render);
+};
+
+App.start(init);
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -19,6 +19,7 @@ let init = app => {
       ~height=200,
       ~border=Style.Border.make(~width=15, ~color=Colors.red, ()),
       ~borderHorizontal=Style.Border.make(~width=8, ~color=Colors.blue, ()),
+      ~borderTop=Style.Border.make(~width=15, ~color=Colors.red, ()),
       ~borderLeft=Style.Border.make(~width=30, ~color=Colors.green, ()),
       (),
     );
@@ -45,8 +46,8 @@ let init = app => {
 
   let render = () =>
     <view style=defaultStyle>
-      <text style=textStyle> "WELCOME TO REVERY!" </text>
       <view style=innerStyle> </view>
+      <text style=textStyle> "WELCOME TO REVERY!" </text>
     </view>;
 
   UI.start(w, render);
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -31,7 +31,7 @@ let init = app => {
       ~top=0,
       ~width=30,
       ~height=30,
-      ~border=Style.Border.make(~color=Colors.black, ~width=3, ()),
+      ~borderHorizontal=Style.Border.make(~color=Colors.black, ~width=3, ()),
       (),
     );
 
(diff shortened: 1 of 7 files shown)


bryphe:
Wow, awesome progress @OhadRau ! 💯

examples/Border.re
```diff
@@ -0,0 +1,56 @@
...
+      ~fontSize=20,
+      ~fontFamily="Roboto-Regular.ttf",
+      ~color=Colors.black,
+      (),
+    );
+
+  let render = () =>
+    <view style=defaultStyle>
+      <view style=innerStyle> </view>
+      <text style=textStyle> "WELCOME TO REVERY!" </text>
```
bryphe: Interesting - based on this I'd expect the text to be inside the borders

src/Geometry/Triangle.re
```diff
@@ -0,0 +1,36 @@
+open Revery_Shaders.Shader;
+
+let create = (x1, y1, x2, y2, x3, y3) => {
```
bryphe: Nice! We have a `Triangle` primitive now 🎉

src/Geometry/Triangle.re
```diff
@@ -0,0 +1,36 @@
...
+  let positions = [|x1, y1, x2, y2, x3, y3|];
+
+  let ranges = [abs_float(x1 -. x2), abs_float(x1 -. x3),
```
bryphe: I see what you mean here - I suppose one option would be to explicitly pass the texture coordinates in `create`

src/UI/ViewNode.re
```diff
@@ -9,6 +9,164 @@ module LayoutTypes = Layout.LayoutTypes;
...
+      (float(axis.width), axis.color);
+    } else if (border.width !== Layout.Encoding.cssUndefined) {
+      (float(border.width), border.color);
+    } else {
+      (0., Colors.black);
+    };
+  };
+
+  open Style;
+  let (topBorderWidth, topBorderColor) =
```
bryphe: Wow, nice work handling all these cases 💯 I like that the example in your screenshot had different colors for each side so we could see this logic exercised

examples/Border.re
OhadRau: Yep, I went back and fixed that. I was setting the width to 0 by default instead of Encoding.cssUndefined as that was causing the issue.

src/Geometry/Triangle.re
OhadRau: I'm sure there's some simple math I'm just forgetting about. I saw a lot of info pointing towards barycentric coordinates but I'm not sure that's the right solution here.

examples/Border.re
bryphe: Ah, ok, cool! I realized I missed your question on the issue too, just responded there.

OhadRau:
Apart from cleaning up the border drawing function, I think I've got all the features/edge cases accounted for here. Let me know if there's anything else I should take a look at or add to this PR.

bryphe:
@OhadRau - looks great to me! I'm ready to bring this in whenever you are. Thanks again for your work on this!

I tested out transparent borders, got some fun results with that:
image

OhadRau:
> @OhadRau - looks great to me! I'm ready to bring this in whenever you are. Thanks again for your work on this!

Awesome! As far as I can tell everything is ready unless you think we should add borders to images/text nodes as well.

> I tested out transparent borders, got some fun results with that

Huh, didn't really think about that case... I just checked to see what browsers do for that with this [JSFiddle](https://jsfiddle.net/9cpknb47/) and it looks like it's pretty much the same behavior, but it might make more sense to check if both edges are going to be rendered before inserting the corner piece. Don't think that should be a difficult fix if that's the behavior we want to go for.

@bryphe

bryphe:
> Huh, didn't really think about that case... I just checked to see what browsers do for that with this JSFiddle and it looks like it's pretty much the same behavior, but it might make more sense to check if both edges are going to be rendered before inserting the corner piece. Don't think that should be a difficult fix if that's the behavior we want to go for.

Actually this is the 'right' behavior from my perspective - it lets you do this CSS trick: https://css-tricks.com/snippets/css/css-triangle 😄 Sorry it wasn't clear from my comment.

So the way we have it is perfect - I take advantage of this behavior sometimes - so I'm glad it works the way it does.

OhadRau:
Ah interesting, had no idea about that. Only other behavior that might be weird is that borders inherit the opacity style, but again looks like that's what's done in CSS as well. Assuming that's correct, I guess this PR is good to go.

bryphe:
> Assuming that's correct, I guess this PR is good to go.

Perfect - awesome contribution to the project, @OhadRau ! Thank you!

PR #133, opened 2018-12-22
[WIP] easing functions

Bretley:


bryphe:
> @bryphe this is the change so far... nothing immediately special since I've only just begun to learn ocaml/reason

Excellent! Looks like a good start.

The next steps I can see are:
- Save the `easing` function into the `animation` in the `start` method (pass it in from the `animationOptions`): https://github.com/bryphe/revery/blob/57de14c6addf1e1e1c8eae8c982d714d88f8d48e/src/UI/Animation.re#L85
- Then, we need to use the `easing` function to get the _real_ `t` value. The animation `tick` happens here:  https://github.com/bryphe/revery/blob/57de14c6addf1e1e1c8eae8c982d714d88f8d48e/src/UI/Animation.re#L61 We could either add the easing function there (like `let t = anim.easing(t);`), or inside the `getLocalTime` function.

In addition, it would be great to have a test case to make sure things are plumbed through. Some examples here: https://github.com/bryphe/revery/blob/57de14c6addf1e1e1c8eae8c982d714d88f8d48e/test/UI/AnimationTest.re#L30

(We could create a simple animation, with `quadratic` easing, with a duration of `1.0` seconds, going from `0.` to `1.` - and `simulateTick(0.5)`. - we'd expect the animation value to be `0.25` in that case)

> I was also wondering more specifically about how you wanted easeIn, easeOut, and easeInOut to be implemented/function. Are they default interpolation functions, or are they composed with an interpolation function

They're usually provided by default - for example, react-native has a bunch of [easing functions](https://facebook.github.io/react-native/docs/easing). There's details about how CSS defines the easing functions here: https://www.w3.org/TR/css-easing-1/ - they're all based on a cubic bezier timing function: https://www.w3.org/TR/css-easing-1/#cubic-bzier-timing-function

Hope that helps give some ideas!

Bretley:
@bryphe, my next hurdle is this fix in _build/default/examples/Bin.re on lines 22 and 33

In the useAnimation, easing must be defined...and when I edit it and run esy build, the changes are undone. Since it seems like I'm changing something downstream, where should I actually go to fix this?

Thanks

Cryza:
@Bretley Thanks for the work on this!

If I understood you correctly, you just changed some logic in `_build/default/examples/Bin.re` and now you're wondering why `esy` would undo your changes, right?

There are several versions of the `Bin.re` file after you ran the build at least once. That's because the compiler copies everything into the respective build folder before actually generating the compile output. By default, the compiled artifacts would go into the `_esy/default/` folder but we currently have our destination folder customized to `_build/`. This means you just changed a build artifact which will be overwritten with the original version from `src/` again on the next build.

TL;DR: Ignore the Reason files in `_build` and `_esy` and only ever change Reason files within `src` :)

Bretley:
@bryphe 
Progress update:
1) Test in AnimationTest.re does not actually change any of the other tests, only adds the quadratic one. I went back and checked this twice myself.
2) oneDimBezier alone is not enough to fully recreate the variety of css easing functions, but I'm 98% certain that a true bezier cubic could be made using [x,y] as a pair of oneDimBezier, then some math, then we can copy every CSS easing function offered.

PR #135, opened 2018-12-22
[DOC] Add information for mac users

czystyl:
I was tried to run examples from this repository yesterday. I faced the problem with running it on my `macOS` (endless build step :D). 
I solved this by preinstalling some libs. Big thanks to @bryphe and @ulrikstrid for help.

The second problem was with a path for built examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ npm install -g esy
 
 #### For `macOS` users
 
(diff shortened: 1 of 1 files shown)


bryphe:
Thank you @czystyl for the PR! Looks great - just added a couple tweaks. Hopefully will save other `macOS` users some trouble 😄

PR #137, opened 2018-12-22
Fix autosize=false in retina

jchavarri:
Tiny fix for retina displays where the size of the root node wasn't taking `pixelRatio` into account.

Before:

image

After:

image


czystyl:
I was wondering why on my mac it's displayed so crazy, and the reason is retina! Thanks 🙏

bryphe:
Phew, thanks for fixing these retina issues, @jchavarri ! 😄 So glad it's being tested!

PR #138, opened 2018-12-22
Start module for 'Revery.UI.Components'

bryphe:
We're almost to the point where we can start really building out our palette of cross-platform components! 🎉 

This creates a `Revery.UI.Components` module, with an initial start `<Clickable />` component.

This will be a place where we can build out more components - `<Button />`, `<Slider />`, `<TextInput />`, `<Label />` - out of the primitives + other components.


bryphe:
click-action

Using the new `<Clickable />` component - with @OhadRau 's borders! 😄

PR #139, opened 2018-12-22
API: Add 'ref' for primitives for direct access to Node

bryphe:
This change proposes adding a `ref` to the primitive nodes, that gives access to the underlying node. This isn't meant to be a common scenario by any means, but there are a few reasons I think we need this:

- For a control like `<Slider />`, we need to know the calculated layout values (the width/height) of the component. We can get that directly via the `node`, but perhaps a better event would be useful for that down the road - like `onLayoutChanged`?
- For a widget like `<Clickable />`, we really want to _capture_ the mouse input at the start of the `mousedown` gesture, and then, when it is released, check if the `mouseup` occurred in the bounds of the widget. We can use the `hitTest` method on the node for this.
- For focus management - we need some underlying handle to call `focus` on - I think the `node` is a natural choice

I'm open to other ideas for this (exposing `ref` always feels dirty 😄 ), but this can at least help us move forward on some of these other UI widgets.


bryphe:
Thanks for the review, @OhadRau ! 🎉 

> The one question I have is whether we want to use the term id here, as this could cause some confusion to people coming from HTML/if we want to use external stylesheets in the future.

Good point, I didn't think about this - I could see confusion. Do you have any alternatives? The only thing I can think of is perhaps making it `uniqueId`, but I'm not sure if that is better?

bryphe:
Thanks for the suggestions, @OhadRau ! Went with `internalId` 👍

PR #140, opened 2018-12-22
Treat mouse events as optional

jchavarri:
Small enhancement that leverages the language + JSX transform to make the `onMouse*` props really optional. This allows to remove the noop from `NodeEvents` but more importantly it avoids to perform all those noop calls for elements that don't set any event listeners, which removes some overhead (not sure really how much impact it would have in performance, but every little step counts! 😄 ).

From the users / API perspective, nothing changes (i.e. they can still pass raw callbacks to `onMouse*` props without the need to wrap these callbacks with `Some`).

### DIFF
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -80,10 +80,13 @@ class node ('a) (()) = {
   pub getMeasureFunction = (_pixelRatio: int) => None;
   pub handleEvent = (evt: NodeEvents.mouseEvent) => {
     let _ =
-      switch (evt) {
-      | MouseDown(c) => _this#getEvents().onMouseDown(c)
-      | MouseMove(c) => _this#getEvents().onMouseMove(c)
-      | MouseUp(c) => _this#getEvents().onMouseUp(c)
+      switch (evt, _this#getEvents()) {
+      | (MouseDown(c), {onMouseDown: Some(cb), _}) => cb(c)
+      | (MouseMove(c), {onMouseMove: Some(cb), _}) => cb(c)
+      | (MouseUp(c), {onMouseUp: Some(cb), _}) => cb(c)
+      | (MouseDown(_), _)
+      | (MouseMove(_), _)
+      | (MouseUp(_), _) => ()
       };
     ();
   };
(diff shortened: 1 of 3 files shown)


bryphe:
> Small enhancement that leverages the language + JSX transform to make the onMouse* props really optional.

Excellent! I like how this reduces the `verbosity` at the primitive level.  I didn't know you could do this sort of destructuring in pattern matching, either:
```
      switch (evt, _this#getEvents()) {
      | (MouseDown(c), {onMouseDown: Some(cb), _}) => cb(c)
```

That's handy!

>  This allows to remove the noop from NodeEvents but more importantly it avoids to perform all those noop calls for elements that don't set any event listeners, which removes some overhead (not sure really how much impact it would have in performance, but every little step counts! 😄 ).

Appreciate you removing this - no need for unnecessary overhead! 👍 

I just had one question that I put inline - but otherwise change looks great to me.

src/UI/Node.re
```diff
@@ -80,10 +80,13 @@ class node ('a) (()) = {
...
     let _ =
-      switch (evt) {
-      | MouseDown(c) => _this#getEvents().onMouseDown(c)
-      | MouseMove(c) => _this#getEvents().onMouseMove(c)
-      | MouseUp(c) => _this#getEvents().onMouseUp(c)
+      switch (evt, _this#getEvents()) {
+      | (MouseDown(c), {onMouseDown: Some(cb), _}) => cb(c)
+      | (MouseMove(c), {onMouseMove: Some(cb), _}) => cb(c)
+      | (MouseUp(c), {onMouseUp: Some(cb), _}) => cb(c)
+      | (MouseDown(_), _)
```
bryphe: Does it make sense to consolidate these last three cases into a single: `| (_, _) => ()` case?

jchavarri: Yeah, this is a bit more verbose but I did this explicitly so the compiler will fail if we ever add a new event to the `mouseEvent` variant type.

Using a wildcard `_` at the end of the switch statement for the first position of the tuple (`evt`) would swallow any future events added to `mouseEvent`. Wildcards are handy but one has to be prudent when using them! 😄

bryphe: Got it! Makes sense to me - it will save future event-adders time figuring out where they need to add new events. 😄 Thanks for the explanation!

Issue #145, opened 2018-12-22
Bug: Border does not render if background is transparent


bryphe:
__Issue:__ If a `border` is specified, but no `background`, the border does not render.

For example, for this component:
```
module Test = (
  val component((render, ~children, ()) =>
        render(
          () => {
            let borderStyle =
              Style.make(
                ~border=Style.Border.make(~width=10, ~color=Colors.white, ()),
                (),
              );

            let innerStyle =
                Style.make(
                    ~backgroundColor=Colors.red,
                    ~width=100,
                    ~height=100,
                    (),
                );

              <view style=borderStyle>
                  <view style=innerStyle />
              </view>
          },
          ~children,
        )
      )
);
```

I would expect to see this:
![image](https://user-images.githubusercontent.com/13532591/50378853-6bc6ee80-05f0-11e9-8565-a671224d3a9e.png)

But get: 
![image](https://user-images.githubusercontent.com/13532591/50378860-88632680-05f0-11e9-8ead-a9df77f60e60.png)

I actually hit this in #138 , which is why I added a pretty transparent background here:
https://github.com/bryphe/revery/blob/d05c343e67c90b81fdc19da74fbf59b5e6a87536/examples/Bin.re#L113

The issue seems to be this check: https://github.com/bryphe/revery/blob/d05c343e67c90b81fdc19da74fbf59b5e6a87536/src/UI/ViewNode.re#L241

OhadRau:
Whoops, I think in my head I was looking at that `c.a` as `style.opacity` when I went back to check that code. I'll fix this real quick

bryphe:
Fixed by #147 - thanks @OhadRau !

OhadRau:
I think this snuck back in when the box-shadows were merged...

bryphe:
Ah, bummer - thanks for catching it @OhadRau . I'll reopen to track.

Also trying to think about ways to catch these sorts of regressions via CI - put some thoughts in #156

PR #147, opened 2018-12-22
Fix border rendering in transparent views

OhadRau:
Quick fix for #145 -- couldn't get it to compile locally so this isn't tested yet. After pulling from upstream I keep getting the following error:

```
File "src/Core/Window.re", line 307, characters 30-52:                                                                                                                                              
Error: Unbound value Glfw.glfwDestroyWindow             
```

Tried to reinstall reason-glfw but I think this might be because of changes to the development version that haven't been pushed to NPM.


OhadRau:
Update: Figured out how to get rid of that error (had to delete the `_esy` dir and reinstall). Tested this and it's working with fully transparent views: 
image

bryphe:
Works great! Thanks @OhadRau for the fix!

PR #148, opened 2018-12-22
Fix hit test retina

### LINKED ISSUE #146: Bug: Click events coordinates not happening where they should in retina screens
There seems to be an issue where the events don't propagate to the layer that one would expect.

It might be related to the coordinates and `pixelRatio` conversion in retina screens.

In the gif below, I'm clicking on the `Click Me` button which should increase the counter, but the events end up reaching the logo image:

click-retina

@bryphe It might be an issue upstream in `glfw`, I wasn't sure so for now opening it here where the `Bin` example is 🙂

jchavarri:
Fixes #146.

Just getting the ball rolling, this seems to work and fix the original issue, but I'm not convinced about this solution as it feels wrong to pass raw dimensions in the events, when everything else is not using the raw dimensions. However, I tried the alternate approach of adjusting the dimensions in `Node#hitTest` but it didn't work as I expected. Maybe it's because of this call to `getWorldTransform`?

https://github.com/jchavarri/revery/blob/3a0ef1a123312c6ff396efa76e8c72d285c8bb32/src/UI/Node.re#L66

### DIFF
--- a/src/UI/Revery_UI.re
+++ b/src/UI/Revery_UI.re
@@ -53,10 +53,11 @@ let start =
     Revery_Core.Event.subscribe(
       window.onMouseMove,
       m => {
+        let pixelRatio = Window.getDevicePixelRatio(window);
         let evt =
           Revery_Core.Events.InternalMouseMove({
-            mouseX: m.mouseX,
-            mouseY: m.mouseY,
+            mouseX: m.mouseX *. pixelRatio,
+            mouseY: m.mouseY *. pixelRatio,
           });
         Mouse.dispatch(mouseCursor, evt, rootNode);
       },


bryphe:
@jchavarri , thanks for cleaning up my mess again 😄 

This looks like the correct fix to me, based on the current way we handle high-DPI / device scaling.

The crux of the issue is there are really two coordinate systems - _screen space_ (`glfwGetWindowSize`) and _pixel space_ (`glfwGetFramebufferSize`). When we make a call like `glfwSetWindowSize(800, 600)` we're using screen coordinates.

On low-dpi / standard screens, the screen space coordinates and pixel space coordinates are 1:1 - easy 😄 

On high-dpi screens, the pixel space coordinates could be some multiple of screen space coordinates. For example, if the pixel density is 3x, you might end up with a window with a screen space of `800x600` but pixel space dimensions of `2400x1800`.

There are multiple ways of handling this, but one design principle is key:
- Developers shouldn't have to explicitly worry about this (unless they want to!) - otherwise it'll be difficult to build apps that look correct across density displays. They should only worry about screen-space.

To accommodate this - we do a scaling step from node -> layout. This happens here:
https://github.com/bryphe/revery/blob/5ea7eb4209327b87d736c9b0f9e5e08b6173f5e2/src/UI/Style.re#L149

Another case we care about is the font loading:
```
      let font =
        FontCache.load(
          style.fontFamily,
          style.fontSize * parentContext.pixelRatio,
        );
```
We want to make sure we load full-density bitmaps - otherwise, we'd load up some low-res bitmaps and stretch them out - not ideal 😄 

One other thing we could do, that we don't handle currently, is getting the proper density image - for example, for iOS apps, they usually have multiple versions of an image (`image.png`, `image@2.png`, image@3.png`) - and they pick the right density image depending on the device DPI. That's something we could do too at some point.

So our layout actually happens in the _pixel space_! And, we use the calculated layout to hit-test via the `getWorldTransform` call. However, it looks like the mouse coordinate are in _screen space_ - so there is indeed a mismatch there - which is corrected with this PR (with the `pixelRatio` multiplication, they are now in _pixel space_).

I believe this is the right approach - given our current state of the world 👍 

The alternative would be to 'invert' the way we handle this scaling - use screen-space coordinates all the way down, including for layout (which would mean we could use these screen-space mouse coordinates for hit-testing), and do a final transform to go from screen-space to pixel-space (I think just setting our projection matrix in rendering to the 'screen' width/height would be sufficient). This might end up making the pipeline simpler (and less math - wouldn't need to do the scaling in the `Style.re`) - everything is just 'screen space' until the final rendering, where we could set the scaling via the projection matrix: https://github.com/bryphe/revery/blob/5ea7eb4209327b87d736c9b0f9e5e08b6173f5e2/src/UI/UiRender.re#L59 (if we went with this alternative approach, we'd just use the 'screen space' `Window.getSize` to set the dimensions for this projection matrix, instead of `Window.getFramebufferSize`)

__TL:DR;__ This is the right fix for the way things are today 😄 Maybe we can simplify / streamline this pipeline though, in the future!

bryphe:
Thanks @jchavarri for the quick fix for this!

jchavarri:
Thanks @bryphe !

Does this fix mean that if users subscribe to mouse events they'll see the coords in pixel space? That was my only concern, but I assume from your explanation that's not the case, right?

PR #149, opened 2018-12-23
Add simple easing functions

Bretley:
@bryphe 
Progress update:
1) Test in AnimationTest.re does not actually change any of the other tests, only adds the quadratic one. I went back and checked this twice myself.
2) oneDimBezier alone is not enough to fully recreate the variety of css easing functions, but I'm 98% certain that a true bezier cubic could be made using [x,y] as a pair of oneDimBezier, then some math, then we can copy every CSS easing function offered.

3) New PR to get this all off of master and on to a feature branch :)

### DIFF
--- a/examples/Bin.re
+++ b/examples/Bin.re
@@ -2,7 +2,6 @@ open Revery;
 open Revery.Core;
 open Revery.Math;
 open Revery.UI;
-open Revery.UI.Components;
 
 module Logo = (
   val component((render, ~children, ()) =>
@@ -21,6 +20,7 @@ module Logo = (
                   duration: Seconds(8.),
                   delay: Seconds(1.0),
                   repeat: true,
+                  easing: Animated.linear,
                 },
               );
 
@@ -32,6 +32,7 @@ module Logo = (
                   duration: Seconds(4.),
                   delay: Seconds(0.5),
                   repeat: true,
+                  easing: Animated.linear,
                 },
               );
 
@@ -68,6 +69,7 @@ module AnimatedText = (
                   duration: Seconds(1.),
                   delay: Seconds(delay),
                   repeat: false,
+                  easing: Animated.linear,
                 },
               );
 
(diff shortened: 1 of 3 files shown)


src/UI/Animation.re
```diff
@@ -29,8 +30,30 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
     toValue: float,
     repeat: bool,
+    easing: (float) => float,
   };
 
+  /*y=u0(1−t)3+3u1(1−t)2t+3u2(1−t)t2+u3t3
+    u0, u3 = 0,1; t in [0,1] 
+    equivalent to a normal bezier curve where p1x = 1/3 and p2x = 2/3
+   */
+  let genOneDimBezierFunc = (u1:float, u2:float) => {
```
bryphe: Nice! Thanks for setting this up, @Bretley 👍

test/UI/AnimationTest.re
```diff
@@ -40,13 +41,35 @@ test("Animation", () => {
...
+          duration: Time.Seconds(1.),
+          delay: Time.Seconds(0.),
+          toValue: 1.,
+          repeat: false,
+          easing: Animated.quadratic,
+        },
+      );
+
+    TestTicker.simulateTick(Time.Seconds(0.5));
+    expect(myAnimation.value.current).toBe(0.25);
```
bryphe: Excellent, thanks for the test!

examples/Bin.re
```diff
@@ -100,43 +103,6 @@ module AnimatedText = (
       )
 );
 
-module SimpleButton = (
```
bryphe: Did you mean to delete this here, or was it a merge issue?

Bretley: @bryphe  this is a merge issue for sure. At the time I was writing this, there was no button in the Bin.re test file. I tried to fetch upstream/master and then merge it with this branch, but git told me this branch was already up-to-date even though this code was missing. Not sure what commands I need to run to fix it.

bryphe: Ah ok. What I usually do for this is just get the latest, ie `git checkout master examples/Bin.re` - and then reapply the changes (adding the easing functions). Think that's the easiest way here, but there may be other ways.

bryphe:
Looks great @Bretley , thank you! Will bring it once the build completes 👍

PR #153, opened 2018-12-23
load images and fonts from executing dir

### LINKED ISSUE #141: Load assets relative to executable path
__REPRO:__ 
- Run `esy x Bin.exe` 

__Expected:__ App should launch.
__Actual:__ `Fatal error: exception Fontkit.FontKitLoadFaceException("[ERROR]: Unable to load font at FT_New_Face\n")`

The issue is that when we load fonts, or assets like images, we load them from the current _working directory_. This is problematic, because the user should be able to launch the executable from anywhere.

Because of this limitation - we have to include this awkward instruction in our `README.md`:
```
After you build, the executables will be available in the _build\install\default\bin folder.

NOTE: Currently the executables must be run from install\default\bin, since the assets are there.
```

However, the user should just be able to run via `esy x Bin.exe` to try out the example app.

Streamlining this would make the first-run experience much smoother, and also unblock #136 .

__Some things we need to do:__
- Have an API to get the executable's directory. I have an API like this I was working on in `rench`: <https://github.com/bryphe/rench/pull/15> - but we could just port over the relevant code, too:
```
let getExecutingDirectory = () => {
    Filename.dirname(Sys.argv[0]);
};
```
- Update our font loading and asset loading path to use the executingDirectory instead of the current working directory. We'd probably want to handle this in our `TextNode` and `ImageNode` classes, or we could handle it lower in the stack (`ImageRenderer`, `FontCache`). Append the executing directory + requested asset.

__Considerations:__
- We need to make sure this works fine in the JS strategy, too! I think all the logic should be the same if `getExecutingDirectory` simply returns `/` (the root) in the JSOO environment.

In the future, we might want to make our asset loading more flexible - some scenarios we'll potentially need to address:
- Looking for font files in the system font directory
- Allow custom paths (perhaps an app wants to put all its assets in a different place)

Akin909:
Fixes #141 @bryphe hopefully this addresses the problem of loading fonts and images using `esy x Bin` (tested and does work) in the way you intended

### DIFF
--- a/src/Core/Environment.re
+++ b/src/Core/Environment.re
@@ -12,3 +12,8 @@ let sleep = (t: Time.t) =>
   if (isNative) {
     Unix.sleepf(Time.to_float_seconds(t));
   };
+
+let getExecutingDirectory = () =>
+  Filename.dirname(Sys.argv[0]) ++ Filename.dir_sep;
+
+let getWorkingDirectory = () => Sys.getcwd();
--- a/src/Core/Environment.re
+++ b/src/Core/Environment.re
@@ -14,6 +14,6 @@ let sleep = (t: Time.t) =>
   };
 
 let getExecutingDirectory = () =>
-  Filename.dirname(Sys.argv[0]) ++ Filename.dir_sep;
+  isNative ? Filename.dirname(Sys.argv[0]) ++ Filename.dir_sep : "";
 
 let getWorkingDirectory = () => Sys.getcwd();
--- a/src/UI/FontCache.re
+++ b/src/UI/FontCache.re
@@ -1,14 +1,16 @@
+open Revery_Core;
 type fontInfo = (string, int);
 type t = Hashtbl.t(fontInfo, Fontkit.fk_face);
 
 let _cache: t = Hashtbl.create(100);
 
 let load = (fontName: string, size: int) => {
+  let execDir = Environment.getExecutingDirectory();
   let ret: Fontkit.fk_face =
     switch (Hashtbl.find_opt(_cache, (fontName, size))) {
     | Some(fk) => fk
     | None =>
-      let fk = Fontkit.load(fontName, size);
+      let fk = Fontkit.load(execDir ++ fontName, size);
       Hashtbl.add(_cache, (fontName, size), fk);
       fk;
     };
--- a/src/UI/ImageRenderer.re
+++ b/src/UI/ImageRenderer.re
@@ -1,5 +1,6 @@
 open Reglfw;
 open Reglfw.Glfw;
+open Revery_Core;
 
 type t = {
   mutable hasLoaded: bool,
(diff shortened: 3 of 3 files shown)


jchavarri:
This is awesome @Akin909 ! 🎉  Some related thoughts:

- The `esy.json` file I think it's an artifact used by older versions of `esy`. In the most recent versions, it was replaced by an `esy.lock` folder
- The addition of `Sys.getcwd()` probably will have some implications in projects targeting the browser through jsoo. js_of_ocaml doesn't support `Sys` that much, and from a quick test I did, it returns `/static` which doesn't make any sense in the folder structure of the test project I was using (there is no folder in that project called `static`). So maybe we have to create some abstraction over the different platform implementations like happened with `fontkit` or `glsw`. Maybe we could call it `reason-system`... or build it from `rench` (although, `rench` is more opinionated towards a Node.js-like API as far as I understand). I'd be happy to look into this as part of https://github.com/bryphe/revery/pull/116.
- Probably out of scope 😄  but I'd love to play around ideas around handling compile-time known assets paths at build time. I'm still unsure how that would look like exactly, but my idea right now is to read these paths at compile time, maybe through a ppx, read the assets from the ppx binary, and convert the `load` expression into an assignment to a binding of the whole path binary data (maybe as a string? or as binary data?). Something like https://github.com/johnwhitington/ppx_blob.

OhadRau:
Awesome work so far! This is a great first step towards shippable revery software.

@jchavarri If you're interested in compile time asset loading, that's some thing I was gonna try to make actually. I've thought of a few ways of doing it and once I settle on one I'll go ahead and write a PPX for it:

* Automatically guessing which files are statically accessible (if the path given to whatever `File.load` is a string literal && the file exists)
* Having a [@static] annotation that can be used when loading a file to inform the compiler that it's a static asset
* We can output each file as a linkable object file (e.g. https://www.linuxjournal.com/content/embedding-file-executable-aka-hello-world-version-5967) and generate extern declarations
* We could also make an array or hashtable of strings (binary or otherwise) that we load, basically the same as what ppx_blob would do
* For each of these, we have to let File.load know that these files are already loaded. We can do this by using a hashtable/other data structure of cached files and then inserting that as an optional parameter/overriding an empty data structure when we've loaded files

Akin909:
@jchavarri good point re. `Sys` not working in all environments :+1:, you sound like you've given it way more thought than I did so for this PR I've just put a guard to only try the `Sys` call if the env is native, but an actual future solution that can do a similar thing per environment would be awesome.

Re. the `esy.lock.json`, it seems to have already been in the project, is it unnecessary now, seems to change whenever I run `esy` on a branch without necessarily changing dependencies.

The ppx idea sounds very cool (I'm pretty new to reason and ocaml so would be beyond me at this point :laughing: )

bryphe:
Just tried out this PR - I can now run `esy x Bin.exe` on Windows! 🎉  This makes it so much easier to run the demos (at least for me), and means #136 is unblocked now.

PR #155, opened 2018-12-26
Memoize fontkit.fk_shape

### LINKED ISSUE #115: Performance: Cache results of `hb_shape` calls
As can be seen from the profile here:

image

We spend a _lot_ of unnecessary time shaping text over and over. We should be caching or memoizing this call - the shape results are constant for a (font family, text) tuple.

Akin909:
fixes #115, memoizes the fn in a similar way as the `getTexture` function

### DIFF
--- a/src/UI/FontRenderer.re
+++ b/src/UI/FontRenderer.re
@@ -39,8 +39,9 @@ type dimensions = {
   height: int,
 };
 
-let _memoizedFontShape = Memoize.make(Fontkit.fk_shape);
-let shape = (font, text) => _memoizedFontShape(font, text);
+let shapeFont = ((font, text)) => Fontkit.fk_shape(font, text);
+let _memoizedFontShape = Memoize.make(shapeFont);
+let shape = (font, text) => _memoizedFontShape((font, text));
 
 let measure = (font: Fontkit.fk_face, text: string) => {
   let shapedText = shape(font, text);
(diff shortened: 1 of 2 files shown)


src/UI/FontRenderer.re
```diff
@@ -39,8 +39,11 @@ type dimensions = {
   height: int,
 };
 
+let _memoizedFontShape = Memoize.make(Fontkit.fk_shape);
```
bryphe: There's one gotcha with our built-in memoization - it only supports _single argument functions_.

This means, with the current implementation, we'd actually only be memoizing the 'font' parameter. To speed up this code-path, we'll need to cache the tuple of (font, text).

So I think we'd need to update our memoize function to take in a pair of `(font, text)` - that's why we have the somewhat awkward syntax in our other memoized functions: https://github.com/revery-ui/revery/blob/7b8ccf7e9644d22359d512331bdcb4b63762fecd/src/UI/FontRenderer.re#L14

```
let _getTexture = ((font: Fontkit.fk_face, codepoint: int)) => {
...
}


let _memoizedGetTexture = Memoize.make(_getTexture);
let getTexture = (font: Fontkit.fk_face, codepoint: int) =>
  _memoizedGetTexture((font, codepoint));
```

Note the extra `(`/`)` in the `_getTexture` and the way we call `_memoizedGetTexture`. The `_getTexture` is actually taking a single argument of `(font, codepoint)`, and the `getTexture` is converting its arguments to a tuple to send them into `_memoizedGetTexture`.

Let me know if you have any questions!

PR #167, opened 2018-12-27
Documentation: Simplify example execution

bryphe:
With @Akin909 's fix in #153 - we can now streamline our the instructions for example execution 🎉 

I've updated the steps to just be `esy x Hello.exe`


bryphe:
Thanks for reviewing, @OhadRau !

Issue #164, opened 2018-12-27
UI Infrastructure: Event Bubbling


bryphe:
At the current time, we have a very simple event model for handling mouse events. 

That logic is in https://github.com/revery-ui/revery/blob/master/src/UI/Mouse.re , specifically here:
```
    let isNodeImpacted = n => n#hitTest(pos);
    let nodes: ref(list(Node.node('a))) = ref([]);
    let collect = n =>
      if (isNodeImpacted(n)) {
        nodes := List.append(nodes^, [n]);
      };
    Node.iter(collect, node);
    List.iter(n => n#handleEvent(eventToSend), nodes^);
```

This is _very simple_ - we check for all the nodes that pass the 'hit-test', and dispatch the event to all of them.

However, this is unexpected behavior and not intuitive if you're coming from web programming - as webdevs, we'd expect the event to dispatch to the _top-most_ element, and bubble up from there!

We need to implement this event-bubbling behavior in revery. This will be useful not just for the initial mouse events, but for all sorts of other events - like keyboard input, etc.

## Proposal

We add a `UiEvents` module that has a method `bubble = (node, event)`. The `bubble` would do a few things:
- It would wrap the event with some extra methods, like `stopPropagation` or `preventDefault` - like we'd expect in Web events.
- It would keep some internal state of whether propagation was stopped or default was prevented (ie, via `ref`s).
- It would call `handleEvent` for each node in the hierarchy. If `stopPropagation` is called, it should discontinue the traversal up the hierarchy.

## Testability

We should be able to craft unit tests that exercise this with some simple `Node` objects that have handlers that call `stopPropagation`, along with counters that validate whether the events were hit.

## Application

We can hook this up to the mouse event bubbling behavior today, by picking out the _top-most node_ that passes the hit-test. We need to make sure our z-index tracking is working correctly for this.

Akin909:
@bryphe @OhadRau was thinking of looking into this but if either of you have any advice re *how* to implement this in reason :smile: , I'd typically be thinking taking the event object adding some methods to it, and updating internal state of the event based on the function calls (but from an initial look into it, a lot of those assumptions don't carry over e.g. adding methods to an object for a start :laughing:)

Akin909:
Gonna give this a go, but might reach out for help at some point :+1:

Issue #144, opened 2018-12-22
API: Cursor Handling


bryphe:
Once https://github.com/bryphe/reason-glfw/issues/66 is implemented, we'll have the ability to change the cursor. This is important to give users the UX they expect - a way to show when an element is clickable when you hover over it, an I-beam to show the user that text input is available, etc.

When we have those APIs, we'll have to decide on a way to expose them in `revery`.

I think the most intuitive approach, for users coming from React in the browser, would be to have a `cursor` style property: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

We could leverage our mouse tracking + hit testing to figure out the 'cursor' style of the node the mouse is over, and decide how to call the `glfwSetCursor` API, based on that.

bryphe:
For example, even though its great we have some 'clickability' in #138  - the experience is not optimal:

![image](https://user-images.githubusercontent.com/13532591/50378813-ce6bba80-05ef-11e9-90c3-4d4f100fee1b.png)

When the mouse is over the 'Click Me' button, we should change from the pointer to a hand to indicate 'clickability.

PR #168, opened 2018-12-27
Add cursor styles API

OhadRau:
This addresses #144, implementing cursors as a style property (similar to how colors work now). I've also updated `Clickable` to use the `pointer` style cursor. Example of the API:
```
module Cursors = Revery_UI.MouseCursors;
<view style=Style.make(cursor: Cursors.text, ())>
  <text> "This looks like a textbox when you hover over it!" </text>
</view>
```

This also implements a really hacky precursor to event bubbling, where we select the top-most element below the mouse based on the depth in the DOM and choose the cursor from that node. While this works for basic cases, #164 would provide a much more flexible way of doing this (so let me know if you think I should block this until that's merged in). I also added a `Window.t` parameter to `Mouse.dispatch` to allow the cursor to get set in there (since `glfwSetCursor` requires that and we always have the window in scope when calling `Mouse.dispatch`).

Finally, here's a screenshot if you're interested in what this looks like:
image


OhadRau:
Just realized that this test is failing. Guess it's time to figure out what to do with the test, since it won't have a glfwWindow. I'm going to make it optional for now, but I'm open to refactoring this further in the future

bryphe:
> This addresses #144, implementing cursors as a style property (similar to how colors work now). I've also updated Clickable to use the pointer style cursor.

Awesome! 🎉 

> This also implements a really hacky precursor to event bubbling, where we select the top-most element below the mouse based on the depth in the DOM and choose the cursor from that node. 

This seems like a reasonable 'next step' to me. We need this logic to select the top-most element prior to having bubbling anyway - so this seems like it brings us closer to the logic we need in #164 .

> Just realized that this test is failing. Guess it's time to figure out what to do with the test, since it won't have a glfwWindow. I'm going to make it optional for now, but I'm open to refactoring this further in the future

IMO, it'd be ideal if we could keep the implementation of `Window/glfwWindow` out of the events. Thinking about the discussion here: https://github.com/revery-ui/reason-reactify/issues/35 - it might be that we want to extract a `revery-core` with some canonical set of components/events/layout, and so introducing some of our GLFW-specific implementation details would make that harder. In addition, having GLFW-specific details makes it harder to test this behavior in isolation.

One option I could see for removing this coupling would be to expose an event from the `Mouse` module - like `onCursorChanged`. Then, our logic here could look like:
```
switch (deepestNode^) {
         let glfwWin = window.glfwWindow;
         let cursor = node#getCursorStyle();
         Event.dispatch(onCursorChanged, cursor);
    };
```

We'd also need a place to listen to that event - the place that would make sense at the moment is the place where we bind all our other events between the 'native' world and the reconciler world: https://github.com/revery-ui/revery/blob/master/src/UI/Revery_UI.re

We could add a handler here, like:
```
 let _ =
    Revery_Core.Event.subscribe(
    Mouse.onCursorChanged, 
      cur => Glfw.glfwSetCursor(window, cur);
    );
```

It means that the `Mouse` itself doesn't have to care about the GLFW world, at all - it just can communicate the intent and let whatever downstream handler take care of the details.

The nice thing about this 'decoupling' is that it makes it easier to test, too - it gets us closer to being able to validate the logic w/o any rendering dependencies:
```
  test("onCursorChangedEvent gets dispatched with proper cursor", () => {
      let cursor = Mouse.Cursor.make();

      let count = ref(0);
     let cursor = ref(Cursors.arrow);
      let f = cur=> { count := count^ + 1; cursor := cur };
      let node =
        createNodeWithStyle(Style.make(~width=100, ~cursor=Cursors.text, ~height=100, ()));
      node#setEvents(NodeEvents.make(~onMouseDown=f, ()));

      let _ = Revery.Core.Event.subscribe(Mouse.onCursorChanged, f);

      Mouse.dispatch(
        cursor,
        InternalMouseMove({mouseX: 50., mouseY: 50.}),
        node,
      );
   
      expect(count^).toBe(1);
      expect(cursor^).toBe(Cursors.text);
    });
```
(I think there would still be issues as the cursors themselves depend on GLFW dependencies, but perhaps the `lazy` helps us here).

I find that writing tests helps not only to build up our safety net but also to rationalize the dependencies and how they relate. Just some thoughts - let me know what you think!

OhadRau:
Thanks for all this feedback. I'll try to fix this ASAP.

>(I think there would still be issues as the cursors themselves depend on GLFW dependencies, but perhaps the lazy helps us here).

Hm, so testing out something similar in utop:
```ocaml
utop # let x = lazy (print_endline "Hi");;
val x : unit lazy_t = <lazy>
utop # let y = lazy (print_endline "Bye");;
val y : unit lazy_t = <lazy>
utop # x = y;;
Exception: Invalid_argument "compare: functional value".
```

One workaround might be a 2-layer approach here, where layer one sets the cursor through some variant (i.e. `type cursorShape = Arrow | Text | Pointer`) and then layer two picks that out of some kind of lazy cache (this could be a record, hashmap, or just a function returning the current values we have).

bryphe:
Sorry about that, @OhadRau ! Just kicked off CI.

Thanks for adding the test case and abstracting out from the GLFW window implementation - looks great! 👍

Issue #54, opened 2018-11-16
Performance - Advanced: ContainerNode that uses a render texture


bryphe:
Often, for UI widgets / controls, they are relatively static - and don't require a re-render very frequently.

For those cases, it doesn't really make sense for us to re-render the entire widget every frame. This is costly and involves lots of transforms. If it doesn't get updated very often, it makes sense to render the widget to a render target (a texture). When it's cached - we'd just render a quad + that rendered texture.

__Proposal:__ We'd add a `<container>...</container>` tag that does this caching (it would correspond a `ContainerNode` that handle this).

The downside is that, if a widget is rendered frequently, it ends up being more expensive to render to the texture, plus use that texture to render a quad.

I thought about if it would be possible to automate this, but I think the application developer has the right 'domain' to know about this pattern. It would never be functionally necessary to use this `<container />`

In addition to the `ContainerNode` - itself, we'd have to add some dirty-tracking the container node would need to know how to check its children to see if invalidation of the cached texture is necessary.

It's similiar to how in CSS developers would use `translate3d(0, 0, 0)` to force layer promotion (ie, https://aerotwist.com/blog/on-translate3d-and-layer-creation-hacks/) - it's a similar concept - this would just be more explicit.

PR #171, opened 2018-12-29
Fix #170 - set a 'dirty' flag in response to reconciliation

### LINKED ISSUE #170: `setShouldRenderCallback` only returns true for animations
Currently, `setShouldRenderCallback` returns true only when there's an animation changing. However, there's other situations where it should always re-render, such as a style changing, an event happening, or a node's children/parent being updated. The program should also re-render whenever it loses focus, since windows overlapping over the draw area won't get cleared after they're moved (see image in #152). Furthermore, changing the callback doesn't seem to work (can't figure out why, but if you take Akin's example code and add `Window.setShouldRenderCallback(w, () => true);` it doesn't change the result). 

Regardless, we need a better metric for deciding when to re-render elements going forward as this can cause a lot of confusion to the user. One way to do this while keeping performance is to only re-render certain elements, marking them for re-rendering based on some of the events stated above (event, style change, parent/child change, & window losing focus. We might also need to take state changes into account).

bryphe:
__Issue:__ If a component calls `setState`, the UI does not get refreshed in response. This is very problematic for a UI, since the interesting interactivity will occur duet o state transitions.

__Defect:__ Our `setShouldRenderCallback` should be smarter - as pointed out in #170 by @OhadRau , there are additional cases we need to handle here.

__Fix:__ The `reason-reactify` reconciler was intended to help with this. It exposes both a `onBeginReconcile` and `onEndReconcile` events that are dispatched whenever a sub-tree is updated. For example, if you call `setState(...)` from some component, that event will be dispatched with the target node as the first argument. For now, we set a `dirty` flag to `true`, and clear that `dirty` flag on render.

There's a test case in `reason-reactify` that show how this functionality is intended to work:
https://github.com/revery-ui/reason-reactify/blob/cc0dbb453068f3693fc7fb692626c7d2f4ececcb/test/ContainerTest.re#L58

This PR addresses the particular case of updating the UI in response to nodes changing (which handles the style changing case, parent changing, an event happening, etc - as they all go through the reconciler).

__Future work:__
- As called out in #170 - there may be other cases where we need to leverage such a flag, like when a window loses focus.
- Incremental rendering - today, we always render the _full screen_. However, especially with something like #54 , we may be able to do much better in terms of just rendering a subtree.
- I suspect there are still _reconciler_ (`reason-reactify`) bugs we will run into - but that is good, it's expected we'll shake out the bugs in that space as we build out more real-world scenarios. 

Fixes #170 , but we may want to track some of the corollary work called out (ie, re-rendering on window losing focus) in a separate issue.

### DIFF
--- a/src/UI/Revery_UI.re
+++ b/src/UI/Revery_UI.re
@@ -39,7 +39,7 @@ let start =
     ) => {
   let uiDirty = ref(false);
 
-  let onEndReconcile = (_node) => uiDirty := true;
+  let onEndReconcile = _node => uiDirty := true;
 
   let rootNode = (new viewNode)();
   let container = UiReact.createContainer(~onEndReconcile, rootNode);
@@ -85,11 +85,14 @@ let start =
       },
     );
 
-  let _ = Reactify.Event.subscribe(FontCache.onFontLoaded, () => {
-    Window.render(window);
-  });
+  let _ =
+    Reactify.Event.subscribe(FontCache.onFontLoaded, () =>
+      Window.render(window)
+    );
 
-  Window.setShouldRenderCallback(window, () => uiDirty^ || Animated.anyActiveAnimations());
+  Window.setShouldRenderCallback(window, () =>
+    uiDirty^ || Animated.anyActiveAnimations()
+  );
   Window.setRenderCallback(
     window,
     () => {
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for the review @OhadRau !

PR #173, opened 2018-12-30
Feature/event bubbling

### LINKED ISSUE #164: UI Infrastructure: Event Bubbling

akinsho:
This PR will eventually add event bubbling to revery and will fix #164.

### DIFF
--- a/src/UI/UiEvents.re
+++ b/src/UI/UiEvents.re
@@ -80,7 +80,7 @@ let getDeepestNode = (node: node('a), pos) => {
 let rec traverseHeirarchy = (node: node('a), bubbled) =>
   BubbledEvent.(
     /*
-     track if default prevent or propagation stopped per module
+     track if default prevent or propagation stopped per node
      stop traversing node hierarchy if stop propagation is called
       */
     if (bubbled.shouldPropagate) {


OhadRau:
Hmm, correct me if I'm wrong but I think `Mouse.dispatch` is always called with the root node, so I think this would fail to bubble since the root has no parent. You can see the way I'm doing the pseudo-bubbling in the cursor-styles PR here: https://github.com/OhadRau/revery/blob/cursor-styles/src/UI/Mouse.re#L100-L117 (i.e. find the deepest node in the hierarchy and then get it's cursor style property which more or less bubbles up).

The `getDepth` method is implemented through a `_depth` reference that gets calculated when a node's parent changes (https://github.com/OhadRau/revery/blob/cursor-styles/src/UI/Node.re#L134-L137) and then I implemented `getCursorStyle` like this: https://github.com/OhadRau/revery/blob/cursor-styles/src/UI/Node.re#L71-L77. I think this sort of approach can be used for traversing the node hierarchy here too, but I might've missed something in your changes.

src/UI/UiEvents.re
```diff
@@ -0,0 +1,96 @@
...
+      | Some(parent) => traverseHeirarchy(parent, bubbled)
+      | None => ()
+      };
+    }
+  );
+
+let bubble = (node: node('a), event: mouseEvent) => {
+  /* Wrap event with preventDefault and stopPropagation */
+  let evt = BubbledEvent.make(event);
+  traverseHeirarchy(node, evt);
```
bryphe: NIT: Spelling - should be `traverseHierarchy`

src/UI/UiEvents.re
```diff
@@ -0,0 +1,96 @@
...
+      };
+  };
+};
+
+let rec traverseHeirarchy = (node: node('a), bubbled) =>
+  BubbledEvent.(
+    /* track if default prevent or propagation stopped per module
+       stop traversing node hierarchy if stop propagation is called
+        */
+    if (bubbled.shouldPropagate) {
```
bryphe: Logic looks great here - thanks for implementing the bubbling 👍

src/UI/UiEvents.re
```diff
@@ -0,0 +1,96 @@
...
+    let alreadyExists =
+      List.fold_left(
+        (found, elem) => found || elem.id == id,
+        false,
+        events^,
+      );
+    alreadyExists ?
+      wrappedEvent :
+      {
+        events := List.append([], [wrappedEvent]);
```
bryphe: It seems like we're always adding to the `events` list - but I didn't see where events are getting removed? It wasn't quite clear to me why we need a list of the events.

If we only care about the 'active' event - would it make sense to use a ref like `let activeEvent = ref(event)`?

src/UI/UiEvents.re
```diff
@@ -0,0 +1,96 @@
...
+      List.map(
+        evt => evt.id == id ? {...evt, defaultPrevented: true} : evt,
+        events^,
+      );
+    List.find(evt => evt.id == id, events^);
+  };
+
+  let make = event => {
+    let id = getId();
+    let wrappedEvent = {
```
bryphe: If `shouldPropagate` and `defaultPrevented` were defined either as `ref(bool)`, we could keep the implementation of `stopPropagation` and `preventDefault` local to the event, like:
```
let shouldPropagate = ref(true);
let defaultPrevented = ref(false);

let stopPropagation = () => shouldPropagate := false;
let preventDefault = () => defaultedPrevented := true;

let wrappedEvent = {
  id,
  event,
  shouldPropagate,
  defaultPrevented, 
  stopPropagation,
  preventDefault,
}; 
```

But perhaps I'm missing an advantage of them being in a list?

src/UI/UiEvents.re
Akin909: :+1:

test/UI/MouseTest.re
```diff
@@ -48,6 +49,26 @@ test("Mouse", () => {
...
     });
   });
+
+  test("bubbleEvent", () =>
+    test(
+      "test that state is updated per event when stop propagation is called",
+      () => {
+      let cursor = Mouse.Cursor.make();
+      let testEvent =
+        Mouse.internalToExternalEvent(
```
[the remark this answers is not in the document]
Akin909: Thanks, I've tweaked this now to just use the `MouseMove`

src/UI/Mouse.re
```diff
@@ -93,6 +94,7 @@ let dispatch =
   let pos = getPositionFromMouseEvent(cursor, evt);
 
   let eventToSend = internalToExternalEvent(cursor, evt);
+  bubble(node, eventToSend);
```
bryphe: I think this should actually be inside the `if(!handleCapture(..))` block.  The reason is - when a capture is active, it takes priority over other events. It gets 'first dibs' on handling the event, in other words. Helpful for a bunch of scenarios!

Could we replace that entire block inside with your new `bubble` function? The previous strategy collected all the nodes, and dispatched against the deepest node - but I think your `bubble` strategy supercedes now.

At the very least - I think we could replace:
```
    List.iter(n => n#handleEvent(eventToSend), nodes^);
```
with the new `bubble` function.

It seems like we still need some logic there to set the proper cursor style - but I think we could move `bubble` in to replace that `n#handleEvent` call.

src/UI/UiEvents.re
bryphe: Cool! I like the new approach better, thank you for making the change!

test/UI/MouseTest.re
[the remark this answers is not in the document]
bryphe: Excellent, thanks!

PR #174, opened 2019-01-02
Add Calculator demo

OhadRau:
Still a few things left to do, but this is a mostly working calculator.

Stuff (maybe) left to go:
* [x] Make operators show up in the display window before evaluation
* [x] Change font so we can have a backspace symbol
* [x] Figure out how to get relative sizing to work (is this possible with Revery yet?)
* [ ] Scale font size along with buttons so that it doesn't look tiny (maybe?)

I'm also not sure how great my use of state was here, but I guess since we don't do state the same way as ReasonReact (i.e. state+action) this would be pretty typical of a Revery app. Regardless, would like any suggestions for how to make this easier to understand or improve the code in general.

image


OhadRau:
Relative spacing is currently really weird to do with the `<Clickable>` tag, since we can't override it's style. We might want to block this PR on a change that'll allow us to override the Clickable's `flex-grow` and such. With `flex-grow` on the `<view>` inside the button we grow vertically. `flex-grow` on a `view` outside almost works, but the clickable area stays within the `<Clickable>` tag which isn't growing horizontally.

bryphe:
@OhadRau - this is awesome work! It's great to finally see an _interactive_ demo come together.

Testing on Windows:

calculator-demo

Like, it feels like you could build a __usable__ app with this now! So cool!

> I'm also not sure how great my use of state was here, but I guess since we don't do state the same way as ReasonReact (i.e. state+action) this would be pretty typical of a Revery app. Regardless, would like any suggestions for how to make this easier to understand or improve the code in general.

I think the way you used state here is perfectly appropriate (and highlights the usage of hooks within `revery`). We do have a concept of a redux-like state store, but I don't think it's necessary to use that in here. The nice thing about encapsulating the state the way that you did is the `<Calculator />` could be easily extracted as a component and embedded into other apps. It might be the case that we don't need the concept of an app-wide state store at all, and can just delegate to `useReducer`/`useState`. (One thing I'm interested in is introducing _time-travel debugging_ like the redux dev tools have - but I'm thinking that, with some of the thinking @cristianoc and @jchavarri have done on revery-ui/reason-reactify#47, we could get that for component state too. It's totally a tangent so I'll log a separate issue to track this).

> Relative spacing is currently really weird to do with the <Clickable> tag, since we can't override it's style. We might want to block this PR on a change that'll allow us to override the Clickable's flex-grow and such. 

Good point. This was an oversight by me, we should have a `style` property exposed on `<Clickable />`. I'm happy to either bring this PR in as-is (and then we can fix `<Clickable />` but update this example), or fix `<Clickable />` prior to merging this - whatever you'd prefer @OhadRau .

Sorry about the CI build issues - fixing this in #178 

Awesome example and thanks again! 👍

bryphe:
> Change font so we can have a backspace symbol

For this, we could bring in font-awesome, and render the backspace icon: https://fontawesome.com/icons/backspace?style=solid

I had some prototype code for this that I just ported over to the Oni2 repo: https://github.com/onivim/oni2/commit/2260c99700bd8f2f5dea79a28dcddc05a86da6f0

bryphe:
(Also, if you bring in latest `master`, the build should succeed)

OhadRau:
Awesome, I believe that @bmathews was working on exactly that in his `clickable-styles` branch, so I guess I'll just wait until we have that merged in (unless you'd prefer to have this in master now so we have a good demo).

bmathews:
@OhadRau @bryphe 

That is what I was going for, yep, but I'm not a reason expert so there might be a better way to do it:

https://github.com/revery-ui/revery/compare/master...bmathews:clickable-styles?expand=1

jchavarri:
Hey @OhadRau @bryphe I was looking at this calculator example (which is awesome!! 🚀 ) and wanted to explore how `useReducer` would look in this case, on one side because it's such a exemplary "state machine" case, plus it was a good chance to test that API that was added recently.

I created a PR from this branch in https://github.com/OhadRau/revery/pull/1/files but this is the gist of the different types:

```reason
type state = {
  operator, /* Current operator being applied */
  result: float, /* The actual numerical result */
  display: string, /* The equation displayed */
  number: string /* Current number being typed */
};

type action =
  | BackspaceKeyPressed
  | ClearKeyPressed(bool) /*AC pressed */
  | DotKeyPressed
  | NumberKeyPressed(string)
  | OperationKeyPressed(operator)
  | PlusMinusKeyPressed
  | ResultKeyPressed;
```

then the reducer (i'm skipping the `eval` function which would be almost 1:1 what exists already):

```reason
let reducer = (state, action) =>
  switch (action) {
  | BackspaceKeyPressed =>
    state.number == "" ?
      state :
      {
        ...state,
        number: String.sub(state.number, 0, String.length(state.number) - 1),
      }
  | ClearKeyPressed(ac) =>
    ac ?
      {...state, operator: `Nop, result: 0., display: ""} :
      {...state, number: ""}
  | DotKeyPressed =>
    String.contains(state.number, '.') ?
      state : {...state, number: state.number ++ "."}
  | NumberKeyPressed(n) => {...state, number: state.number ++ n}
  | OperationKeyPressed(o) =>
    let (result, display) = eval(state, o);
    {operator: o, result, display, number: ""};
  | PlusMinusKeyPressed =>
    if (state.number != "" && state.number.[0] == '-') {
      {
        ...state,
        number: String.sub(state.number, 1, String.length(state.number) - 1),
      };
    } else {
      {...state, number: "-" ++ state.number};
    }
  | ResultKeyPressed =>
    let (result, display) = eval(state, `Nop);
    {operator: `Nop, result, display, number: showFloat(result)};
  };
```

and finally the consumption in the component:

```reason
let ({display, number, _}, dispatch) =
  useReducer(
    reducer,
    {operator: `Nop, result: 0., display: "", number: ""},
  );
```

I think either `useState` or `useReducer` are fine, we can add `useReducer` to another example in the future 😄 Let me know what you think!

OhadRau:
Yeah that actually looks good to me @bmathews. In terms of functions to extend style that's probably the cleanest we can get it. The only alternative I can really think of would be to define styles without the Style.make function and to use the {...spread} syntax for records, but that's up to the user to do on their own.

OhadRau:
@jchavarri Awesome! Didn't realize we have a useReducer function but I think that'll actually help me clean up the current code a lot. Since that's not in any examples already this could be a good place to add it.

examples/Calculator.re
```diff
@@ -0,0 +1,382 @@
...
+              | Key.KEY_4 => dispatch(NumberKeyPressed("4"))
+              | Key.KEY_5 => dispatch(NumberKeyPressed("5"))
+              | Key.KEY_6 => dispatch(NumberKeyPressed("6"))
+              | Key.KEY_7 => dispatch(NumberKeyPressed("7"))
+              | Key.KEY_8 => dispatch(NumberKeyPressed("8"))
+              | Key.KEY_9 => dispatch(NumberKeyPressed("9"))
+
+              | _ => ()
+            };
+            /* TODO: Pretty sure this isn't supposed to go in the render() function.
```
bryphe: Good intuition, I think it is subscribing to the event on each re-render!

You could try this: 
```
useEffect(() => {
let unsubscribe = Event.subscribe(window.onKeyDown, respondToKeys);
unsubscribe
});
```

On each re-render, it will unsubscribe from the previous event subscription.

jchavarri: @bryphe Would this be a good opportunity to use the `MountUnmount` condition? Like in [this test](https://github.com/revery-ui/reason-reactify/blob/944c40124edd51315c8bc4691fec1258fc8f7363/test/HooksUseEffectTest.re#L55-L62)?

bryphe: @jchavarri - I think it'd be reasonable (and better for perf)! Not sure if we'd hit that long-lived `setState` issue though in revery-ui/reason-reactify#44 ?

bryphe:
One other cool thing - with revery-ui/reason-fontkit#13, and aside from a minor glyph bug (revery-ui/reason-fontkit#14 - not related to this PR)... this calculator actually works and is usable in the browser too as a WebGL app! 😍 

image

examples/Calculator.re
OhadRau: @bryphe Is `useEffect` exposed right now? I don't see it in `Hooks` and doesn't look like there's access to `UiReact` anywhere. Want me to add this to the PR?

bryphe: Ah sorry - it wasn't exposed! Thanks for adding it @OhadRau 👍

OhadRau:
Merged in @bmathews' clickable-styles branch & then used that to make the buttons scale. Might still investigate making the font size scale along with the buttons since it looks a little funny, but in the meantime would you mind reviewing this again since it's some new changes @bryphe?

bryphe:
Just grabbed the latest! Code looks great, and I love the way calculator works now (handling key input is a big plus!). Thanks for the `clickable-styles` work too @bmathews !

image

I really like the first-class icon we have for backspace too. Consider me signed-off @OhadRau  👍

bryphe:
Just checked the build - looks like it hit this error: ```
2019-01-08T00:05:40.8605848Z File "examples/dune", line 2, characters 4-227:
2019-01-08T00:05:40.8606706Z 2 |     (names Hello Autocomplete Flexbox Border Boxshadow DefaultButton Overflow)
2019-01-08T00:05:40.8607220Z 3 |     (preprocess (pps lwt_ppx))
2019-01-08T00:05:40.8607358Z 4 |     (package Revery)
2019-01-08T00:05:40.8607469Z 5 |     (public_names Hello Autocomplete Flexbox Border Boxshadow DefaultButton Calculator Overflow)
```

**Comment by bryphe** [comment]
Ah looks like there is still one more build failure:
```
File "src/UI/Style.re", line 122, characters 15-556:
Error: Some record fields are undefined: overflow
```

I pushed up a [commit with the fix](https://github.com/revery-ui/revery/commit/27970c37efa9d420b9935a0cb59fb67a2584b300), you might just need to cherry-pick it (`git cherry-pick 27970c37`). Hopefully CI is green after that!

OhadRau:
Merging now! Thanks for all the help

PR #177, opened 2019-01-02
Move borders into width/height bounding box

### LINKED ISSUE #175: Border styles cause incorrect layout and hitTest
With a 50px border (orange) added to the button element, it is no longer centered correctly in the flex layout, and the inner contents (white) of the element expand unexpectedly. 

image

Example repo:
https://github.com/revery-ui/revery/compare/master...bmathews:clickable-styles?expand=1

OhadRau:
This PR fixes #175. Previously, borders were rendered outside of the computed dimensions of a node. As #175 points out, this was the incorrect way to do things, which caused the hitbox to be incorrect, the node to appear uncentered, and the inside of the node to appear too large when a large border width was used. Big thanks to @bmathews for finding this!

Here's a screenshot of the before/after:
Before
After

### DIFF
--- a/src/UI/ViewNode.re
+++ b/src/UI/ViewNode.re
@@ -36,6 +36,9 @@ let renderBorders =
   let (bottomBorderWidth, bottomBorderColor) =
     borderStyle(style.borderBottom, style.borderVertical, style.border);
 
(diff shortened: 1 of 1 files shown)


bryphe:
Nice catch @bmathews, and thanks for the fix @OhadRau . Looks great!

PR #179, opened 2019-01-03
API: Expose the context API from Revery.UI

bryphe:
Need to bring out a few extra `reactify` methods and make them available from `Revery.UI`.


bryphe:
Needed by onivim/oni2#6

PR #183, opened 2019-01-04
CI: Check formatting

bryphe:
This adds a CI check to verify the formatting is correct, and adds an `esy format` command to run `refmt` against all the files in the repo.


bryphe:
Thanks for the idea @jchavarri ! `husky` would be a nice way to enforce this - either as a precommit/prepush hook - would save some time across the board 😄 

Unfortunately it looks like there are issues integrating `husky` with `esy` at the moment: https://github.com/esy/esy/issues/623

jchavarri:
Ah, cool didn't know. Thanks @bryphe for pointing me to that issue! 🙂

PR #190, opened 2019-01-05
High-DPI display clean-up

bryphe:
__Issue:__ There is a lot of complexity today in how we handle _high-dpi_ displays (ie, retina displays).

The challenge is there are really two 'coordinate systems' that appear. The first is _screen space_ - if I ask GLFW for a window of size `800x600`, these are _screen space_ coordinates. The other coordinate system is pixel space. For a high-dpi display, like retina, there might be a multiple of 2-3 pixels per screen space.

So, for a window of _screen space_ size `800x600`, with a display with a pixel density of 3, the functions will return the following:
- `glfwGetWindowSize` - `800x600`
- `glfwGetFramebufferSize` - `2400x1800`

At the end of the day, when we're drawing, internally, we care about the framebuffer size - but we don't want the user to have to worry about this. Their app should look the same regardless of DPI - imagine how confusing it'd be if you ask for a window of width `800`, and then create an element of `800` - and it only takes up a third of the screen 😨 The browser takes care of this for DOM elements, and we should do the same in `revery`.

We do have a scaling solution in place, however, we convert all the layout nodes / sizes to the _pixel space_. This leads to some confusion, and the DPI information leaking out - for example, if I create a `200x200` rectangle, grab a `ref` to it, and ask for the dimensions, in a high-dpi screen with pixel density 3, I'll get 600x600 for the rectangle. This complicates things like hit-testing and mouse events where you always need to be cognizant and converting between these coordinate systems.

To address this, and simplify the pipeline, we'll have all the UI rendering / layout / events happen in _screen space_. In addition, we'll set the projection matrix to be in screen space, so the coordinate system for vertices and positions is in screen space, even if the frame buffer is larger (the projection matrix will convert it for us).

There are still cases internally where you have to be cognizant of the DPI:
- Rendering fonts - If the user asks for a font size 20 - freetype renders in pixel space, so we have to convert the font size, otherwise we'll get blurry text in a high-dpi display.
- Images - it'd be nice to pick high-dpi images like iOS/Apple does, depending on DPI.

We also expose the `Window.getDevicePixelRatio` as an API, but hopefully the user doesn't have to worry about it too much. There are definitely cases where you'd want to render different UI based on the display density, so it makes sense to keep that around. But hopefully keeping the majority of our pipeline in _screen space_ makes it simpler to reason about events / positions.

It's also perf boost - we save a lot of 'scaling' calculations we were doing when creating layout nodes from styles!

This idea was originally discussed in https://github.com/revery-ui/revery/pull/148 - wanted to run it by you @jchavarri 👍 

As an aside, I found that I'm able to simulate a high DPI display following the instructions here: https://stackoverflow.com/questions/12124576/how-to-simulate-a-retina-display-hidpi-mode-in-mac-os-x-10-8-mountain-lion-on

Tested in the following environments:
- [x] Standard DPI - native
- [x] Standard DPI - webgl
- [x] High DPI (2x) - native
- [x] High DPI (2x) - webgl
- [x] Test auto-sizing example (completion) - standard DPI - native
- [x] Test auto-sizing example (completion) - high DPI (2x) - native

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
   "homepage": "https://github.com/bryphe/revery#readme",
   "esy": {
     "build": ["dune build --root . -j4"],
-    "buildsInSource": "_build",
     "install": [
         "esy-installer Revery.install",
         "esy-installer Revery_Core.install",
@@ -44,8 +43,7 @@
     "@opam/cmdliner": "1.0.2",
     "@opam/js_of_ocaml": "github:ocsigen/js_of_ocaml:js_of_ocaml.opam#db257ce",
     "@opam/js_of_ocaml-compiler": "github:ocsigen/js_of_ocaml:js_of_ocaml-compiler.opam#db257ce",
-    "@esy-ocaml/reason": "facebook/reason#ab49908",
-    "reason-glfw": "github:bryphe/reason-glfw#e01ab0d"
+    "@esy-ocaml/reason": "facebook/reason#ab49908"
   },
   "peerDependencies": {
     "ocaml": "~4.6.0"
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "homepage": "https://github.com/bryphe/revery#readme",
   "esy": {
     "build": ["dune build --root . -j4"],
+    "buildsInSource": "_build",
     "install": [
         "esy-installer Revery.install",
         "esy-installer Revery_Core.install",
@@ -43,7 +44,8 @@
     "@opam/cmdliner": "1.0.2",
     "@opam/js_of_ocaml": "github:ocsigen/js_of_ocaml:js_of_ocaml.opam#db257ce",
     "@opam/js_of_ocaml-compiler": "github:ocsigen/js_of_ocaml:js_of_ocaml-compiler.opam#db257ce",
-    "@esy-ocaml/reason": "facebook/reason#ab49908"
+    "@esy-ocaml/reason": "facebook/reason#ab49908",
+    "reason-glfw": "github:bryphe/reason-glfw#e01ab0d"
   },
   "peerDependencies": {
     "ocaml": "~4.6.0"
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
   "homepage": "https://github.com/bryphe/revery#readme",
   "esy": {
     "build": ["dune build --root . -j4"],
+    "buildsInSource": "_build",
```
bryphe: TODO: Remove

bryphe:
Was starting to look at implementing a very minimal `overflow: hidden` implementation using `glScissorRect`, but hit some complications with the coordinate systems here.

I also hit the ref-sizing issue mentioned when I was implementing a PoC slider control: https://github.com/revery-ui/revery/tree/bryphe/prototype/slider

bryphe:
I'll bring this in for now so I can bring in the overflow change (next step for a `<ScrollView />`!), but let me know if there is any feedback (or if anyone sees issue on high-DPI displays!)

src/Core/Window.re
```diff
@@ -128,20 +130,28 @@ let create = (name: string, options: windowCreateOptions) => {
   | _ => ()
   };
 
-  let w = Glfw.glfwCreateWindow(options.width, options.height, name);
+  let w = Glfw.glfwCreateWindow(1, 1, name);
```
jchavarri: This is setting the window to the minimum (1x1) and then it changes it to the actual size below, right? What's the benefit of calling `glfwSetWindowSize`?

bryphe: Oops! I accidentally left this in there (prior to bringing over the `glfwGetWindowSize`). Previously - I didn't have the `glfwGetWindowSize` API available, so I was using this 'hack' to trigger a callback to measure the actual width/height of the window... but now that the `glfwGetWindowSize` is available, there's no advantage and this is just unnecessary complexity. Will revert and submit a fix shortly.

Thanks for looking it over @jchavarri - and nice catch!

Issue #180, opened 2019-01-03
Resizing window seems to cause division by zero


WhoAteDaCake:
The culprit seems to be: [UiRender](https://github.com/revery-ui/revery/blob/master/src/UI/UiRender.re#L32)

This occurs because of convertion from float to int, when our ratio is < 1.
Easiest way to fix it would be to just add

`let pixelRatio = pixelRatio === 0 ? 1 : pixelRatio;`

But I am not sure if that's the best solution. I would be happy to submit a PR for it.

bryphe:
Thanks @WhoAteDaCake ! Yes, I saw this come up a few times too, but haven't narrowed it down.

> Easiest way to fix it would be to just add

> `let pixelRatio = pixelRatio === 0 ? 1 : pixelRatio;`

Seems reasonable to me. I suspect that during the resize - the callbacks for getting the _screen_ window size and _framebuffer_ size occur at different times - and depending on when render is called (if the screen size is updated, but not framebuffer), we could hit this.

Alternatively - now that I think about it - perhaps it would be best to just support a `float` pixel ratio? I believe there are some cases where the pixel ratio gets reported as a fractional value, like `2.5`, so that seems like the best long-term fix.

WhoAteDaCake:
I think using float might be a good alternative. I imagine that would give more accurate results. don't think, that it would make performance worse, as we would only need 2 conversions for sizes instead of 1 (In order to pass style sizes)

bryphe:
> How would that affect the performance? 

Good question @WhoAteDaCake , thanks for thinking about perf!

I don't think switching this from `int` to `float` will introduce a new bottleneck in our rendering - worst case, if we find that it does, we could cache the value on the `Window.t` instance. 👍

tcoopman:
This error still happens for me on current master?

`Fatal error: exception Division_by_zero`

bryphe:
Yes, looks like I'm still hitting it when 'maximizing' and then 'restoring' the example app

PR #191, opened 2019-01-05
Use float pixel ratio

WhoAteDaCake:
Will prevent division by 0 crash as mentioned in #180.

### DIFF
--- a/src/UI/TextNode.re
+++ b/src/UI/TextNode.re
@@ -117,7 +117,7 @@ class textNode (text: string) = {
       /* TODO: Cache font locally in variable */
       let style = _super#getStyle();
       let font =
-        FontCache.load(style.fontFamily, style.fontSize * pixelRatio);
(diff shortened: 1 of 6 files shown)


WhoAteDaCake:
Not quite sure if `NodeDrawContext` should also take float, that might mean some extra work

WhoAteDaCake:
Thanks @OhadRau I've updated the tests

src/UI/TextNode.re
```diff
@@ -117,7 +117,7 @@ class textNode (text: string) = {
       /* TODO: Cache font locally in variable */
       let style = _super#getStyle();
       let font =
-        FontCache.load(style.fontFamily, style.fontSize * pixelRatio);
+        FontCache.load(style.fontFamily, int_of_float(float_of_int(style.fontSize) *. pixelRatio));
```
bryphe: Nice catch here!

bryphe:
Merging now, thanks for the contribution @WhoAteDaCake !

PR #194, opened 2019-01-06
Add initial implementation of inset shadowing

Akin909:
#### Outstanding
- [ ] Render the inset shader so the center is transparent and the shadow on the edges are the shadow color specified

### DIFF
--- a/examples/Boxshadow.re
+++ b/examples/Boxshadow.re
@@ -18,7 +18,7 @@ let init = app => {
   Window.setPos(win, centerX, 0);
 
   let parentWidth = width - 10;
-  let parentHeight = height / 2;
+  let parentHeight = height;
 
   let parentStyles =
     Style.make(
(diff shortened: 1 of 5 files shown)


Akin909:
@bryphe I could use a hand here, I've gotten to the point where I'm rendering another shader on top of the original, which I intend to represent the inset, I'm having trouble figuring out how to create a shader that is translucent at the center or at least the same color as the underlying quad but then blurs with the color of the shadow specified.

Not sure if it's reasonable to actually just draw the main quad then layer the shadow quad then place a smaller version of the main quad on top. That sounds pretty messy though and having had a look on GLSLPlayground I'm sure there's a way to do it much nicer than that I just havent a clue how :laughing: :sob:

OhadRau:
I don't know if this is the "right" way to do it, but you can use something like a fresnel shader to make the shadow appear only on the edges.

Akin909:
@OhadRau thanks :+1: it's worth a shot since I don't really know what else to try :smile:

bryphe:
Looks like we'll be able to do this with Skia! Here's one example: https://fiddle.skia.org/c/a09da324e79af60c897914458c229d25

PR #200, opened 2019-01-08
Bugfix: 'ref' not available reliably

bryphe:
__Issue:__ Discovered this while working on a prototype slider control: [`bryphe/prototype/slider`](https://github.com/revery-ui/revery/tree/bryphe/prototype/slider).

Snippet of code with the issue:
```
let (ref, setRef) = useState(None);
         ...
let onMouseDown = (_evt) => {
       switch(ref) {
       | Some(r) => print_endline("got ref: " ++ string_of_int(r#getInternalId()));
       | None => print_endline("no ref");
       };
}
         ...
<view onMouseDown style ref={(r) => setRef(r)}> 
```

__Defect:__ The `ref` would not be available reliably. The `ref` callback is being called correctly, but when we use `setRef` it's happening at a time when the instance hasn't been mounted completely in the reconciler - so it's not able to persist the state. Later renders give back `None`.

__Fix:__ My proposed fix for this is to have this event be 'queued' such that we only dispatch it when reconciliation is complete. This is doable since we have the `ref` model implemented completely in 'user-space' - in other words, `reason-reactify` doesn't model `ref` today.

This would create an event queue for nodes - and only dispatch it on request (which `revery` will flush these pending events after every render).

__Open Questions__
- Should we look at add `ref` as a primitive by default in `reason-reactify`. If so, would it make sense to have a `useRef` hook we include as well? If we push it to the framework - we might be able to come up with a more efficient implementation.
- This also seems related to the idea of batching updates used in `ReactMini` and mentioned in https://github.com/revery-ui/reason-reactify/pull/51 - which also makes me consider that the reconciler would be the right place to push-back this fix. Depending on how that issue is fixed - I suspect the batching updates - enforcing that 'useState' is called at a correct time with more precision - would also address the issue.

__TODO:__
- [x] Add test case exercising the issue
- [x] Implement event-queue fix
- [x] Better comments in the `Node` module
- [x] Fix formatting


bryphe:
Fix with the upgrade to `brisk-reconciler` - no need to queue up ref callbacks anymore.

PR #205, opened 2019-01-12
add input component and example

Akin909:
This is a v. basic implementation of an input component there's a fair bit that could be added (down the line I think) like some padding, Events (mouse, focus etc), but this is a starting point for that.

### Updated
* Position cursor on the left when placeholder is present
* Add padding so text isn't flush with border
* Added text cursor
* Added mouse focusing
* add props to allow for more styling

input-2

### Issues/Bugs discovered

* I've found that trying to use an effect based on state, when the effect has be specified as on mount means the state value is in a closure __I think__, though switching to `Always` means that a handler is never missed, example

```reason
let ... = React.Hooks.effect(
  Always, /* if always this handler overrides any other handler registered for the same event*/
  (event) => 
    Some(
     Event.subscribe(
       window.keyDown,
       ()  => {
        if (state.isFocused) {
          onChange(state.value) /* on mount this value is old */
         }
       }
    )
  )
```

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -30,7 +30,7 @@ let state: state = {
     {name: "Box Shadow", render: _ => Boxshadow.render()},
     {name: "Focus", render: _ => Focus.render()},
     {name: "Stopwatch", render: _ => Stopwatch.render()},
-    {name: "Input", render: w => Input.render(w)},
+    {name: "Input", render: w => InputExample.render(w)},
   ],
   selectedExample: "Animation",
 };
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -97,18 +97,11 @@ let reducer = (s: state, a: action) => {
 let init = app => {
   let maximized = Environment.webGL;
 
-  let size = Monitor.getPrimaryMonitor() |> Monitor.getSize;
-
   let win =
     App.createWindow(
       app,
       "Welcome to Revery!",
-      ~createOptions={
-        ...Window.defaultCreateOptions,
-        width: size.width,
-        height: size.height,
-        maximized,
-      },
+      ~createOptions={...Window.defaultCreateOptions, maximized},
     );
 
   let render = () => {
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -30,6 +30,7 @@ let state: state = {
     {name: "Box Shadow", render: _ => Boxshadow.render()},
     {name: "Focus", render: _ => Focus.render()},
     {name: "Stopwatch", render: _ => Stopwatch.render()},
+    {name: "Input", render: w => Input.render(w)},
   ],
   selectedExample: "Animation",
 };
@@ -37,11 +38,10 @@ let state: state = {
 let noop = () => ();
 
 let getRenderFunctionSelector: (state, Window.t) => React.syntheticElement =
-  (s: state) => {
+  (s: state) =>
     List.filter(x => String.equal(x.name, s.selectedExample), state.examples)
     |> List.hd
     |> (a => a.render);
-  };
 
 module ExampleButton = {
   let component = React.component("ExampleButton");
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -97,11 +97,18 @@ let reducer = (s: state, a: action) => {
 let init = app => {
   let maximized = Environment.webGL;
 
+  let size = Monitor.getPrimaryMonitor() |> Monitor.getSize;
+
   let win =
     App.createWindow(
       app,
       "Welcome to Revery!",
-      ~createOptions={...Window.defaultCreateOptions, maximized},
+      ~createOptions={
+        ...Window.defaultCreateOptions,
+        width: size.width,
+        height: size.height,
+        maximized,
+      },
     );
 
   let render = () => {
--- a/src/UI_Components/Input.re
+++ b/src/UI_Components/Input.re
@@ -56,18 +56,15 @@ let reducer = (state, action) =>
   | ClearWord => {...state, value: ""}
   };
 
-let handleKeyDown = (~dispatch, ~onSubmit, event: Events.keyEvent) =>
+let handleKeyDown = (~dispatch, event: Events.keyEvent) =>
   switch (event.key) {
   | Key.KEY_BACKSPACE => dispatch(Backspace)
-  | Key.KEY_ENTER => onSubmit(event)
   | _ => ()
   };
 
-let handleKeyPress = (dispatch, event: Events.keyPressEvent) =>
-  String.length(event.character) > 0 ?
-    dispatch(UpdateText(event.character)) : ();
+type changeHandler = (~value: string) => unit;
 
-let noop = _evt => ();
+let noop: changeHandler = (~value as _value) => ();
 
 include (
           val component(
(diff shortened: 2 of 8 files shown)


Akin909:
The reconciler is I *think* behaving oddly as when I try and set state in the example component it doesn't get updated correctly

src/UI_Components/Input.re
```diff
@@ -0,0 +1,209 @@
...
+let noop = (~value as _value) => ();
+
+include (
+          val component(
+                (
+                  render,
+                  ~children,
+                  ~window,
+                  ~color=Colors.black,
+                  ~fontSize=18,
```
bryphe: For some of these properties (`fontSize`, `color`, etc) - would it make it sense to take in a `style` instead? I think the tricky thing is letting a user set `Style` but having a set of default values that aren't set (and if they aren't set, they get overridden) - we don't have a good solution for that yet (but we should)

For example, I could see customizing the `fontFamily`, `border`, or `backgroundColor` as common scenarios here too.

src/UI_Components/Input.re
```diff
@@ -0,0 +1,209 @@
...
+                  ~width=200,
+                  ~height=50,
+                  ~value=None,
+                  ~placeholder="",
```
bryphe: That's really neat that there is placeholder text - just tried out the demo 💯

src/UI_Components/Input.re
rauanmayemir: Have you thought about using polymorphic lists instead of records for style? You could then have a common set of props and extra props per component. (and do not set values for props that are not set)

Akin909: @rauanmayemir I've been thinking about that a lot recently (was going to raise an issue) as I've used  https://github.com/SentiaAnalytics/bs-css which looks like - 
```reason
  let card = style([
    display(flexBox),
    flexDirection(column),
    alignItems(stretch),
    backgroundColor(white),
    boxShadow(~y=px(3), ~blur=px(5), rgba(0, 0, 0, 0.3)),
    /* You can place all your theme styles in Theme.re and access as normal Reason module */
    padding(Theme.basePadding)
  ]);
```

and [bs-nice](https://github.com/threepointone/bs-nice) which has an even nicer API imo
```reason
let className = css([|
  Display(Flex),
  AlignItems(Center),
  Color(Red),
  Select(":hover", [Color(Blue)])
|]);
```

wokalski: `css` is a sort of odd when it comes to styling because all style properties are accepted by all components (weird.)

Polymorphic variants allow us to have one style (a `list`) used by several components at the same time, as long as the list only contains styles that are shared by them. Taking `Input` and `View` as an example this would work: ``` [`BackgroundColor(c), `Width(300.)]``` but this would not ``` [`BackgroundColor(c), `ContentsAspectRatio(x)]``` because aspect ratio does not make sense for `View`. The actual api can be a little different, i.e. we could use functions like `backgroundColor(c)` but I'm talking about types specifically now.

Akin909:
@bryphe I've played around with this a little trying to work with `Styles` as we have them and updating and changing them in reason, and the core *issue* (a language feature not a real issue) is that non of the `Style` fields are actually falsy and trying to do a merge which overwrites all the `undefined` fields would either require an enormous pattern matching or ... :man_shrugging: (I can't see how else to do it as I essentially need to check each field of both records and if a value is the falsy/undefined one overwrite it)

Not sure where you land on this but I think a list of functions or types as discussed above is a much easier to work with api, but would involve a larger scale change (if you agree), in which case for the time being I can pass in a subset of the style props optionally for this PR and look at the bigger change separately

src/UI_Components/Input.re
Akin909: quite new to reason but could the `style` function take something like
```reason
  style(ViewStyles, [/* list specific to view styles */])
```
and have the first argument somehow :thinking: :confused: inform the type of the second, or just have different style functions for the elements that are different. Although so far we don't really have that kind of variation in what nodes expect, also our current api allows for that i.e. `Style.make(// all possible style arguments`

Akin909:
@bryphe I've made some tweaks to this in response to your feedback and commented a bit above about the why I've gone this way a bit more. Lemme know what you think

examples/Examples.re
```diff
@@ -97,11 +97,18 @@ let reducer = (s: state, a: action) => {
 let init = app => {
   let maximized = Environment.webGL;
 
+  let size = Monitor.getPrimaryMonitor() |> Monitor.getSize;
```
bryphe: Is it necessary to change the example size for this? I'd like to keep the size "as-is" and then showcase some capabilities like fullscreen/maximize/extra windows via #223 and #224

src/UI_Components/Input.re
```diff
@@ -0,0 +1,343 @@
...
+    ) =>
+  component(slots => {
+    let initialState = {value, placeholder, isFocused: false};
+    let (state, dispatch, slots) =
+      React.Hooks.reducer(~initialState, reducer, slots);
+
+    /*
+       TODO: Setting the hook to run only on mount means that the onChange
+       handler is only called with a stale version of state
+       BUT setting the value to ALWAYS causes one handler to to seize
```
bryphe: Interesting! This sounds like a bug. What does it mean by causes one handler to seize control of event listening?

src/UI_Components/Input.re
```diff
@@ -0,0 +1,343 @@
...
+      ~border,
+      ~value,
+      ~placeholder,
+      ~placeholderColor,
+      ~onChange,
+      (),
+    ) =>
+  component(slots => {
+    let initialState = {value, placeholder, isFocused: false};
+    let (state, dispatch, slots) =
```
bryphe: Nice use of the new hooks API!

src/UI_Components/Input.re
```diff
@@ -0,0 +1,343 @@
...
+       component
+     */
+    <Clickable
+      onFocus={() => dispatch(SetFocus(true))}
+      onBlur={() => dispatch(SetFocus(false))}>
+      <View style=viewStyles>
+        <Text style=innerTextStyles text=content />
+        /*
+           TODO:
+           1. Show and hide cursor based on focus
```
bryphe: Nice progress!

bryphe:
Sorry for the delay on reviewing this @Akin909 ! The `<Input />` control looks great to me - excellent starting point and we can continue to refine based on the `TODO`s you mentioned.

Thank you for thinking about the `Style` issue and adding a new PR / issue for it - definitely makes sense to split out - and I like the direction you're investigating 👍 

I only had one piece of feedback - can we leave the example size as-is for now? Or is there a reason it needs to be maximized for the sample?

Thanks for the work on this! It's great progress (will be very useful for Oni2 also - we need this for the command palette / quick open / etc)

examples/Examples.re
Akin909: @bryphe no worries can change that back. On my machine it opens as a really small window in the corner and when I go to maximize it crashes due to the resize bug, thought it was a "bug" that it didn't open with a larger size

src/UI_Components/Input.re
Akin909: The other handler doesn't respond at all it's almost as if only one handler is registered, I only noticed when I added the second input example basically with `Always` set the key events registered for any other component that implements a handler for the same event aren't called / don't respond.

src/UI_Components/Input.re
Akin909: :+1: The cursor bit is actually tied to having mouseover and mouseout, what I meant by the remaining comment there is that the mouse cursor currently responds to clicks but not to mousing over (although that's not actually browser behaviour so think the comment is invalid)

Akin909: I just checked and the cursor doesn't appear or disappear in a browser based on mousing over or out only as a result of clicking in or out

examples/Examples.re
Akin909: Reverted :+1: @bryphe

Akin909: Will do :+1:

PR #207, opened 2019-01-14
Integrate brisk reconciler core

bryphe:
# WIP: Replace `reason-reactify` with `brisk-core`

This change:
- Migrates our primitives to use the `nativeComponent` from brisk
- Migrates our components to the brisk component model
- Migrates our rendering logic to get the pending updates from brisk
- Migrates example

Remaining work:
- [x] Finish `Hello.re` migration
- [x] Get fix for `useState` in nested tree (https://github.com/briskml/brisk/pull/31)
- [x] Integrate `useEffect` implementation
- [x] Implement animation primitive w/ new `useEffect`
- [x] Hook up key handling in Calculator w/ `useEffect`
- [x] Find way to detect changes (so we can implement `uiDirty` primitive and only render when an update is needed)
- [x] Replace `link:` resolution with an NPM package for `brisk-core`
- [x] Update `esy.lock` w/ dependency update
- [x] Fix issue using `Hooks.effect` with condition `OnMount`


examples/Hello.re
```diff
@@ -1,149 +1,188 @@
...
-                  easing: Animated.linear,
-                },
-              );
-
-            <view onMouseDown onMouseUp>
-              <image
+/* open Revery.UI.Components; */
+
+
+module Logo {
```
bryphe: @wokalski - here's the example I was working on that had the `useState` issue.

src/UI/UiRender.re
```diff
@@ -18,15 +18,34 @@ open RenderPass;
...
 
   /* Perform reconciliation */
-  Performance.bench("updateContainer", () =>
-    UiReact.updateContainer(container, component)
-  );
+  /* Performance.bench("updateContainer", () => */
+  /*   UiReact.updateContainer(container, component) */
+  /* ); */
+
+  let latest = switch(lastUpdate^) {
```
bryphe: @wokalski - here's the code I'm using for our 'render' loop. I also temporarily hard-coded our logic to _always_ re-draw (before, we'd listen to events whenever a `useState`/`useReducer` went through, and only update then).

We're definitely missing the `RenderedElement.update` call - but my understanding is that the `useState` should queue up a pending update, which should be received via `flushPendingUpdates`, and then `executeHostViewUpdates` should apply that change to the tree.

package.json
```diff
@@ -37,13 +37,15 @@
...
+    "rejest": "^1.2.0",
+    "@brisk/core": "*"
   },
   "resolutions": {
     "@opam/cmdliner": "1.0.2",
     "@opam/js_of_ocaml": "github:ocsigen/js_of_ocaml:js_of_ocaml.opam#db257ce",
     "@opam/js_of_ocaml-compiler": "github:ocsigen/js_of_ocaml:js_of_ocaml-compiler.opam#db257ce",
-    "@esy-ocaml/reason": "facebook/reason#ab49908"
+    "@esy-ocaml/reason": "facebook/reason#ab49908",
+    "@brisk/core": "link:D:/brisk/core"
```
bryphe: @wokalski - FYI - you'll need to change this if you try out this branch!

src/UI/Primitives.re
```diff
@@ -8,83 +8,114 @@
...
+            node#setStyle(style);
+            Obj.magic(node);
+        },
+        configureInstance: (~isFirstRender as _, node) => {
+            let events = NodeEvents.make(~ref?, ~onMouseDown?, ~onMouseMove?, ~onMouseUp?, ~onMouseWheel?, ());
 
-let text =
-    (
+            /* TODO: Proper way to downcast? */
+            let tn: TextNode.textNode = Obj.magic(node);
```
wokalski: Admittedly this is problematic about this API if you don't work with abstract types.

bryphe: Definitely. We'll likely revisit the decision to use class nodes or experiment / profile some alternate approaches.

wokalski:
https://github.com/briskml/brisk-reconciler you can use this repo now!

bryphe:
Excellent, thanks @wokalski ! Just waiting on CI but I think this is almost ready 🎉 

Are you planning on publishing an NPM package, too?

wokalski:
@bryphe if it’s not problematic for others I’d rather not just yet. I would like to rename a few types which might are technically breaking changes. Also, there might be further changes to the Hooks api (I.e. maybe you’ll return hooks from render). But when we have the ppx and the naming is ok, I’ll start publishing to npm.

PR #212, opened 2019-01-18
Example - Game Of Life

tcoopman:
This is a new example for revery. The [Game Of Life](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life).

It's currently very much a WIP but I'm already creating the PR to get some feedback.

Todo:

- [x] Merge master
- [x] Remove core
- [x] Make the `Tick` action work in a loop and add `Start`, `Stop`
- [ ] Add `Faster`, `Slower` actions.
- [ ] Add a dropdown button to pick different start universes.
- [x] Make the viewport flexibel (move it up, down, left, right and zoom in,out)
- [x] Cleanup

Here are some of my current questions:

- I'm a little clueless on how to do the tick, so I'm waiting for an example
- All other feedback is very welcome, I haven't cleaned anything up yet, so that is still needed.


examples/GameOfLife.re
```diff
@@ -0,0 +1,536 @@
...
+    React.element(make(~fontFamily, ~contents, ~onClick, ()));
+  };
+};
+
+module Cell = {
+  let component = React.component("Column");
+
+  let make = (~cell, ~onClick, ()) =>
+    component((_slots: React.Hooks.empty) => {
+      let clickableStyle =
```
bryphe: Thanks for removing the border for now!

One other optimization we can make - since this `make` code path is hot for large boards, we could factor the creation of the styles out, so we don't have to recreate them every time `make` is called. This will save the garbage collector a bit of work!

bryphe:
This is a great example, @tcoopman ! It really exercises the pipeline and is a good 'stress test' as we zoom in (already identified a couple of performance bottlenecks). The performance fixes we make as a result of this will translate directly over to other apps, too.

The code looks great 👍 Cool use of the `reducer` hook + the tick + the `evolve` function. It's neat to see everything work together.

> Add a dropdown button to pick different start universes.

I love this idea - would be fun to add gliders or other primitives in too 😄 , but unfortunately we don't have a dropdown control at the moment! Would be nice to have an issue tracking adding one.

tcoopman:
* Created issue #221.
* Styles are created as much as possible outside of `make`
* Use fontawesome for the buttons (they can still use some layout work)

bryphe:
> I've switch to the default button and cleaned some things up. I'm happy enough to have this merged if none of you have any issues left

Perfect! Let's merge it. We can continue to improve and experiment with perf w/ this example.

> There are some improvements to the layout that I would like to bring, but those can be added after merging.

Sounds like a plan!

Thanks for all your work on this example, @tcoopman 👍

PR #215, opened 2019-01-19
Feature/Update Style API

Akin909:
I was inspired by @wokalski suggestion in #205 idea of a change to the style api that would change it from
```reason
<Text
  style=Style.make(
     ~color=Colors.red
     ~backgroundColor=Colors.blue
    /* a bunch of stuff */
    (),
    )
  />
```

to 
```reason
 <Text
  style=[
   fontSize(1)
   fontFamily(green)
   flexDirection(row)
   width(percent(90)) /* future plan that can come based on this */
  ]/>
```
Other examples of similar patterns are `brisk`, `bs-nice` and `bs-css`.
Current implementation so far is heavily inspired by `brisk` :laughing: 

### Advantages
The advantages of this would be a simpler, less error prone to newbie syntax (like forgetting to pass `unit` at the end of the make function) less exposure of the internal api of things like `BoxShadow` so this way it can be re-implemented and a user wouldn't have to refactor their usage of `BoxShadow.make(~specificLabels)`
Also merging styles would be easier as a list can be easily iterated and modified vs a record type

### DIFF
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -2,39 +2,36 @@ open Revery.UI;
 open Revery.Core;
 
 let defaultStyle =
-  Style.make(
-    ~backgroundColor=Colors.white,
-    ~position=LayoutTypes.Relative,
-    ~left=100,
-    ~top=100,
-    ~width=200,
-    ~height=200,
-    ~border=Style.Border.make(~width=15, ~color=Colors.red, ()),
-    ~borderHorizontal=Style.Border.make(~width=8, ~color=Colors.blue, ()),
-    ~borderTop=Style.Border.make(~width=15, ~color=Colors.red, ()),
-    ~borderLeft=Style.Border.make(~width=30, ~color=Colors.green, ()),
-    (),
-  );
+  Style.[
+    backgroundColor(Colors.white),
+    position(`Relative),
+    left(100),
+    top(100),
+    width(200),
+    height(200),
+    border({width: 15, color: Colors.red}),
+    borderHorizontal({width: 8, color: Colors.blue}),
+    borderTop({width: 15, color: Colors.red}),
+    borderLeft({width: 30, color: Colors.green}),
+  ];
 
 let innerStyle =
-  Style.make(
-    ~backgroundColor=Colors.yellow,
-    ~position=LayoutTypes.Relative,
-    ~left=0,
-    ~top=0,
-    ~width=30,
-    ~height=30,
-    ~borderHorizontal=Style.Border.make(~color=Colors.black, ~width=3, ()),
-    (),
-  );
+  Style.[
+    backgroundColor(Colors.yellow),
+    position(`Relative),
+    left(0),
+    top(0),
+    width(30),
+    height(30),
+    borderHorizontal({color: Colors.black, width: 3}),
+  ];
 
 let textStyle =
-  Style.make(
-    ~fontSize=20,
-    ~fontFamily="Roboto-Regular.ttf",
-    ~color=Colors.black,
-    (),
-  );
+  Style.[
+    fontSize(20),
+    fontFamily("Roboto-Regular.ttf"),
+    color(Colors.black),
+  ];
 
 let render = () =>
   <View style=defaultStyle>
--- a/examples/Border.re
+++ b/examples/Border.re
@@ -9,10 +9,10 @@ let defaultStyle =
     top(100),
     width(200),
     height(200),
-    border({width: 15, color: Colors.red}),
-    borderHorizontal({width: 8, color: Colors.blue}),
-    borderTop({width: 15, color: Colors.red}),
-    borderLeft({width: 30, color: Colors.green}),
+    border(~width=15, ~color=Colors.red),
+    borderHorizontal(~width=8, ~color=Colors.blue),
+    borderTop(~width=15, ~color=Colors.red),
+    borderLeft(~width=30, ~color=Colors.green),
   ];
 
 let innerStyle =
@@ -23,7 +23,7 @@ let innerStyle =
     top(0),
     width(30),
     height(30),
-    borderHorizontal({color: Colors.black, width: 3}),
+    borderHorizontal(~color=Colors.black, ~width=3),
   ];
 
 let textStyle =
--- a/src/UI/Style.re
+++ b/src/UI/Style.re
@@ -65,88 +65,6 @@ type t = {
   cursor: option(MouseCursors.t),
 };
 
-let extend =
-    (
-      style: t,
-      ~backgroundColor=style.backgroundColor,
-      ~color=style.color,
-      ~width=style.width,
-      ~height=style.height,
-      ~flexBasis=style.flexBasis,
-      ~flexDirection=style.flexDirection,
-      ~flexGrow=style.flexGrow,
-      ~flexShrink=style.flexShrink,
-      ~alignItems=style.alignItems,
-      ~justifyContent=style.justifyContent,
-      ~position=style.position,
-      ~top=style.top,
-      ~bottom=style.bottom,
-      ~left=style.left,
-      ~right=style.right,
-      ~fontFamily=style.fontFamily,
-      ~fontSize=style.fontSize,
-      ~marginTop=style.marginTop,
-      ~marginLeft=style.marginLeft,
-      ~marginRight=style.marginRight,
-      ~marginBottom=style.marginBottom,
-      ~margin=style.margin,
-      ~marginVertical=style.marginVertical,
-      ~marginHorizontal=style.marginHorizontal,
-      ~borderTop=style.borderTop,
-      ~borderLeft=style.borderLeft,
-      ~borderRight=style.borderRight,
-      ~borderBottom=style.borderBottom,
-      ~border=style.border,
-      ~borderHorizontal=style.borderHorizontal,
-      ~borderVertical=style.borderVertical,
-      ~transform=style.transform,
-      ~opacity=style.opacity,
-      ~overflow=style.overflow,
-      ~boxShadow=style.boxShadow,
-      ~cursor=?,
-      _unit: unit,
-    ) => {
-  let ret: t = {
-    backgroundColor,
-    color,
-    width,
-    height,
-    flexBasis,
-    flexDirection,
-    flexGrow,
-    flexShrink,
-    justifyContent,
-    alignItems,
-    position,
-    top,
-    bottom,
-    left,
-    right,
-    fontFamily,
-    fontSize,
-    transform,
-    marginTop,
-    marginLeft,
-    marginRight,
-    marginBottom,
-    margin,
-    marginVertical,
-    marginHorizontal,
-    borderTop,
-    borderLeft,
-    borderRight,
-    borderBottom,
-    border,
-    borderHorizontal,
-    borderVertical,
-    opacity,
-    overflow,
-    boxShadow,
-    cursor,
-  };
-  ret;
-};
-
 let make =
     (
       ~backgroundColor: Color.t=Colors.transparentBlack,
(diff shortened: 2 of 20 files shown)


src/UI/Style.re
```diff
@@ -287,3 +287,117 @@ let toLayoutNode = (s: t) => {
...
+  | `BorderRight(Border.t)
+  | `BorderBottom(Border.t)
+  | `Border(Border.t)
+  | `BorderHorizontal(Border.t)
+  | `BorderVertical(Border.t)
+  | `Transform(list(Transform.t))
+  | `Opacity(float)
+  | `Boxshadow(BoxShadow.properties)
+  | `Cursor(option(MouseCursors.t))
+];
```
wokalski: This also gives you more room when it comes to accepting props. For instance it doesn't make sense for View to accept the `FontFamily prop!

examples/Border.re
```diff
@@ -2,39 +2,36 @@ open Revery.UI;
...
-    ~width=200,
-    ~height=200,
-    ~border=Style.Border.make(~width=15, ~color=Colors.red, ()),
-    ~borderHorizontal=Style.Border.make(~width=8, ~color=Colors.blue, ()),
-    ~borderTop=Style.Border.make(~width=15, ~color=Colors.red, ()),
-    ~borderLeft=Style.Border.make(~width=30, ~color=Colors.green, ()),
-    (),
-  );
+  Style.[
+    backgroundColor(Colors.white),
```
bryphe: I'm on board with this. Looks pretty concise and easy-to-understand, compared to where we came from?

bryphe: I also like that this would more easily support the style 'inheritance' that we really need for custom components.

examples/Border.re
```diff
@@ -2,39 +2,36 @@ open Revery.UI;
...
+    position(`Relative),
```
bryphe: What's the advantage of using the polymorphic variant for `'Relative`? My understanding is that the main benefit is that you can extend them - but it comes at a price - the compiler might not catch errors or typos as easily (ie, `position('Reltive)` Some comments here: http://2ality.com/2018/01/polymorphic-variants-reasonml.html https://v1.realworldocaml.org/v1/en/html/variants.html

I would not expect we would 'extend' the `Relative` positioning, so the benefit of using polymorphic variants isn't clear to me - does it save anything syntactically?

src/UI_Components/Button.re
```diff
@@ -3,43 +3,44 @@ open Revery_Core;
...
       (),
     ) =>
   /* children, */
   component((_slots: React.Hooks.empty) =>
     <Clickable onClick={disabled ? noop : onClick} ?onFocus ?onBlur ?tabindex>
-      <View style={containerStyle(~width, ~height, ~disabled, ~color)}>
-        <Text style={textStyle(~fontSize, ~fontFamily)} text=title />
+      <View
+        style=Style.[
+          position(`Relative),
```
bryphe: Thanks for converting everything to this new paradigm!

test/UI/StyleTest.re
```diff
@@ -0,0 +1,65 @@
+open Rejest;
+open Revery_UI;
+open Revery_Core.Colors;
+open Style;
+
+test("Style API tests", () => {
```
bryphe: Nice! Thank you for adding some tests.

Do we have any tests exercising the 'inheritance' behavior? Like if I do something like `Style.inherit([style1, style2, style3])`, what the expected outcome is?

src/UI/Style.re
bryphe: Cool!

src/UI/Style.re
```diff
@@ -287,3 +287,172 @@ let toLayoutNode = (s: t) => {
   };
   ret;
 };
+
+/* -------------------------------------------------------------------------------
+        Styles: As a list of Polymorphic variants
+   -------------------------------------------------------------------------------*/
+
+type styleProps = [
+  | `Position(LayoutTypes.positionType)
```
bryphe: Following from above - what's the benefit of using polymorphic variants here? Do regular variants not support this?

src/UI_Components/Clickable.re
```diff
@@ -38,17 +41,23 @@ let make =
       onClick();
     };
 
-    let style2 =
-      Style.extend(style, ~opacity, ~cursor=MouseCursors.pointer, ());
+    /* let style2 = */
+    /*   Style.extend(style, ~opacity=animatedOpacity, ~cursor=MouseCursors.pointer, ()); */
```
bryphe: Ah, I guess the 'inheritance' behavior hasn't been implemented yet - that piece would be nice to have with this PR (since it's one of the benefits of moving to this new model!)

src/UI/Style.re
bryphe: It sounds like one benefit is that you don't need to have the module prefixed with it, which is nice!

And the typos can be avoided using `styleProps` in the type annotation, it seems.

examples/Border.re
bryphe: Read more and answered my question below 😄

bryphe:
Looks great to me @Akin909 ! Thanks for the feedback / help @wokalski and @rauanmayemir as well. It would be nice to move to a similiar API to other 'native' UIs, and I see the advantages of this new proposal.

I'm on board - I think we just need to implement the 'merge' / 'inherit' behavior (and wire it up in the places that were using `Style.extend` previously) - otherwise I'm good to go on this 👍

examples/Border.re
```diff
@@ -2,39 +2,36 @@ open Revery.UI;
...
-    ~left=0,
-    ~top=0,
-    ~width=30,
-    ~height=30,
-    ~borderHorizontal=Style.Border.make(~color=Colors.black, ~width=3, ()),
-    (),
-  );
+  Style.[
+    backgroundColor(Colors.yellow),
+    position(`Relative),
```
wokalski: Variant as arg ☝️

examples/Border.re
```diff
@@ -2,39 +2,36 @@ open Revery.UI;
...
+    left(0),
+    top(0),
+    width(30),
+    height(30),
+    borderHorizontal({color: Colors.black, width: 3}),
```
wokalski: Record as arg ☝️

tcoopman:
> Also, I'd like to know what you think about styles composition. Lists are nice but appending lists is O(n) which feels like a waste for such operation. We could create a tree where prepending would take constant time. We could still use list syntax [like in this snippet](https://gist.github.com/wokalski/46662bd6e4fbeb3ad3223c25165986ee#file-hooksheterogenouslist-re-L1-L3)

Just a question when I read your remark. If something like this is actually changed, shouldn't there be some benchmarks to validate that the change is actually improving things?

wokalski:
@tcoopman of course - benchmarks would be great but we're on the edge here with extremely scarce resources. I have a similar situation in the reconciler where I'd love to have some micro benchmarks already to have some sense of the impact of my changes but I don't. Also - I don't think this part of styling is significant in terms of performance  or maybe there are situations where it is but it's probably not easy to measure. My comment is kind of more about the culture of what we do. I'd like us to always keep in mind the performance characteristics because small things do add up!

wokalski:
@tcoopman good point!

examples/Border.re
Akin909: @wokalski thanks for the feedback :pray: its much appreciated. Re. the records, I'm not sure I understand

> the fields width and color won't be qualified (from the type system's point of view.)

The function is typed and the compiler errors if the wrong type is passed or any field is missing :confused:, my thinking was the syntax for a record would be more easily recognisable to anyone new to reason, and in the layer beneath a record is expected and since they seemed to be 2 similar solutions :man_shrugging:, if there is something lost by doing it with a record though I'm happy with changing it

OhadRau: Another advantage of not using records is that you don't have to open the `Style` module and you can make optional parameters. To get optional parameters you'd have to do something like `borderHorizontal({...defaultBorder, color: Colors.black, width: 3})` which would be kinda annoying.

examples/Border.re
Akin909: Re. this one having had a look at `brisk`, `bs-css` and `bs-nice` it looked almost like this was the general pattern i.e. variants as args I think an advantage of this is that if there is well documented example then its quite intuitive (maybe not immediately clear what a variant is etc.) but you can start to see what you might need to pass in and how.

I think constants would require a user knew what needed to be `open`end (just `Style` probably), from my initial foray into reason I encountered that using `bs-nice` i.e. polymorphic variant as arg didn't exactly know what the syntax meant but was able to pretty easily guesstimate the pattern and the compiler errors also point towards the types of arguments so its easy to pick up.

If it might be an anti-pattern in reason/ocaml again happy to change although, in this vs the case of records I feel like this pattern is more approachable than finding the appropriate constant?

examples/Border.re
Akin909: happy to change to the labeled args syntax although theres quite a few variations of the `border` so don't think any of them will be optional, been trying to avoid the optional syntax as I think forgetting to pass unit is a pretty common issue people will hit with functions like that at least at the beginning almost offsetting the value of having things be optional.

test/UI/StyleTest.re
Akin909: @bryphe I've implemented a `merge` function which I've added tests for it doesn't follow the signature you've laid out largely because I'm not really sure how to do that, took a fair bit of trial and error to get what I did but I'm sure there's a better/more performant way than my current solution (whatever it is I don't know it)

Akin909: I could potential see how you could take a list of lists and maybe run merge on the first tow then use the result and run it on the next but tbh my current solution is not very efficient so using it that way is probably not ideal :face_with_head_bandage:

src/UI/Style.re
Akin909: this is done now `View` nodes now report a type error if you try and pass them `font` props

examples/Border.re
wokalski: To be clear, I'm not sure what's better. There are two extremes:
1. Eveything's just a plain variant
2. We have functions/constants for everything.

In terms of code between this:
```
Style.[
    backgroundColor(Colors.yellow),
    Position.relative, (or /*position(relative)*/)
    ...
```
and this
```
Style.[
    `BackgroundColor(Colors.yellow),
    `Position(`Relative),
    ...
```

My case for the everything is a functional API is:
1. you can have optional args
2. Better documentation

Btw, records had the same "problem" (it's also an advantage) that you have to open the module which contains the records type.

Akin909:
@bryphe any chance we can hold off or at least slowdown style related PRs till this is in, every time theres a new style added, I have to rewrite all the associated styles :cry: its mega time consuming :rofl:

test/UI/StyleTest.re
bryphe: I think its OK to start with what we have here, and then improve. We need proper tools to 'benchmark' anyway before we can truly optimize aggressively!

bryphe:
Thanks for your work on this, @Akin909 ! Is this good to go, or is there any work remaining?

wokalski:
Looks good to me 👍

Akin909:
@bryphe I think given the size of this PR already this is a good place to finish :+1: , but I have a follow up additional change I was going to add to simplify some of the functionality of working with list of styles, specifically around being able to `unwrap` a style e.g. in the input if a user passes a height of x then simplify the process of getting the value which at that point is wrapped in a variant but that can definitely come separately

bryphe:
@Akin909 - sounds good! Let's bring this in so you don't have to deal with anymore merge craziness 😆 and then we can track the remaining issues as follow-up items and PRs.

Thanks for all the hard work on this! And thanks all for the reviews and feedback!

jordwalke:
What about the additional allocations here? (I'm assuming there are some)

bryphe:
@Akin909 - I believe @jordwalke means memory allocations (allocations that can impact the GC). This can be troublesome for performance 

We definitely need tooling to validate this - I just logged an issue to track - #240. But I think what will be important is to come up with some 'benchmark' scenarios that we can test and measure any timing / allocation impact.

PR #225, opened 2019-01-20
Add Window.takeScreenshot API

### LINKED ISSUE #185: API: Window.takeScreenshot
Building off the excellent work that @OhadRau did in bryphe/reason-glfw#72 - exposing the `glReadPixels` API + screenshots in all the platforms, I think it makes sense to have this be a first-class API on our `Window` object.

## Proposal

We add a `Window.takeScreenshot(w: Window.t, fileName: string)` API.

This would:
- At the end of the next render frame, call `glReadPixels` into a new `Image`, and call `Image.save`. These would be synchronous operations

This code that @OhadRau added in the reason-glfw example basically does all of this already - just about wrapping it up in an API:

```
  let captureWindow(w, filename) {
    let size = glfwGetFramebufferSize(w);
    let image = Image.create(~width=size.width, ~height=size.height,
                             ~numChannels=4, ~channelSize=1);
    let buffer = Image.getBuffer(image);
    glReadPixels(0, 0, size.width, size.height, GL_RGBA, GL_UNSIGNED_BYTE, buffer);
    Image.save(image, filename);
    Image.destroy(image);
  };
```

## Use Cases

- For Oni 2 - I'd love to have this to make it easy to snap a screenshot of the editor! IE, bound a key press to 'take screenshot', and have a configurable place for screenshots to be stored
- For Revery - this would be a simple API for #156 to use for automated testing

> __NOTE:__ The proposed implementation would save the image synchronously, so it wouldn't be something used in performance-critical code. But for the use cases outlined above - that's not an issue.

OhadRau:
Great news -- took another quick look at this and figured out what I was doing wrong! Was getting some build errors after I rebased this but I think that's just due to some weird build artifacts since it had to do with fonts being duplicated for some reason... (Fixes #185)


OhadRau:
Realized I really fucked up my git history before somehow but just implemented this in the example with a "Take screenshot" button. Looks to be working great on my end but would appreciate other people trying this out on their PCs (& especially the WebGL version in different browsers/OSes).

Btw, one thing I noticed (which I don't think is related to this) is a really big decrease in performance when I maximize the examples window on my laptop. Takes the FPS from ~70 to ~15 and looks very very choppy. Looks like we're still very far away from the performance we're shooting for :/

src/Core/Window.re
```diff
@@ -323,6 +323,26 @@ let getDevicePixelRatio = (w: t) => {
...
+let takeScreenshot = (w: t, filename: string) => {
+  open Glfw;
+
+  let width = w.framebufferWidth;
+  let height = w.framebufferHeight;
+
+  let image = Image.create(~width, ~height, ~numChannels=4, ~channelSize=1);
+  let buffer = Image.getBuffer(image);
+
+  /* WebGL is weird in that we can't capture with glReadPixels during
```
bryphe: Thanks for the comment here! I was wondering why we needed the extra `render` call.

examples/Examples.re
```diff
@@ -116,7 +133,14 @@ let init = app => {
       />;
     };
 
-    let buttons = List.map(renderButton, s.examples);
+    let buttons = [
+      <ExampleButton
```
bryphe: The UX for the example is confusing to me. All the other examples behave as sort of a 'tab' - you click the example, and the example is rendered. The 'Take Screenshot' behaves differently - it takes a screenshot when you press it, and it always looks highlighted.

Instead - could we have the 'Take Screenshot' example be like the other examples, but have a 'Take Screenshot' button - similiar to the `DefaultButton` example, but instead of just a counter, it would actually run the screenshot code. 



When that button is clicked, it would show the path to where the screenshot is saved. Right now - there is no feedback when the 'Take Screenshot' button is pressed, so it's a bit confusing.

I'm open to other ideas too to showcase this. But the current behavior isn't clear about what's happening - I want to make sure that we show the user where it is saved, so they know something happened. In WebGL - it's not too big of deal, since you see the download pop-up. But for native, it's important to know that something actually happened and where to find the screenshot for inspection.

tcoopman: Just an idea, but wouldn't it be fun to show the screenshot as an image in the example itself? Of course the path would be important too.

It would also be fun to take screenshots from the other examples, like from the game of Life :-). Not sure how to make that into a friendly UI

bryphe:
Thanks @OhadRau for integrating the screenshot API w/ Revery! The `Window.takeScreenshot` API looks great 👍 

I found the UX for the example a bit confusing (it broke the convention of the other examples, and it wasn't clear that anything happened in native when clicking 'Take Screenshot') - I left more specific feedback in the comments. Let me know what you think.

> Btw, one thing I noticed (which I don't think is related to this) is a really big decrease in performance when I maximize the examples window on my laptop. Takes the FPS from ~70 to ~15 and looks very very choppy. Looks like we're still very far away from the performance we're shooting for :/

Could you please log an issue for this? It would be helpful to know the exact environment - I think you mentioned you're on a SurfaceBook 2 + WSL? It would also be helpful in case other people are running into this.

In addition - if you run with the `REVERY_DEBUG=1` environment variable, we do some perf logging - it can help us see where the bottleneck might be.

I tried to reproduce this (I have a SurfaceBook 1), and I tried a bunch of different things - I put it on Battery Saver mode, etc and ran the app fullscreen - it stayed at a smooth 60 FPS for me.

However, one thing I found, was that, it seems in WSL, a lot of people have issues where it's using _software OpenGL_ instead of hardware drivers. Some references:
- https://github.com/Microsoft/WSL/issues/2855
- https://github.com/Microsoft/WSL/issues/637
- https://superuser.com/questions/1238925/use-accelerated-graphics-on-windows-10-linux-subsystem-with-xming

I'd be curious if you've done anything to set up hardware GPU drivers or LibGL pass through to the host - otherwise, it's likely software emulation the OpenGL APIs is taking place, and that would be _very_ slow on a high res display like the SurfaceBook 2. 

Another thing to try would be building & running Revery outside of WSL, just via Win32 command - that should have better driver support and perform better than a WSL build. Curious if that's better!

examples/Examples.re
bryphe: Cool idea @tcoopman !

OhadRau: Little confused about what you mean. Do you want a whole example tab for the screenshot button? I added this as a separate button so that you can take screenshots from any tab, so I want to keep the button globally accessible if possible. But I do agree that the UX isn't great right now. In terms of feedback, I could try to do a modal using a second window to spice it up. This could also preview the image in that window I guess.

bryphe: > Do you want a whole example tab for the screenshot button?

Yes, this is what I was thinking. For the example app, its more about showing the capabilities than anything else - so I think this is reasonable. Longer-term, I want to publish the examples to a website, and as you click on the tabs, it'd show the source code for the selected tab next to the example app - so in this case, they could click the 'Screenshot' tab, and see how to use the `Window.takeScreenshot` API. 

> I added this as a separate button so that you can take screenshots from any tab, so I want to keep the button globally accessible if possible. But I do agree that the UX isn't great right now. In terms of feedback, I could try to do a modal using a second window to spice it up. This could also preview the image in that window I guess.

If you prefer this, we could add a 'button section' at the bottom of the left pane and put the screenshot button there. At least differentiate it so that it is clear that it triggers an 'action' and its not a 'tab'.

OhadRau: Ok thanks for the feedback. Any tips on how I can get a reference to the current window in the other component (needed to call the Window.takeScreenshot function)? I guess I would have to pass it in as a prop of some kind...

bryphe: Ya actually a few of the examples need the window anyway - it gets passed as a parameter to the render function, like:
```
    {name: "Input", render: w => InputExample.render(w)},
```

And then in the `InputExample.re`, it passes the window as a prop: 
```
let render = window => <Example window />;
```

OhadRau:
@bryphe I'm having some trouble loading the images that are dynamically generated. I've tracked down the issue to how the paths are generated when loading image files in ImageRenderer.re. Saving the file follows the `Unix.getcwd()` result (directory you run `esy x Examples` from) whereas loading uses the `Environment.getExecutingDirectory()` result (directory where the exe is located). Because this is automatically prepended to the path of the image, it's not enough to just give the absolute path of an image, instead you'd have to pass something like `../../..` if you wanted to go up to a different directory. I don't think this makes for a great UX here but I'm not sure what the optimal solution is here, maybe it would make sense to check multiple paths? Note this is broader than just this PR specifically, since you wouldn't be able to do `<Image src="~/.myapp/res/img.png"/>` or `<Image src="/var/myapp/res/img.png"/>`, you'd have to expand that into a relative path from the executing directory.

bryphe:
I created a PR to check if the path is absolute in #242 . Not sure if we can handle this show-screenshot case well in WebGL (I guess we could create a [blob URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) for it) - but I'm OK making that piece only-native for now.

OhadRau:
@bryphe Changed the example to be just another tab and display the image. Let me know if there's anything else you think I should change.

image

examples/Examples.re
```diff
@@ -31,6 +31,7 @@ let state: state = {
     {name: "Focus", render: _ => Focus.render()},
     {name: "Stopwatch", render: _ => Stopwatch.render()},
     {name: "Input", render: w => InputExample.render(w)},
+    {name: "Screen Capture", render: w => ScreenCapture.render(w)},
```
tcoopman: wouldn't it be useful to put the Screen Capture button at the bottom of the screen? For example with something like this:

```reasonml      <View
        style=Style.[
          position(`Absolute),
          top(0),
          left(0),
          width(175),
          bottom(0),
          backgroundColor(bgColor),
        ]>
        ...buttons
      </View>

      <View
        style=Style.[
          position(`Absolute),
          bottom(0),
          left(0),
          width(175),
          backgroundColor(bgColor),
        ]>
        screenCaptureButton
      </View>
```

OhadRau: Hm, do you want me to go back and do that? Right now this is setup to be in its own tab as bryan pointed out. This is a little weird because you can't get screenshots of everything but it's more of a proof of concept than anything else so I'm not too concerned about it.

bryphe: I'm OK with it is, for now - what I'd like to do with the example app is host it on a website, and show the example code for the tab side-by-side with the selected example. 

For the screenshot example - it'd be nice to see how easy it is to implement this behavior w/ that code snippet alongside. We can always revisit down the road though if we want to shuffle it around.

bryphe:
Just tried it in both native and JS - looks great @OhadRau ! Thanks for the work on this 👍

OhadRau:
Awesome guess I'll merge this in then!

PR #232, opened 2019-01-22
Fix #180 - Crash when shrinking window

### LINKED ISSUE #180: Resizing window seems to cause division by zero

bryphe:
__Issue:__ When shrinking the window (either by dragging to make the window smaller, or pressing the 'restore' button after maximizing), the app would crash with a division-by-zero error.

__Defect:__ During resize, there are some transient states with the resize where we calculate the pixel ratio as `0.99` - the `float_to_int` is a `floor` operation, so when we run `float_to_int(0.99)`, it gets converted to an integer of 0, and then the division operations down the road fail.

__Fix:__ Leave the floating point value and defer conversion until the endpoints. For usages in finding the font size, we need to convert to an integer at that point. For text, though, we can just use the floating point value as-is.

### DIFF
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -46,7 +46,7 @@ class node ('a) (()) = {
       style.overflow,
       dimensions,
       parentContext.screenHeight,
-      float_of_int(parentContext.pixelRatio),
+      parentContext.pixelRatio,
       () => {
         let localContext =
           NodeDrawContext.createFromParent(parentContext, style.opacity);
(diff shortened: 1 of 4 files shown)


bryphe:
Fixes #180 again

lpalmes:
It would have been really useful to have a call stack/debugger. In the past i have been using https://github.com/reasonml/red but it would be nice to have a solution (maybe even using revery) to go through the stack (like XCode).

PR #235, opened 2019-01-22
[wip] Multiple Windows Support

lpalmes:
Adds support for multi window but it has some issues that needs to be fixed before merging.

Todo:
- [ ] First window steals all events
window events


OhadRau:
One more thing we want for multi-window support to work well is a way to destroy a window safely. Right now, `Window.destroyWindow` doesn't cancel any of the subscriptions to events which will slow down your code if you open and close a ton of windows.

bryphe:
Unfortunately this is out-of-date now with the migration from GLFW -> SDL2 - something we can revisit though with the SDL2 apis (I think it actually is easier to support this in SDL2, due to the way the events are structured)

PR #237, opened 2019-01-22
Update README.md

kyldvs:
Just a minor nit I noticed. The example in readme does not use consistent spacing for tabs. This switches them all to be 4 spaces.


bryphe:
Thanks @kyldvs !

jordwalke:
Should run the examples through `refmt`! (And the syntax highlighting is listed as ocaml, but `reason` is supported and renders the comments correctly).

PR #238, opened 2019-01-22
Run refmt on example

jordwalke:
Also use `reason` syntax highlighting


bryphe:
Looks great! Thank you @jordwalke - much better with the improved highlighting 😄 

I just realized we'll have to update it with the latest `brisk` integration work too.

PR #242, opened 2019-01-23
[WIP] Asset Loading: Allow absolute paths

bryphe:
Addresses an issue in #225 where absolute paths aren't loaded properly (we _always_ prepend the working directory, today!)

TODO:
- [ ] Verify in WebGL strategy
- [ ] Update lockfile
- [ ] Clean up

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@opam/lwt": "^4.0.0",
     "@opam/lwt_ppx": "^1.1.0",
     "@brisk/brisk-reconciler": "*",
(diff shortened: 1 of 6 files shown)


Akin909:
@bryphe @OhadRau maybe you could execute an echo command using `Unix.open_process_in` tested this out and the shell expands it? other than a module that would do that for you the only way to get that kind of expansion is from calling out to a shell, I don't think its necessarily a blocker her but something that can be added on top or afterwards? :man_shrugging:

PR #244, opened 2019-01-23
Convert API for box shadow to labelled args

Akin909:


examples/InputExample.re
```diff
@@ -50,13 +50,13 @@ module Example = {
...
             color(Colors.white),
             margin(20),
-            boxShadow({
-              xOffset: (-5.),
-              yOffset: 2.,
-              color: Colors.black,
-              blurRadius: 20.,
-              spreadRadius: 0.,
-            }),
+            boxShadow(
```
bryphe: Cool! That's neat that we don't need `()` at the end.

src/UI/Style.re
```diff
@@ -336,7 +336,8 @@ let cursor = c => `Cursor(Some(c));
 
 let opacity = o => `Opacity(o);
 let transform = t => `Transform(t);
-let boxShadow = (b: BoxShadow.properties) => `BoxShadow(b);
+let boxShadow = (~xOffset, ~yOffset, ~spreadRadius, ~blurRadius, ~color) =>
```
bryphe: Does it makes sense to set some default parameters here?

Akin909: @bryphe I was thinking about this and was thinking of css and how its implemented there and it seemed a bit too flexible to allow defaults or rather you wouldn't/can't add half of a box shadow which I think is a good design decision because whatever defaults you set would be arbitrary I think if you're going to specify a box shadow then that should involve deciding how you want this to display I think there's no real one size fits all default shadow (except maybe a black-ish color)

another slight concern is re-introducing the unit issue in the case of box shadow (which I don't wont to over-emphasise or exaggerate but would like to avoid if there isn't a very good reason)

bryphe: Sounds reasonable!

Issue #245, opened 2019-01-23
UI - Component: Radio Buttons


bryphe:
We need a concept of 'Radio Buttons' in Revery: <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio>

These should be doable with the current set of primitives we have today. Each Radio Button could be represented by a `<ClickableOpacity />`, and include a glyph (powered by font-awesome - there is a 'dot-circle' glyph that would be perfect for the 'selected' state: https://fontawesome.com/icons/dot-circle?style=regular  and an open circle for the 'unselected' state: https://fontawesome.com/icons/circle?style=regular)

The biggest challenge with these is figuring out the right way to represent them. This React component - `react-radio-buttons`: https://github.com/mu29/react-radio-buttons#readme uses a `<RadioGroup />` and nested `<RadioButton />`. This approach would be a bit awkward today though since we don't have `context` at the moment (the RadioButton would need a way to communicate selection back up to the 'group' - `context` would be a convenient abstraction for that).

bryphe:
All of our current custom components live here: https://github.com/revery-ui/revery/tree/master/src/UI_Components - so that's a good starting point to check out what it looks like to create a component

schinns:
`react-radio-buttons` is a great reference! @bryphe I'll take this on!

bryphe:
Sweet! Thank you @schinns!

jeremyscatigna:
Would love to help if needed 💪

Akin909:
Think #254 closes this (let me know if I'm wrong)

PR #254, opened 2019-01-24
feature/radio-buttons

schinns:
Solution to https://github.com/revery-ui/revery/issues/245 and Radio Button Example

I implemented simplest solution possible. If we want to try a different implementation let me know!

radio-button


examples/RadioButtonExample.re
```diff
@@ -0,0 +1,53 @@
...
+open Revery.UI.Components;
+
+module RadioExample = {
+
+  let component = React.component("RadioExample");
+  let make = () => {
+    component(slots => {
+      let (radioVal, setRadioVal, _slots: React.Hooks.empty) = React.Hooks.state("Select a button!", slots);
+      <View
+        style=Style.[
```
Akin909: Don't think its part of our CI yet but think @bryphe was trying to add `refmt` as part of the process, if you have it installed you can point it at your files and it will autoformat them.
Not a blocker I don't think but would be nice to have :+1:

Akin909:
@schinns this is awesome :100: I left a couple of comments (minor ones :+1:)

schinns:
@Akin909 thanks for your review! Woops didn't mean to push up `.DS_Store`, I removed the file and added it to `.gitignore`! I also ran `refmt` on the files I created. It didn't make any changes. 🤷‍♀️(or maybe I didn't run `refmt` properly?)

examples/RadioButtonExample.re
schinns: I just ran `refmt RadioButtonsExample.re` and I saw some output in the terminal, but there were no changes made to commit.

bryphe: Ah yes it's a bit confusing - you have to actually run `refmt --in-place RadioButtonsExample.re` - otherwise it just writes the output to the console

bryphe:
Awesome @schinns , thank you for the contribution! Very cool 💯 

Sorry about the CI issue, I've kicked it off manually for now - going to open a ticket on Azure CI too though and see if we have something misconfigured. It seems like occasionally it just doesn't run for PRs.

examples/RadioButtonExample.re
schinns: ahh ok got it! thanks!

bryphe:
Just was trying it out - looks awesome @schinns ! Thank you for your work on this.

I noticed the CI kicked off this time, but it looked like there were still some formatting issues. It might be due to different versions of `refmt` - we're using the `3.4.0` reason CLI tools on the build machine.

Hope you don't mind - I formatted and committed that change to your branch so it'd be get picked up here.

Otherwise, looks great! I'll bring this in once the build is green.

bryphe:
CI is green! 🎉 Thank you @schinns - really appreciate the contribution! Bringing in now.

Issue #239, opened 2019-01-22
Add ability to query a style from a list of styles


Akin909:
This would be particularly helpful for building native components e.g. if a user passes a list of styles like
```reason
<Slider
 style=Style.[color(red)]
/>
```
and you wanted to query this in the component so you could do some logic on basis of a value a user has passed in, I'm working on an `unwrap`
function which allows you to get the raw value a user passed in since
as it actually comes down it would be wrapped in `Color(red)`

Akin909:
This involves trying to pull a specific polymorphic variant out of a list and unpacking its value :worried:

Akin909:
@OhadRau have been wrangling trying to get something like the first example so it's good to know (side note where can I find out more about *why* or rather how it does work so I can avoid such rabbit holes in the future)

re the solution you proposed @bryphe mentioned something similar in #256 (I think) and the TLDR is that a `Style.t` is a strict type and so creating one means also filling in all the blanks, some of the properties don't accept `cssUndefined` though so I'd have to set some values like in the case of our defaults so if you tried to see for example if `Style.unwrap(position)` or something you would always get a value but you wouldn't know if it was a default value or what the user set

Akin909:
not entirely sure if you were suggesting writing the values to a record actually but we I did consider creating an unwrap function for all the various types since the second snippet with the fold works, it just seemed very repetitive so been trying to figure out if there is a more generalizable solution

OhadRau:
> have been wrangling trying to get something like the first example so it's good to know (side note where can I find out more about why or rather how it does work so I can avoid such rabbit holes in the future)

Hmm not sure if I have a specific link, but that's something I spent ages trying to figure out in an old project. Essentially, there's no way to tell the type system that the constructor \`X goes with \`X(int), so we can't do something like `select(`X, [`X(0), `Y(1)])`. I believe the workaround used by the project I was looking at (OCaml FuSe) was to let the function construct an `X with a dummy value:

```
select(x => `X(x), [`X(0), `Y(1)])
```
which would allow comparison if you did:
```
switch (style) {
  `X(x) => `X(x) == f(x) /* f is x => `X(x) */
}
```
Note: This still requires us to enumerate every case.

The other issue with this was that there was no way to get it to return the actual result, since you can't access the type of that constructor. So there's no way to tell the compiler:
```
let x: [`X('a)] => 'a
```
[...]

PR #256, opened 2019-01-24
Bugfix/fix input example

Akin909:
This fixes the input component example which #205 broke, it involves having to pull values out of the style list which I wanted to create a sort of *getter* for but having done a bit of reading and wrangling with the compiler doesn't seem straight forward, typing gets a little tricky with variants will keep it in mind and revisit once I've leveled up my reason/ocaml know-how (tracked by #239)


src/UI_Components/Input.re
```diff
@@ -166,6 +166,17 @@ let make =
         style,
       );
 
+    let inputFontSize =
```
Akin909: Also tried generalising this as it seems like a much more common use case to get the specific value of an item in the style list if you want to do some internal logic based on it.

I tried something along the lines of the above but
```reason
  /* Inside a fold_left */
  switch(valueIWant, variantType) {
   | ("marginTop", MarginTop(mt)) => mt
  /* ....etc */
}
```
the compiler doesn't seem to be able to narrow down the return type this way, could have been a question type annotating it properly but I couldn't work out how

bryphe: Maybe for some of these - it'd be worth just calling `Style.applyStyle`, getting a record, and asking that for the value?

Like:
```
let { height, fontSize, color } = Style.apply(style);
```

src/UI/Style.re
```diff
@@ -342,12 +344,23 @@ let overflow = o => `Overflow(o);
 let color = o => `Color(o);
 let backgroundColor = o => `BackgroundColor(o);
 
+/*
+   Helper function to narrow down a list of all possible style props to
+   one specific to a type of component
+ */
+let rec extractViewStyles = (styles: list(allProps)): list(viewStyleProps) =>
```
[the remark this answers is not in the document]
bryphe: Good question, I'm not sure this would work either 🤔

src/UI_Components/Input.re
Akin909: this issue with that approach is that since `style.t` is strictly typed and records in general are strictly typed I can't return a record that maps to *exactly* what was passed in if I return a type `t` then I'm forced to set some default values which aren't all undefined which can muddy the waters in terms of know what was actually in the style list and just that, a potential alternative might be returning another data type like a map or a tuple, ill experiment in a separate PR

PR #257, opened 2019-01-24
update image primitive to use new style api

Akin909:
An oversight of mine while implementing #215 was converting the image primitive to use the same API :man_facepalming: I completely forgot it existed. This change updates it to match the rest of the primitives

### DIFF
--- a/src/UI/Style.re
+++ b/src/UI/Style.re
@@ -250,9 +250,11 @@ type fontProps = [ | `FontFamily(string) | `FontSize(int)];
  */
 type textStyleProps = [ fontProps | props];
 type viewStyleProps = [ props];
+type imageStyleProps = [ props];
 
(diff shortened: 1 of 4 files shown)


src/UI/Style.re
```diff
@@ -250,9 +250,11 @@ type fontProps = [ | `FontFamily(string) | `FontSize(int)];
  */
 type textStyleProps = [ fontProps | props];
 type viewStyleProps = [ props];
+type imageStyleProps = [ props];
```
Akin909: These are here because as our styling gets more nuanced I imagine these will start to grow and drift so rather than have image depend on `view` I've given it its own type so later when we have things specific to an image they can go in here

bryphe: Sounds good. Thanks @Akin909 !

Issue #226, opened 2019-01-20
Small window size on HiDPI screen


Akin909:
This was discussed a bit in #205, I'm currently working on a dell xps 15 (Manjaro Linux, gnome desktop) which has a hidpi display, the resolution is (3840 x 2160) and when opening the examples app it opens with a very small window, possibly a bug with how hidpi sizing is handled.

![examples-app](https://user-images.githubusercontent.com/22454918/51446076-de68de00-1d04-11e9-840a-fb94e53b0aad.png)

OhadRau:
Same here on my SurfaceBook 2. (Running Windows + WSL)

bryphe:
Thanks for the issue @Akin909 ! This helps clear things up a lot (wanted to know if it was _just_ window size, or the text size / scaling that was off too).

I think we'll need two fixes to fix this robustly across platforms:
- Example size - we'll bump up the example size (maybe 1/2 monitor resolution) just so it fills up the right space. This an easy fix.
- Font size / Scaling - this one will require a bit more work. We want the app to look / behave consistent across platforms - but in this case the app is obviously a lot smaller due to the pixel density. For some devices - they have different screen space size vs pixel size, which we detect with work like in #190 . However, in some high-DPI displays, like the Surfacebook or the Manjaro Linux size - it seems the window / framebuffer size are one-two-one - just extra small! We need to be able to detect that case and handle it (ie, scale up the font rendering, like we do for high DPI displays).

A couple of options for handling the scaling:
- There's a new API coming in GLFW 3.3 called `glfwGetWindowContentScale` which seems to be designed for this: <https://www.glfw.org/docs/3.3/window_guide.html#window_scale>. This was logged in glfw/glfw#898. However, GLFW 3.3 is unreleased now (we're on GLFW 3.2.1) - so we might need to find another way.

The patch here might give some ideas (we could potentially adapt): https://github.com/glfw/glfw/commit/16bf872117896fb88493403157fcd367303fd1f6 or there might be ways to get the scaling info.

It looks like the best way in 3.2.1 is to calculate the DPI with `glfwGetMonitorPhysicalSize`: https://www.glfw.org/docs/latest/monitor_guide.html#monitor_size

```
const double dpi = mode->width / (widthMM / 25.4);
```

And use this to decide the 'scaling'. The challenge would be just making sure the logic we choose is also compatible with the Mac logic (which has different screen / high DPI coordinates)

bryphe:
A few things we can start looking at:

[...]

PR #260, opened 2019-01-25
Feature/tree component

### LINKED ISSUE #246: UI - Component: Tree View
It'd be helpful to have a default 'tree-view' style component available. 

Some possible API ideas:
- react-treeview: https://github.com/chenglou/react-treeview
- react-ui-tree: https://github.com/pqx/react-ui-tree
- react-treebeard: https://github.com/storybooks/react-treebeard

We could use font-awesome to find some good icons for expand/collapsed state (or even use triangles from the border styling: https://css-tricks.com/snippets/css/css-triangle/)

Akin909:
This fixes #246 
treeview

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -39,6 +39,7 @@ let state: state = {
     {name: "Input", render: w => InputExample.render(w)},
     {name: "Radio Button", render: _ => RadioButtonExample.render()},
     {name: "Game Of Life", render: _ => GameOfLife.render()},
+    {name: "Tree View", render: _ => TreeView.render()},
   ],
   selectedExample: "Animation",
 };
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -39,7 +39,7 @@ let state: state = {
     {name: "Input", render: w => InputExample.render(w)},
     {name: "Radio Button", render: _ => RadioButtonExample.render()},
     {name: "Game Of Life", render: _ => GameOfLife.render()},
-    {name: "Tree View", render: _ => TreeView.render()},
+    {name: "Tree View", render: w => TreeView.render(w)},
   ],
   selectedExample: "Animation",
 };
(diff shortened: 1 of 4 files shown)


bryphe:
Thanks for the work here, @Akin909 ! I like the way you enabled a custom rendering strategy for the tree, that's great. The component feels very concise and easy to understand.

I'm wondering if we could tweak the examples a bit, visually?

The custom renderer is funny but it was hard for me to follow - like I see "terrible song, awful person" as a child of "awesome tunes, cool dude" and thought it was a bug 😄

For the `Show Empty Nodes` - could we showcase what a child node would look like? The 'X's were a bit confusing for me. Maybe we could title one side 'Custom Render' and the other 'Default Renderer'.

A couple other minor design feedback for the example:
- Could we wrap the entire container in a <View /> and vertically center it?
- Could we put a bit more spacing + a border or line in between them to visually differentiate the two trees?

Thanks for your work on this, Akin! Can't wait until we can use it for the explorer for Oni 2 😍

Akin909:
@bryphe thanks for the review, yeah the examples are a bit sloppy (particular the record one) was actually already making some of the changes you recommended locally so will push that once its tied up

Akin909:
Side note: making examples is pretty hard since all my sizing is completely out of whack no idea how any of these things will look on someone else's machine

OhadRau:
I guess all trees are binary trees here. Could we do it more generally with any number of children? I think that would be needed for a file browser for example

Akin909:
Seems to be the flaky CI happening here the build etc. all pass

Akin909:
could use a hand with the CI here @bryphe tried re-pushing but it still fails in CI but runs locally

bryphe:
> Side note: making examples is pretty hard since all my sizing is completely out of whack no idea how any of these things will look on someone else's machine

Definitely - we need to fix this issue - sorry it's a pain right now. We don't handle 'scaling' correctly - need to fix this using calculations from `glfwGetMonitorPhysicalSize` in #226. 

> could use a hand with the CI here @bryphe tried re-pushing but it still fails in CI but runs locally

Restarted! And green now 😄 

Code changes look good to me @Akin909 , thanks for updating the examples!

PR #261, opened 2019-01-26
Dependency: Upgrade brisk reconciler (commit 7901934)

bryphe:
This picks up @wokalski 's fix for briskml/brisk-reconciler#8 - thank you for the help & fix, @wokalski ! 💯


Et7f3:
We could then remove 
https://github.com/revery-ui/revery/blob/master/examples/Examples.re#L198-L202

Issue #53, opened 2018-11-16
Performance - Text Rendering: Use texture atlas for characters


bryphe:
Rendering text is currently very expensive in Revery, because it involves lots of context-switches to jump between shaders (this is made even worse by the fact that we currently regenerate textures every frame - but that's a separate issue).

The text rendering could be significantly improved by having a texture atlas that contains all the rendered glyphs - then, we could render a line of text in a single pass (or at least, a more minimal set of passes), as opposed to the situation today - where we always render each quad / texture in a single pass.

There's an excellent TextureAtlas implementation by @cryza in Oni here: https://github.com/onivim/oni/blob/master/browser/src/Renderer/WebGLRenderer/TextRenderer/GlyphAtlas/GlyphAtlas.ts that could be useful here 👍

Cryza:
I've started working on a Reason port of the v1 texture atlas. I'll tell you once I've come up with something reviewable! It will probably require some changes in `reason-fontkit` and `reglfw` as well, though 😅

bryphe:
@Cryza - awesome! Thank you very much for your help with this 💯

bryphe:
> It took me quite some time but I think I now have fiddled around enough in a very diffuse way so that I now have a good understanding of what we can do to get a similar approach working as we had in oni v1.

Awesome! Thanks for the update @Cryza .

> I decided to postpone that to keep this issue focused for now though. Just writing everything into the buffer as a float should be sufficient for now. We might want to revisit this later.

Sounds reasonable to me for now.

> This means for now, we need to decide:

> - Keep one render pass per glyph and stay compatible with IE, Edge and Safari
> - Optimise to one render pass per text node and drop compatibility with IE, Edge and Safari

Thanks for calling out the options here! Given that the first option sounds like it is less effort to implement for now, I'm on-board with that. Getting a 'texture atlas' gives us a big leg-up when we decide to implement more aggressive optimizations, and we can follow up with other approaches.

One thing to note - there is an extension for WebGL 1 for `ANGLE_instanced_arrays`, which I believe would add sufficient instancing support for our scenarios. It seems to be very widely supported according to webgl stats: https://webglstats.com/webgl/extension/ANGLE_instanced_arrays - so that might be a future possibility.  I logged an issue in the reason-glfw repo to pick this up eventually: https://github.com/bryphe/reason-glfw/issues/27

bryphe:
Skia integrated in #567 handles this for us, now!

PR #265, opened 2019-01-26
Introduce a glyph atlas

manuhornung:
This is supposed to be the first iteration to solving https://github.com/revery-ui/revery/issues/53

Renders all glyphs inside a `<Text>` with one render pass per glyph (just as before) but avoids uploading one texture to the GPU every time. This is achieved by copying them into a globally shared glyph atlas texture and only uploading the texture whenever new glyphs are added.

This isn't by far the final upgrade level that we can achieve with the atlas-based rendering strategy. I would like to solve some points before we merge this first PR and afterwards, we should continue improving on it.

#### TODO
- [ ] Remove the `resolutions` entries for `reason-glfw` and `reason-fontkit` once they are released with the required changes (see https://github.com/bryphe/reason-glfw/pull/87 and https://github.com/revery-ui/reason-fontkit/pull/18)
- [ ] Improve glyph layout on atlas to avoid overlaps with varying font sizes
- [ ] Improve performance by only uploading the atlas if it changed and only before the actual rendering instead of after every glyph
- [ ] Possibly use a multi-pass subpixel renderer (native) and shader (both) to avoid a visual regression in the JS version (the shader makes the glyphs a bit too thick on the JS version ATM)

#### Things to consider after this PR
- Rendering per `<Text>` element instead of per glyph by using the `ANGLE_instanced_arrays` extension
- Introducing fractional pixel offset and multiple rendered variants to improve kerning on low-dpi screens
- Performance analysis and optimisation


bryphe:
@manuhornung - nice! So cool to see it coming together. Just tried out the PR locally; great progress.

>  Remove the resolutions entries for reason-glfw and reason-fontkit once they are released with the required changes (see bryphe/reason-glfw#87 and revery-ui/reason-fontkit#18)

I just reviewed both of those and they look good! We can bring them in whenever you're ready. 

>  Improve glyph layout on atlas to avoid overlaps with varying font sizes

Cool - I noticed some artifacts (edges) between the characters - I suspect this might be related to that. We might need a pixel padding around each glyph.

> Improve performance by only uploading the atlas if it changed and only before the actual rendering instead of after every glyph

I did notice a perf regression. On native, it felt somewhat slower, and on the browser - it felt significantly slower. So we'd need this prior to checking in, so that we don't regress perf. Hopefully it isn't too hard to integrate this, though.

> Possibly use a multi-pass subpixel renderer (native) and shader (both) to avoid a visual regression in the JS version (the shader makes the glyphs a bit too thick on the JS version ATM)

Nice! I noticed the fonts seem to render differently in JS (thicker) but wasn't sure why.

And your list of next steps is 🔥 🔥 🔥 . Incredible work @manuhornung !

manuhornung:
Really nice that you took the time to check it out right away! It's great to know that it also works on Windows :)

> I did notice a perf regression.

I noticed it, too. I would have been surprised if there hadn't been one though since the current state actually _increases_ the transferred amount of data between CPU and GPU instead of decreasing it :D The Chrome profiler also measured more than 96% of CPU time being spent in `glTexImage2D.
That's due to the fact that the current logic uploads the _full atlas_ (16 MB at the moment on native, 67 MB on JS) after _each added glyph_. That's quite a bit more compared to the ~ 1.2 - 5 kB from before.

I'll see if I can split the filling of the atlas and the actual rendering into two iterations through the glyphs to reduce the uploads to one per text node. If that's not enough, we might even manage to only upload the changed fraction of the atlas using `glTexSubImage2D` instead of the complete bitmap.

> I noticed the fonts seem to render differently in JS (thicker) but wasn't sure why.

This is caused by our font shader still only considering the alpha channel of the texture while the texture itself was rendered using subpixel antialiasing. That means the left and right edges of the glyphs will be coloured in a pure shade of red or blue, respectively, with also the alpha set accordingly. We reproduce it wrongly though because we simply set all colour channels to the given colour and multiply the alpha with the texture value. Bottom line, we add two more subpixels to the edges of all glyphs, extending them by 2 * 2/3 of a pixel in some places. That's why they look thicker :)

bryphe:
Sounds like the Skia work will take care of this for us! 😍

PR #266, opened 2019-01-26
[WIP] Bugfix: harfbuzz+freetype2 build failure w/ cross-compiler toolchain

bryphe:
__Issue:__ Freetype2/Harfbuzz fail to build when the `x86_64-w64-mingw32-gcc` toolchain was in the PATH - those packages erroneously assumed if those were in the path, we must be on Windows.

__Fix:__ In the packages:
- `esy-freetype2`: esy-packages/esy-freetype2#d9b9324
- `esy-harfbuzz`: esy-packages/esy-harfbuzz#d1da4da

They both ran a 'test script' to test the compilation out. This really isn't necessary in the main build (should just be a test case). Since that test script is causing problems, removed it from the main build.

This PR just updates deps in the revery repo

### DIFF
--- a/esy.lock/index.json
+++ b/esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "eb6ca7567a413bda44466421f9331d4b",
+  "checksum": "dd6d05b25318ecbd7d9aeab4d9908bd3",
   "root": "revery@link:./package.json",
   "node": {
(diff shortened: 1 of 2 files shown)


bryphe:
Build failure is due to a bug with `cmake` caching on Windows. A fresh build should address this.

PR #267, opened 2019-01-26
[WIP] Use dune for formatting

bryphe:


bryphe:
This fixes the `esy format` command for OSX. On Windows, due to esy/esy#839, you'll need to run `esy format:windows` as a workaround.

PR #268, opened 2019-01-28
Feature/Dropdown Component

### LINKED ISSUE #221: UI - dropdown component

samatar26:
Fixes #221.

kapture 2019-02-04 at 0 19 56

I'm not entirely sure if this is the right approach 😅 , but hope it's a step in the right direction. 

In terms of the API, I'm not sure which approach would be best. Whether to have a Dropdown component with subcomponents as children, like so: 

```jsx
<Dropdown>
  <Dropdown.Item value="1" label="one"/>
</Dropdown>
```

Or the current approach where we'd have a Dropdown component and pass items in through props, or perhaps even both?   

I've had a look at the browser's implementation of `select` and found that the width of the select is equal to the longest `option`. The only thing I could think of off the top of my head to achieve something similar is rendering the longest label of all the dropdown items and getting the width of that. 

I'm sure I'm missing a few other things 😅

### DIFF
--- a/examples/DropdownExample.re
+++ b/examples/DropdownExample.re
@@ -13,6 +13,8 @@ let containerStyle =
     justifyContent(`Center),
   ];
 
+let textStyle =
+  Style.[fontFamily("Roboto-Regular.ttf"), fontSize(20), marginBottom(20)];
 /*    TODO: Allow user style overwrite for Dropdown component
  */
 
@@ -27,9 +29,15 @@ module Example = {
   ];
 
   let make = () =>
-    component((_slots: React.Hooks.empty) =>
-      <View style=containerStyle> <Dropdown items /> </View>
-    );
+    component(slots => {
+      let (selectedItem, setSelectedItem, _slots: React.Hooks.empty) =
+        React.Hooks.state(List.nth(items, 0), slots);
+
+      <View style=containerStyle>
+        <Text style=textStyle text={"Selected Item: " ++ selectedItem.label} />
+        <Dropdown items onItemSelected={item => setSelectedItem(item)} />
+      </View>;
+    });
 
   let createElement = (~children as _, ()) => React.element(make());
 };
--- a/examples/DropdownExample.re
+++ b/examples/DropdownExample.re
@@ -1,0 +1,37 @@
+open Revery.UI;
+/* open Revery.Core; */
+open Revery.UI.Components;
+
+let containerStyle =
+  Style.[
+    position(`Absolute),
+    top(0),
+    bottom(0),
+    left(0),
+    right(0),
+    alignItems(`Center),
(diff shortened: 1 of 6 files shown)


Akin909:
@samatar26 if there's an easy way to do both I think that would be great, since if it can only be passed down as props but say you had a drop down that had a custom component for one item then passing via props, is quite restrictive

examples/DropdownExample.re
```diff
@@ -0,0 +1,37 @@
...
+    position(`Absolute),
+    top(0),
+    bottom(0),
+    left(0),
+    right(0),
+    alignItems(`Center),
+    justifyContent(`Center),
+  ];
+
+/*    TODO: Allow user style overwrite for Dropdown component
```
Akin909: the input container might be a good place to look for how to do overwriting styles, I'm currently trying to make the process of selecting styles a user passed in more easily accessible but for now the input is the best example

bryphe:
Very cool @samatar26 ! Great start.

Agree re @Akin909 's feedback - it'd be ideal for it (by default) to stretch to accommodate the largest item. I believe this is how it behaves in the browser:
image

A few other ideas for it:

- Could we add a down arrow icon on the right side of the dropdown? This will help to convey that it is a dropdown, and not a button or some other element.  We could use the caret-down font awesome icon: https://fontawesome.com/v4.7.0/icon/caret-down

- Could we expand the example to show the currently selected item as a label above the dropdown? We might need to add `onItemSelected`  or similiar callback from the dropdown component to notify the consumer that there is a new item selected.

Thanks for the work on this!

samatar26:
@bryphe @Akin909 I've updated the GIF in the first comment with the current state of the PR. I've got an issue where when trying to get the longest item's width using either `FontRenderer` or just creating a new textNode and using `Layout`, that it initially renders two dropdowns. 

One with 0 width and one after the measurements have been calculated. Clicking on the dropdown triggers a rerender and removes the 0 width dropdown, but the first label always seems to be renderer regardless. Any idea where I should look? Thanks for help!

bryphe:
Hi @samatar26 - is there anything left on this? Would you like help with the merge? 

Looks good to me, if you're ready 👍

samatar26:
Hey @bryphe, It's still missing the ability to add `<options>` as subcomponents to the `<Dropdown/>`, I  think I was struggling to find a way to access the children's props. I guess it's something I could also add in a future PR. Other than that, I think I can/should clean up the way the the dropdown's rendering it's children.

samatar26:
@bryphe I think you're right in that it looks good, should be ready to merge👍

bryphe:
@bryphe I think you're right in that it looks good, should be ready to merge👍

Excellent! Thank you @samatar26 for the contribution - merging now! 😄

PR #270, opened 2019-01-28
Update README.md

mozmorris:
Ubuntu 18.04 also required the extra packages as mentioned under WSL, consolidating into one place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ No additional dependencies needed.
 
 #### For `Windows` Subsystem for Linux (`WSL`)
 
-Follow instructions above for specific packages for the Linux distribution being used.
+Make sure to check the specific requirements for your Linux distribution, above.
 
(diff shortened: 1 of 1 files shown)


mozmorris:
I don't thinks so myself.

Having run both environments myself, I would always look for the Ubuntu/Linux instructions when doing anything WSL related. And then usually for a "caveats/notes" section regarding anything WSL specific. In this case, the only addition would be the notes about hardware acceleration.

bryphe:
Thanks @mozmorris for the PR and clarifying these instructions!

I added an additional Note under WSL:
```
> Follow instructions above for specific packages for the Linux distribution being used.
```

Sorry @OhadRau - I missed your suggestion when I made that tweak - agree with adding the note though 😄  I'll reword it - your suggestion was a bit more succinct.

bryphe:
Bringing it in now - thanks @mozmorris for the PR and @OhadRau and @Akin909 for the feedback 👍

PR #274, opened 2019-01-28
Feature/style getter

### LINKED ISSUE #239: Add ability to query a style from a list of styles

Akin909:
@OhadRau I've tried using a GADT as you suggested on discord for the getter, it solves the type issues :+1: but It doesn't match the right value, I tried the example you posted, which does work but once I plug in the actual style values it doesn't catch the right value.

Does anything come to mind from how I've implemented it as to why?
if this works it fixes #239


Akin909:
Oh I think it might be because the defaults arent from the selector module so they are essentially, not the same so the comparison is failing

OhadRau:
@Akin909 any examples of values that don't work? Or is it not working for anything? The type checker should hopefully catch it if we're returning the wrong thing

Akin909:
@OhadRau none of them work atm its odd because all the types are what's expected, I tried running your example with the values as is and that works but plugging in the values from `allProps` which is the combination of all style prop variants doesn't seem to match only the default clause at the bottom of the switch is hit its almost as though the variants dont match

OhadRau:
@Akin909 looks like the issue is just that the input example doesn't provide a fontfamily so it's always going to fail. Editing that file worked on my end, if you want I can push the code that I used for that.

Akin909:
@OhadRau completely right :man_facepalming: * 20 thanks for checking it out, was just a case of testing the an absent value which is why I kept getting the default SIGH :laughing:, will add some tests to this and replace the boilerplate we've been using to solve this problem

bryphe:
Thanks @Akin909 for implementing this + tests, and @OhadRau for the help and review! The Selector does look a lot simpler to use 👍

Issue #250, opened 2019-01-23
Native Integration: Menu API


bryphe:
Like Electron's [`Menu` API](https://electronjs.org/docs/api/menu), it'd be very useful from Revery to create native application menus.

The Menu API from electron is a good starting point, but I wonder if we could use `<JSX />` syntax to create a more React-like experience for creating these menu items. In particular, one thing I found awkward about the Electron API was updating menu items. I believe we could streamline this experience and simplify the API a bit.

wokalski:
I think this should be a common library shared by brisk and revery. cc @rauanmayemir

rauanmayemir:
Indeed. You could actually depend on `brisk-macos` and have `Menu` available. Here's an example how to use them https://github.com/briskml/brisk/blob/master/tester-macos/bin/Menu.re.

They don't support updates, but that could easily be improved.

Previously, I've had them as a lib like `brisk-cocoa-menu`, but then decided to move bingings in because it sucked to maintain them separately.

bryphe:
Makes sense to share the native components out and have a consistent API surface. I was initially considering a `revery-native` library to host some of these. 

Some 'native' APIs I'd like to expose from it:
- `Menu` (tracked by this issue - and sounds like already a lot of work done on the MacOS side!)
- Accessibility (TBD)
- 'Tray' integration
- Desktop notifications
- System dialogs
- Potentially text input / IME

wokalski:
@rauanmayemir note that revery has support for all platforms

bryphe:
BTW it'd be fine to host this in another place too - if you wanted to host this in brisk, maybe a `brisk-platform` for these common elements (ie, factoring out the `Menu` API, etc) would be a good starting point. Didn't mean to imply it all has to live in Revery 😄 Just a common place for us to share some of these bindings.

It's great if the API for these pieces is transferable - I could imagine someone prototyping in Revery, and then finalizing their platform-specific apps in Brisk.

rauanmayemir:
> It's great if the API for these pieces is transferable between brisk/revery - I could imagine someone prototyping in Revery, and then finalizing their platform-specific apps in Brisk.

I think this should drive our priorities. 👍

On the one hand, Brisk is aiming to target platform-specific native apis and provide bindings for it and it would make sense for Revery to use it under the hood as an implementation detail. But on the other hand, we ourselves as Brisk would probably prefer higher-level api to be the same for end-users.

[...]

Issue #249, opened 2019-01-23
Native Integration: 'dialog' API


bryphe:
This issue is tracking the integration of a 'dialog' API for displaying native system dialogs for opening, saving files, alerting, etc.

As an inspiration, the Electron API would be a great start: https://electronjs.org/docs/api/dialog

This will require both creating Reason stubs and then using the C FFI to call the appropriate APIs for each platform.

bryphe:
This lib looks like an interesting one to provide a consistent API for cross-platform file-picker dialogs: https://github.com/mlabbe/nativefiledialog

OhadRau:
Do we need to pull in another dependency for this? It doesn't seem like an enormous amount of code, so we might just want to write our own API for this from scratch.

PR #276, opened 2019-01-29
Native Integration: Platform alert dialog

bryphe:
This is a proposal for how we could (potentially) implement a native layer for Revery, re #250 #249 

This adds a `Revery_Native` module (which actually talks to the native Win32/Cocoa APIs). An open question is how we'll implement support on Linux platforms - what's the best way to handle this in X11? I see some apps "roll their own" platform dialogs, which seems pretty intense. I'm leaving this stubbed out for now, but perhaps someone more knowledgeable than me in X11 can help out.

This is what the 'native' example looks like on Windows:
native alert

Also works on OSX (Cocoa) and JS with these changes. The Linux/X11 strategy is just a stub right now.

This depends on the work done in bryphe/reason-glfw#89 to expose a `glfwGetNativeWindow` API, which gives us access to the platform-specific window.

There is also a wrapper on top - these are accessible via the `Revery.Platform` API. The reason for the indirection is that, if we need to implement these primitives on the Revery side, we could check if the functionality is available (ie, via `Revery_Native.Dialog.alertSupported()`), and then, if it isn't, create a Revery-based fallback.

There's also some messiness in terms of how we cast from the `value` -> `void *` -> whatever platform-specific pointer is being used (`HWND`, `NSWindow *`, `unsigned long int`, etc..).

I was hoping I could conditionally include `c_names` based on the platform - but it turns out this is not possible. My current approach uses a lot of `#ifdef` platform detection, which isn't really ideal. This is what it looks like now: 
```
  CAMLprim value
  revery_alert(value vWindow, value vMessage) {
      CAMLparam2(vWindow, vMessage);
      const char *szMessage = String_val(vMessage);
      void* pWin = (void *)vWindow;

  #ifdef WIN32
      revery_alert_win32(pWin, szMessage);
  #elif __APPLE__
      revery_alert_cocoa(pWin, szMessage);
  #else
      printf("WARNING - Not implemented: alert");
  #endif
      return Val_unit;
  }
```
But I'm sure there's better ways to do this...  Some ideas were:

- Implementing the platform APIs as sub-packages
- Virtual modules to implement the back-ends

I like both of these ideas, but not sure how to implement them.

Once we've iterated on this API surface a bit, we could look at moving it out to a shared library like `revery-native` that `brisk` could use (if the APIs are helpful / reusable at all) cc @wokalski @rauanmayemir

### DIFF
--- a/Revery_Native.opam
+++ b/Revery_Native.opam
@@ -1,0 +1,7 @@
+opam-version: "1.2"
+version: "dev"
+maintainer: "bryphe@outlook.com"
+author: ["Bryan Phelps"]
+build: [
+
+]
--- a/esy.lock/index.json
+++ b/esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "2f7d5682e52b8533cfa0cd393ad506b0",
+  "checksum": "172de2e0481d422d5d08539ac434ba67",
   "root": "revery@link:./package.json",
   "node": {
     "revery@link:./package.json": {
@@ -9,7 +9,7 @@
       "source": { "type": "link", "path": ".", "manifest": "package.json" },
       "overrides": [],
       "dependencies": [
-        "reason-glfw@github:bryphe/reason-glfw#d6c5f44@d41d8cd9",
+        "reason-glfw@3.2.1013@d41d8cd9",
         "reason-gl-matrix@0.9.9302@d41d8cd9",
         "reason-fontkit@2.0.6@d41d8cd9", "ocaml@4.7.1003@d41d8cd9",
         "flex@1.2.2@d41d8cd9", "@opam/lwt_ppx@opam:1.2.1@db1172a7",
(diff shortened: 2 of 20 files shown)


OhadRau:
We could try https://github.com/janestreet/ppx_optcomp for conditional compilation.

tcoopman:
Dune also has support for conditional compilation. You can see an example here https://github.com/mirage/irmin-watcher/blob/master/src/dune

Unfortunately esy doesn't support it yet https://github.com/esy/esy/issues/777

For the Linux dialogs, I would think that picking gtk or Qt dialogs is the preferred choice. I'm not sure what electron supports but will check when I'm at my laptop

tcoopman:
Electron uses Gtk dialogs on Linux. This PR might be relevant as well: https://github.com/electron/electron/pull/15293.
`GtkFileChooserNative` should drop to a native dialog chooser.

bryphe:
Thanks @tcoopman - very helpful! I read about separate compilation but never saw a usage example - that linked helped clarify how it could work.

The Electron PR is useful too - following their precedent makes a lot of sense. It seems like a bit of work to get Gtk3 hooked up to our build, so I tried to sum up the work and track the remaining issues here: #278

PR #281, opened 2019-01-29
Add missing revery_native.install script

samatar26:
@bryphe After pulling the latest master branch, I wasn't able to run the examples. I think it's because of the new package you've added, so I've added it to the install list under esy in the package.json.


bryphe:
Oh man, always forget to add this - thank you for catching it & fixing it, @samatar26 !

PR #294, opened 2019-01-31
Add linux alert

### LINKED ISSUE #278: Native Integration - Linux: Implement alert API using GtkMessageDialog
Following up from #276 , which implements `alert` for OSX/Windows, this is tracking integrating a Gtk strategy for the 'alert' dialog on Linux. There's a few different pieces we'll need - right now, there is only a placeholder stub where the Linux strategy should be:
https://github.com/revery-ui/revery/blob/0beaee47dee1f9fdfa7a45d823a6f1f094790a13/src/Native/dialog.cpp#L37

@tcoopman gave some very helpful suggestions - Electron uses GTK for its Linux platform UI - helpful PR is here: https://github.com/electron/electron/pull/15293

A few considerations:

#### Dependencies

Will we need to add `libgtk-3-dev` as a dependency for our build environments?

#### Build

We have a `discover.ml` script to help populate the right C compiler settings: https://github.com/revery-ui/revery/blob/0beaee47dee1f9fdfa7a45d823a6f1f094790a13/src/Native/config/discover.ml#L17 - but we'll need to figure out the right place  for header files (`-I`) and libs (`-lgtk3`). Is there a common place where this would usually be found? Or will we need something like `pkg-config` to discover these?

#### Functionality

The code snippets here seem very useful for create a dialog:
https://developer.gnome.org/gtk3/stable/GtkMessageDialog.html

```
GtkDialogFlags flags = GTK_DIALOG_DESTROY_WITH_PARENT;
dialog = gtk_message_dialog_new (parent_window,
                                 flags,
                                 GTK_MESSAGE_ERROR,
                                 GTK_BUTTONS_CLOSE,
                                 "Error reading “%s”: %s",
                                 filename,
                                 g_strerror (errno));
gtk_dialog_run (GTK_DIALOG (dialog));
gtk_widget_destroy (dialog);
```

Once we have the proper compilation settings, we should be able to use this code as-is. I believe for now we could pass `NULL` for the parent_window, until we have a good way to go from the X11 window handle from `glfwGetNativeWindow` to some GTK type.

Akin909:
Very early WIP implementation of gtk dialog alert to fix #278 (please ignore the commented out code etc. just trying to get stuff working)

Progress so far:

No longer chokes on `#include <gtk/gtk.h>` - currently need to tweak the borrowed snippet from #278 as an example to create an alert doesn't work as is, I think

Things to note/note to self:
- [x] I have to rework the `discover.ml` so it can use the pkg_config functionality on linux but behave as normal on other systems

- [x] Fix GTK crash on opening dialog - things compile but the function to open a dialog causes a crash, note this also happens in a separate project when I try and create a dialog with the same function (might be specific to my machine or more likely a bug in the code)


src/Native/dialog.cpp
```diff
@@ -1,41 +1,42 @@
...
 #include <caml/callback.h>
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
 
 #ifdef WIN32
-  #include "ReveryWin32.h"
+#include "ReveryWin32.h"
 #elif __APPLE__
-  #include "ReveryCocoa.h"
+#include "ReveryCocoa.h"
```
Akin909: courtesy of `clang-format` :man_facepalming: (I'm hoping its formatting this how `cpp` code should be formatted if not can revert its changes)

src/Native/dialog_gtk.c
```diff
@@ -0,0 +1,10 @@
+#include <gtk/gtk.h>
+
+void revery_alert_gtk(void *pWin, const char *szMessage) {
+  GtkDialogFlags flags = GTK_DIALOG_DESTROY_WITH_PARENT;
+  dialog = gtk_message_dialog_new(parent_window, flags, GTK_MESSAGE_ERROR,
```
Akin909: `filename` and `parent_window` currently cause errors as well as the use of dialog - have had time to look into why yet

src/Native/dune
```diff
@@ -4,12 +4,13 @@
     (preprocess (pps lwt_ppx))
     (library_flags (:include flags.sexp))
     (js_of_ocaml (javascript_files dialog.js))
-    (c_names dialog_cocoa dialog_win32)
+    (c_names dialog_cocoa dialog_win32 dialog_gtk)
     (cxx_names dialog)
     (c_flags (:include c_flags.sexp))
+    (c_library_flags (:include c_library_flags.sexp))
```
Akin909: separate config opt for `c_library` not equivalent to `library_flags` or any of the others
see [here](https://dune.readthedocs.io/en/latest/quick-start.html#defining-a-library-with-c-stubs-using-pkg-config)

src/Native/config/discover.ml
```diff
@@ -1,21 +1,44 @@
-type os =
-  | Windows 
-  | Mac 
-  | Linux 
-  | Unknown 
+module C = Configurator.V1
+
```
Akin909: Split out the functionality to derive flags so each platform can specify a getter, also ran `ocamlformat` on the file as I was having difficulty following its layout especially being new to ocaml syntax and things in the file didn't quite follow the docs so kept running into confusing syntax errors

Akin909:
@bryphe @OhadRau if either of you have a linux machine I'd really appreciate some help testing this. Locally I'm getting a segfault with the following error
```
(process:8099): Gtk-CRITICAL **: 14:33:21.922: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed
[1]    8099 segmentation fault (core dumped) 
```
The only reference I can find anywhere to this is [this](https://github.com/GNOME/gtk/commit/3ff7f1fd43965d425f2e7fe81f9caa03f1fa97ea) PR from 2012

I'm trying to figure out whether this error is general or relates to some config option or messing about I've done on my setup, don't think I've messed with my gnome setting too much but I might have done.

You should be able to build this locally and run the dialog example just to see what happens. As  I mentioned above the crash happens in a completely separate project just trying to draw a dialog on my machine, other stuff like creating a window with buttons etc works fine.

OhadRau:
don't currently have a native Linux machine but I can test in WSL if that helps.

OhadRau:
Having trouble getting this to compile. Still fails when trying to include `gtk.h` -- not sure if I'm doing something wrong, but I have pkg-config + libgtk2.0-dev + libgtk-3-dev installed so maybe I need to clear some build artifacts?

Akin909:
thanks for trying it out @OhadRau, if you run 
`pkg-config --libs gtk+-3.0` or `pkg-config --cflags gtk+-3.0` do they work?

not really familiar with WSL not sure how this would work on it tbh

OhadRau:
Yeah, turns out it was just an issue with a cached build so I just deleted `_esy` and it compiled. I'm getting the segfault as well though :/

Akin909:
@OhadRau that's at least consistent behavior, I think it might be something undocumented/not clearly documented in the API. I think that the example snippets probably miss some important bit of setup I need to do as I've tried some other more complete examples that involve opening a full window and that works. Maybe the style context (gtk's terminology) is gotten from the parent window somehow :man_shrugging:

Akin909:
Ah think I've got it now, the widgets etc. all have to called inside of a `gtk_application`, I've had to create a gtk window, to attach the dialog to in order to get it to run, wondering if I can somehow associate the dialog with revery's window :thinking:

Akin909:
Some progress here, this now correctly creates a dialog a few things to note,

The issues I was having seemed to be because the dialog wasn't associated with an application +/- window (previously tried creating a window, so think this might be more related to the dialog not being associated with an application but Im really not sure)

I v. briefly tried seeing if I could use the window handle that's passed in from glfw but not entirely sure how to convert it to a GTKWindow found [this](https://gist.github.com/mmozeiko/2401933b1fa89e5d5bd238b33eab0465) but it didn't exactly work but I only tried very briefly

Outstanding
- [ ] Closing the dialog seems to stall the revery window not sure why

src/Native/dialog_gtk.c
```diff
@@ -0,0 +1,30 @@
...
+  gtk_dialog_run(GTK_DIALOG(dialog));
+  gtk_widget_destroy(dialog);
+}
+
+void revery_alert_gtk(void *pWin, const char *szMessage) {
+  // TODO: figure out how to convert the pointer from an X11 window handle
+  // to a GTK window, see (for inspiration):
+  // https://gist.github.com/mmozeiko/2401933b1fa89e5d5bd238b33eab0465
+  GtkApplication *app;
+  app = gtk_application_new("org.gtk.revery", G_APPLICATION_FLAGS_NONE);
```
Akin909: not sure how exactly `atom` gets around this but given that it uses `chromium` and custom library on top of `gtk` some examples from that code base are unlikely to apply to this

Akin909:
@bryphe I've got this working locally and would appreciate you having a look over, two things to note which I've left as TODO comments is the conversion of the window handle to a `GTKWindow` not essential for now as this runs without it,

The second thing is that I narrowed down why this wasn't initially working to having to create a `gtk_application` without it seems like some context that was required to run the dialog was missing.

I'm not sure how `glfw` creates the windows but it appears much like other native windows on system which makes me think there's probably a gtk application created when a revery window is opened which could be re-used as I'm not sure what the consequences are of creating a new application.

Re. adding `lib-gtk` as far as I can tell we use an ubuntu vm for our linux pipeline on CI which I think should have gtk installed by default. Can't tell if it will need to be explicitly added (I think maybe not as the CI seems to be hitting the recurrent issue we've been getting)

src/Native/dialog_gtk.c
```diff
@@ -0,0 +1,32 @@
+#include <glib/gi18n.h>
```
bryphe: Ah, I think we'll need to wrap this in something like:
```
#ifndef WIN32
#ifndef __APPLE
...
#endif
#endif
```

Since we don't want this code to be picked up for Windows / Mac builds. This is probably why we're getting that header file build failure on those environments.

src/Native/config/discover.ml
```diff
@@ -1,21 +1,41 @@
...
+  {cflags= ["-I"; "."; "-x"; "objective-c"]; libs= []; flags= []}
+
+let get_linux_config c =
+  let default = {libs= []; cflags= []; flags= []} in
+  match C.Pkg_config.get c with
+  | None -> default
+  | Some pc -> (
+    match C.Pkg_config.query pc ~package:"gtk+-3.0" with
+    | None -> default
+    | Some conf -> {libs= conf.libs; cflags= conf.cflags; flags= []} )
```
bryphe: Does this return an include directory to, for the location of the header files? We might need that to fix the issue finding `glib/gi18n.h` on Linux in CI.

src/Native/dialog_gtk.c
```diff
@@ -0,0 +1,32 @@
...
+}
+
+void revery_alert_gtk(void *pWin, const char *szMessage) {
+  /*
+   * TODO:
+   * 1. figure out how to convert the pointer from an X11 window handle
+   * to a GTK window, see (for inspiration):
+   * https://gist.github.com/mmozeiko/2401933b1fa89e5d5bd238b33eab0465
+   *
+   * 2. Get reference to revery application, is there an existing
```
bryphe: It looks like if we can figure out issue 1, we might also be able to get the `GtkApplication` associated with the window with the `gtk_window_get_application` method: https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-get-application

I suspect there would already be a Gtk application created. We can track this in a separate issue, though

src/Native/config/discover.ml
bryphe: Do we need to include `libgtk-3-dev` as a dependency in CI? 

We have our set of includes here: https://github.com/revery-ui/revery/blob/be1ca7b3f64473baad85ed8a7df0e4f9877c4faa/azure-pipelines.yml#L21 and I don't believe we get GTK yet.

Akin909: @bryphe `glib/gi18n` might eventually be necessary but isn't actually needed in this current iteration so I will remove it. re. the function it does return the header locations for the package specified in this case "gtk+-3"
I'll add it to the CI I saw that but thought it might be bundled by with the distro even in CI

Akin909: Side note we can continuously call `pkg_config.query` if we need to derive the paths for other includes (not that this is always necessary)

bryphe:
Sorry about the CI builds not triggering - following up here: https://twitter.com/bryphe/status/1092806881406406656

Kicked off a manual build w/ the latest to see how it works 🤞

bryphe:
CI is green now! 🎉 Awesome work @Akin909 - we have the foundation in-place now for native linux integration!

Issue #227, opened 2019-01-21
Clicks start target and end target should match.


jordwalke:
In order to detect a click action the start should be required to match the end target. Right now it seems only the end target (mouse up) is listened to.

It would be nice if revery left room for gesture handlers to be built - a click is just a special case of a more complicated gesture system. but even with clicks, you often want very gesture-like cancellation. For example, if you mouse down, and then someone scrolls on the trackpad, the scrolling gesture should be able to prevent the click gesture from being recognized.

wokalski:
cc @ rauanmayemir we talked [about it yesterday](
https://github.com/briskml/brisk/pull/37#issuecomment-455859767). It'd be interesting to have the same API surface (we could have an empty .mli) as a dep to enforce it automatically.

bryphe:
Definitely! We actually have a `TODO` here tracking this:
https://github.com/revery-ui/revery/blob/3e8602acc2992c6e5e5475df0d1e18c7ee2109e3/src/UI_Components/Clickable.re#L29

A couple bugs with the existing logic - if you 'mouse down' on the `<Clickable />`, and move out, it will stay w/ opacity 1.0 (it should not). When the `<Clickable />` component was first implemented, we didn't really have faculty to handle both this case and the 'mouse up' case - but we do now with the `Mouse.setCapture` API.

Improved behavior for `<Clickable />` should be to:
- On `mousedown`, call `Mouse.setCapture`.
- On the captured `mousemove` - check if cursor position is in bounds, and set opacity appropriately.
- On the captured `mouseup` - check if cursor position is in bounds, and only dispatch if it is.

To get the bounds, we'll need to hold a `ref` to the clickable, and use `ref#getBoundingBox` to check if the cursor is inside the bounds. The `<Slider />` example shows something similiar: https://github.com/revery-ui/revery/blob/3e8602acc2992c6e5e5475df0d1e18c7ee2109e3/src/UI_Components/Slider.re#L48

jordwalke:
> It'd be interesting to have the same API surface (we could have an empty .mli as a dep to enforce it automatically.)

Check out Dune's upcoming virtual libraries!

PR #296, opened 2019-02-01
Bugfix: Clickable now tracking its bounds and mouse events appropriately

saitonakamura:
Fixes: #227

@bryphe I have a concern about constantly calling `setOpacity` on `mouseMove` (cause it may happen a lot)
https://github.com/revery-ui/revery/pull/296/files#diff-b70d90548201b5505a0a67ee349e4a53R42

Do we need to throttle it maybe, or it's somehow throttled for us?

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -12,8 +12,8 @@ open Revery_Math;
 type clickFunction = unit => unit;
 let noop = () => ();
 
-let isMouseInsideRef = (_ref: node, mouseX: float, mouseY: float) => {
-  let clickableDimensions: BoundingBox2d.t = _ref#getBoundingBox();
+let isMouseInsideRef = (ref: node, mouseX: float, mouseY: float) => {
+  let clickableDimensions: BoundingBox2d.t = ref#getBoundingBox();
   let pointVec = Vec2.create(mouseX, mouseY);
   BoundingBox2d.isPointInside(clickableDimensions, pointVec);
 }
@@ -30,8 +30,8 @@ let make =
       children: React.syntheticElement,
     ) =>
   component(slots => {
-    let (clickableRef, _setClickableRef, slots) = React.Hooks.state(None, slots);
-    let setClickableRef = r => _setClickableRef(Some(r));
+    let (clickableRef, setClickableRefOption, slots) = React.Hooks.state(None, slots);
+    let setClickableRef = r => setClickableRefOption(Some(r));
 
     let (animatedOpacity, setOpacity, _slots: React.Hooks.empty) =
       React.Hooks.state(0.8, slots);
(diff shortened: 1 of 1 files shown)


src/UI_Components/Clickable.re
```diff
@@ -23,19 +30,47 @@ let make =
...
-     * This logic isn't really correct,
-     * for the case where you hold down the mouse,
-     * move around (leave the item), and come back.
-     */
-    let onMouseDown = _ => setOpacity(1.0);
-    let onMouseUp = _ => {
+    let onMouseMove = (mouseX: float, mouseY: float) => {
+      switch (clickableRef) {
+        | Some(clickable) => {
+          /* Is it ok to run it on every event? Do we need to throttle mouse move or it's already throttled? */
```
bryphe: I believe this should be OK. The timing of the mouse events is platform dependent (we use the `glfwSetCursorPosCallback` method to get this: <https://www.glfw.org/docs/latest/input_guide.html#cursor_pos>) - which I believe should be ~1 per frame (essentially each time we call `glfwPollEvents`). It could be different for the JS builds, though.

One potential optimization would be to double-check if the current opacity value is already set (for example, don't set to 1.0 or 0.8 if it is already 1.0 or 0.8, respectively - that may save the reconciler some work and save us from re-rendering).

src/UI_Components/Clickable.re
```diff
@@ -23,19 +30,47 @@ let make =
...
+        | None => ()
+      }
+      
       setOpacity(0.8);
-      onClick();
+      Mouse.releaseCapture();
+    }
+
+    let onMouseDown = _ => {
+      Mouse.setCapture(
```
bryphe: Looks great, @saitonakamura ! This is exactly the logic we needed to improve this interaction - thank you! 👍

src/UI_Components/Clickable.re
```diff
@@ -7,10 +7,17 @@
 
 open Revery_UI;
 open Revery_Core;
+open Revery_Math;
 
 type clickFunction = unit => unit;
 let noop = () => ();
 
+let isMouseInsideRef = (_ref: node, mouseX: float, mouseY: float) => {
```
bryphe: Could you please rename `_ref` -> `ref`?

bryphe:
Thanks @saitonakamura , looks great! Looks like there were some formatting issues blocking the CI; I fixed those with `esy format` and committed. Will bring it once CI is green 👍

src/UI_Components/Clickable.re
```diff
@@ -23,19 +30,47 @@ let make =
       children: React.syntheticElement,
     ) =>
   component(slots => {
+    let (clickableRef, _setClickableRef, slots) = React.Hooks.state(None, slots);
```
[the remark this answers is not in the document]
bryphe: Ah thanks, yes, those should be fixed too! Sorry about that

PR #298, opened 2019-02-01
Convert autocomplete styles

Akin909:
This somehow was missed out :thinking:


bryphe:
Ah nice, thanks! I guess it might not be compiled at the moment (it's not included in the Examples). Glad you caught it!

PR #299, opened 2019-02-01
Center example container and increase size

Akin909:
Center the examples window and increase its size to be half of the available window height and width, this is a partial fix for #226


bryphe:
Looks great, thank you @Akin909 ! 💯

PR #304, opened 2019-02-03
Add padding to styles

### LINKED ISSUE #302: Style: Missing padding
Are there any plans to add `padding`? I asked about this in the Discord and was told it might actually already exist and that it might have been forgotten? 🙂

samatar26:
Fixes #302

### DIFF
--- a/src/UI/Selector.re
+++ b/src/UI/Selector.re
@@ -26,6 +26,15 @@ type selector('a) =
   | MarginHorizontal: selector(int)
   | Margin2: selector(xy)
   | Margin4: selector(coords)
+  | PaddingTop: selector(int)
+  | PaddingLeft: selector(int)
+  | PaddingRight: selector(int)
+  | PaddingBottom: selector(int)
+  | Padding: selector(int)
+  | PaddingVertical: selector(int)
+  | PaddingHorizontal: selector(int)
+  | Padding2: selector(xy)
+  | Padding4: selector(coords)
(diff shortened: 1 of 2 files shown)


Akin909:
@samatar26 is padding implemented in the actual `node`? I don't think so, so adding it to the style would probably not change anything.

I was thinking that to actually implement this you would probably have to increase the size of the `quad` which is drawn in the GL layer

We use openGL to draw the elements that make up the text and views, to add padding I was thinking you'd have to increase the dimensions of the element to be drawn by whatever amount of padding the style specified in `src/UI/Node.re`.

I was also wondering whether the way it works in CSS would a good starting point i.e. if `border-box` then force it to have the same dimensions but restrict the available size for its content by whatever the specified padding is. Although that I think will be more complicated, @bryphe might have given this some thought already?

samatar26:
@Akin909 Good point. It currently applies the padding, but only in the style of `box-sizing: border-box`, i.e. keeping the same dimensions.

Akin909:
@samatar26 that's great the logic must already be in place somewhere I thought maybe it wasn't in which case looks good tbh I think its fine to have it work that way i.e. the border-box, think `content-box` is a much less common use case

Akin909:
@samatar26 could you add the padding to the `Selector` as well its a module that allows us to pull out values from a list of styles, it also has a long list of props that will need updating other than that if it working then looks good to me :+1:

samatar26:
Done 👍

Issue #288, opened 2019-01-30
UI - ScrollView: Momentum scrolling + Bounce


bryphe:
Momentum scrolling is a great, naturally interaction for scrolls - it'd be great to have this instead of the 'instant' scroll behavior today. The momentum scrolling gives more context during the interaction and more control - making it easier for you to scroll to the right element. 

#### A physically based scroll model

- When using the mousewheel, __apply a 'force' to the scroll thumb__. `Force = mass * acceleration`, and we can simplify and just assume a mass of 1 for the thumb, which would be `Force = acceleration`. Each frame, we'd increase the velocity of the thumb based on the `acceleration  * t`. In addition, we'd adjust the actual position of the thumb based on velocity (`position = position + velocity * t`). This would be similiar for other gestures, like a scroll gesture on touch.

- When clicking on the track, use an appropriate force. This would be similiar to the mousewheel, but would need to be calibrated to use more 'force'.

- __Model 'friction'__ - without any friction, as soon as you scroll, the cursor would just fly to the end! So we'd need to model friction. Friction is a constant resistive force - this could be simplified by subtracting the acceleration by a constant factor * t every frame. This could be modeled in more detail by static friction and kinetic friction, but I'm not sure if that actually is useful for the UX.

- __Model 'springs' at the boundaries__ - Without modelling any sort of spring at the boundaries, the scroll would be smooth until the end, and then it would just instantly stop. iOS implements a nice 'bounce' effect - this can be modeled physically by using the spring equation `F=-kx`, where `k` is a spring constant (can play with different values), and `x` is the distance passed the boundary. In effect, as you go further out the boundary, the force applied increases - this gives a nice bounce effect. Different values of `k` give different types of bounce (the stiffness of the spring). We'd have to calculate the force due to spring acceleration with this, and apply it to our formula too.

#### Implementation

To manage momentum scrolling, we'd need to store some state for each scrollbar:
- The current position (this is already tracked)
- The current velocity
- The current acceleration

We'd also need to track the time between each frame (`t` or delta time, to apply to the equations).

While the physics 'simulation' is in effect, we'd want to render every frame at (60fps). This could be implemented by setting a state value (like the last frame time) every time we want a new frame - this will trigger a re-render because the reconciler will mark the tree as dirty. We should do this whenever a force is applied, or whenever the velocity is greater than a certain threshold. Once the scrollbar has 'settled' and is no longer in motion, we should no longer re-render.

Each time we render, we'd want to calculate a new position:

```
[...]

PR #307, opened 2019-02-04
Scroll for smaller amounts

jchavarri:
No companion issue. Found while investigating #288.

When the wheel event would return `0.x` we wouldn't move the scroll position. This PR changes the calculation order so that we multiple first by 25 and then round, so every event causes some effect, even if small.

### DIFF
--- a/src/UI_Components/ScrollView.re
+++ b/src/UI_Components/ScrollView.re
@@ -101,7 +101,7 @@ let make =
 
         let scroll = (wheelEvent: NodeEvents.mouseWheelEventParams) => {
           let newScrollTop =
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks @jchavarri , nice catch!

Looks like CI is failing due to formatting - might want to try `refmt --inplace src/UI_Components/ScrollView.re` But code change looks good to me 👍

jchavarri:
Thanks @bryphe ! I ran refmt and now CI is happy.

One small thing I realized is that to make sure `refmt` runs with the same version as CI, one can use `esy shell` and then `refmt --in-place src/UI_Components/ScrollView.re` from that shell. 😄 

Merging then! :shipit:

PR #310, opened 2019-02-05
Examples: Analog Clock Example

### LINKED ISSUE #41: Example: Clock
A 'wall clock' would be a great example for showcasing the transform functionality in #40 - the rotations could be used to implement the hour / minute / second hands. In addition, it could be packaged as a custom component, so that there'd be a clear cut example of a stateful custom component.

lessp:
Hey all!

I took a stab at creating an Example Clock, I'm still very new to Revery and the Reason-world so any improvements are most welcome, of course! 👋 

The mov to gif-conversion seems to have left some second ticks behind! 🤷‍♂️ 

analog-clock-revery

Closes #41


lessp:
> The example looks great! The only blocker I see is fixing the lockfile build issue (one option would be to revert the lockfile commit for now - and we could track upgrading esy on our CI machines in a separate issue/PR).

Sweet! 

Looks like I forgot to push the latest changes (visually), so I did that too.
Also merged in the latest changes from the master-branch accepted the changes from that lock-file. 

Should work, hopefully! 🙂 

--

Looks like I'm using esy `0.5.6`

bryphe:
Green CI - looks great 👍 Thanks again for all the work on this, @lessp ! Bringing it in now.

Issue #42, opened 2018-11-14
Example: To-do list


bryphe:
A to-do list is the canonical example for UI frameworks, so we should have one here too.

There's a few things we need first:
- Checkbox control
- Mouse input

schinns:
I will start working on this 😄

bryphe:
Sweet, thank you @schinns !

bryphe:
Fixed by #311 - thanks @faisalil !

schinns:
thanks @faisalil !

albertorestifo:
Given that #311 has been merged, can we close this?

bryphe:
Fixed with #311

PR #311, opened 2019-02-06
Example: Adding a very basic version of a todo app

faisalil:
- I am very new to reason, so any suggestions are welcome even if they are minor nits. 
- As part of this, I did a workaround to an issue in Input and made the event subscriptions Always instead of OnMount, this breaks having multiple Input fields but this needs to be fixed differently. 
- The goal is to make this match the full todo mvc example, however, I wanted to send something out as a start and then take some fixes to Input before continuing with this.


bryphe:
@faisalil - awesome, thank you very much! Can't wait to try this out

src/UI_Components/Input.re
```diff
@@ -87,31 +90,36 @@ let make =
      */
     let slots =
       React.Hooks.effect(
-        OnMount,
+        Always,
```
bryphe: Nice catch on this issue @faisalil  - thanks for investigating & fixing it!

examples/TodoExample.re
```diff
@@ -0,0 +1,225 @@
...
+            | _ => Colors.lightSkyBlue
+            }
+          }
+          onClick={() => onPickingFilter(NotCompleted)}
+        />
+      </View>
+    );
+
+  let createElement = (~children, ~currentFilier, ~onPickingFilter, ()) =>
+    React.element(make(children, currentFilier, onPickingFilter));
```
bryphe: Nit: I think `currentFilier` should be `currentFilter`?

examples/TodoExample.re
```diff
@@ -0,0 +1,225 @@
...
+      List.map(
+        item => item.id == id ? {...item, isDone} : item,
+        state.todos,
+      );
+    {...state, todos};
+  | ChangeFilter(filter) => {...state, filter}
+  };
+};
+
+module FilterSection = {
```
bryphe: One thing you might be interested in - @jchavarri and @wokalski are looking at ways to streamline  the component creation to a function / remove boiler plate here: https://github.com/briskml/brisk-reconciler/issues/6

Nothing to be changed since we don't have it yet, but would be great to get your feedback on that since you have a fresh perspective!

examples/TodoExample.re
```diff
@@ -0,0 +1,225 @@
...
+  task: string,
+  isDone: bool,
+};
+
+type filter =
+  | All
+  | Completed
+  | NotCompleted;
+
+let textOfFilter = (filter: filter) => {
```
bryphe: One convention that comes up in OCaml (and that Reason inherited) is a convention of using a type `t` with a module, especially if there are functions that use it. So you could potentially refactor this to something like:
```
module Filter {
   type t =
   | All
   | Completed
   | NotCompleted;

   let show = (v: t) => switch(v) {
   | All => "All"
   | Completed => "Completed"
   | NotCompleted => "NotCompleted"
   };
};
```

Not something that is strictly necessary, but it's a convention that I found interesting as I start diving into Reason - the ubiquity of a module plus a `type t`. 
- '1 Module for 1 thing': https://github.com/ostera/reason-design-patterns/blob/master/patterns/1-module-1-thing.md

It's a little bit different than thinking in an OO language - a lot of times the Module will have a core type + functions that operate on that top. So comes up a lot in the standard modules - like there is a `Hashtbl.t('a, 'b)` that the Hashtbl module provides functions for: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Hashtbl.html

No changes necessary - just thought you might be interested in this 😎 Nice use of variants for this!

bryphe:
Just got to play with it a bit! Overall the code looks excellent @faisalil - I just put some minor nits / commentary in, but it looks great. How was your first Reason experience?

I took a quick screen grab of testing it out in case anyone else wants to see it:
todo mvc

I noticed one bug that I think is a reconciler issue - when I switch between 'Completed' and 'Not Completed' tasks, sometimes the checkbox state doesn't change. The logic in the sample looks correct to me. 

We could try specifiying a `key` for the todo items to workaround this (the `element` function takes an optional `key` parameter: https://github.com/briskml/brisk-reconciler/blob/f545ba090984794acfacd8f4ff7896a921b3d67f/lib/Brisk_reconciler.rei#L53
) and an example usage in a test case:
https://github.com/briskml/brisk-reconciler/blob/f545ba090984794acfacd8f4ff7896a921b3d67f/test/Test.re#L82

But we can track this separately, too - I do think it's possibly a value not being updated by the reconciler.

The only blocker I see is that CI is failing due to formatting - could you try running `esy format` & committing the changes?

So cool that you tried it out and built this @faisalil , really appreciate it!

wokalski:
@bryphe oh no... I should have a couple of fixes coming in the coming days. I hope it'll get fixed!

faisalil:
Thanks for the comments @bryphe , I will get to addressing the comments and fixing the merge conflicts tomorrow.

bryphe:
Thanks @faisalil ! I'll publish a new version of the playground (https://www.outrunlabs.com/revery/playground/) once its in! 👍

examples/TodoExample.re
faisalil: Thanks. Fixed it.

examples/TodoExample.re
faisalil: That will be great. I will check this out. Most of the code to create a basic component didn't feel natural. glad you are looking into it.

examples/TodoExample.re
faisalil: Very useful. Thanks for sharing. One goal I have in helping here is to ramp up on Reason, so thanks for this.

faisalil:
> ) and an example usage in a test case:
> https://github.com/briskml/brisk-reconciler/blob/f545ba090984794acfacd8f4ff7896a921b3d67f/test/Test.re#L82

@bryphe , calling 

>               let key1 = Key.create();

Gives an error : "Error: Unbound value Key.create"

I am not sure about the right syntax and I cannot find any documentation on this.

faisalil:
> Just got to play with it a bit! Overall the code looks excellent @faisalil - I just put some minor nits / commentary in, but it looks great. How was your first Reason experience?

@bryphe, I am actually liking reason. The startup of the test app is amazing and this kind of perf is always very attractive. 
The editor support is kind of week (in VSCode). I wish show definition works so you can see all the options you are calling. Do you guys use Oni? Is the support better? I am not that familiar with vim so it might be a challenge to use anything but vscode.

faisalil:
Change should be ready for final review.

bryphe:
> @bryphe , calling

> ```
>           let key1 = Key.create();
> ```

> Gives an error : "Error: Unbound value Key.create"

Ah, looks like this needs to be `React.Key.create` - all of the brisk module is accessed behind a `React` module.

> @bryphe, I am actually liking reason. The startup of the test app is amazing and this kind of perf is always very attractive.

Very cool! @jordwalke would be happy to hear this 😄 

> The editor support is kind of week (in VSCode). I wish show definition works so you can see all the options you are calling. Do you guys use Oni? Is the support better? I am not that familiar with vim so it might be a challenge to use anything but vscode.

There are still some challenges with language support - but I believe the best environment at the moment is OSX + VSCode + `reason-vscode`.

However, there is a really powerful language tool called `merlin` of ocaml/reason: https://github.com/ocaml/merlin and @andreypopp has a plugin that uses it in VSCode - https://github.com/andreypopp/vscode-merlin - I think this will be the best long-term experience, but haven't tried it yet. I'd like to bundle it in Oni2 so that reason users have a great out-of-box experience (like you get with VSCode+typescript).

bryphe:
And the sample looks great to me, @faisalil - thanks for all your work on this! 🎉  This is a sample I've wanted for a long time - fixes #42

PR #313, opened 2019-02-06
Feature/mouseenter/leave/out/over events

### LINKED ISSUE #186: UI Events: onMouseOver, onMouseOut
I'd like to be able to implement 'hover styles' on `<Button />`, `<view />`'s, etc.

Ideally, we could have something like this:
```
        let defaultBackgroundColor = Colors.red;
        let hoverBackgroundColor = Colors.green;
        let (backgroundColor, setBackgroundColor) = useState(defaultBackgroundColor);

        let style = Style.make(~backgroundColor, ..., ());

       <view onMouseOver={() => setBackgroundColor(hoverBackgroundColor)} onMouseOut={() => setBackgroundColor(defaultBackgroundColor)} />
```

This might be a bit verbose, but would allow us to have more convenient wrappers on top of these primitives.

I think the work done by @OhadRau in #168  opens the door for this - we know which element the mouse is over. We could keep track of the 'current' element we are over, and if that changes, dispatch the `onMouseOut` event, and dispatch the `onMouseOver` event to the new element.

To have similiar behavior as the browser - `onMouseOut` and `onMouseOver` would __bubble__ (like @Akin909 's implementation in #173 ). However, there are counterpart events `onMouseLeave` and `onMouseEnter` that _do not_ bubble.

samatar26:
Will eventually fix #186. 

kapture 2019-02-06 at 21 24 29

I am currently able to dispatch a mouseOver and MouseOut event (albeit not bubbling yet, should've just called them onMouseEnter and onMouseLeave 😆  😆 ) to the relevant elements. I'm currently using a `getDeepestNode` util, but will probably have to keep track of all the elements parents as hovering over the children shouldn't fire an oMouseOut event. I'm also a little unclear about `handleCapture, releaseCapture` and capturing in general in the `Mouse` module, so it could be that my entire approach is wrong but will do some further investigating. Any general feedback on this so far is much appreciated and thanks for giving me the time to work on this!!


samatar26:
I've updated quite a bit of the logic and the mouseEnter and mouseLeave seem to be behaving exactly like in the browser. Will add tests and iterate on some of the logic/functions as I'm still getting to grips with Reason 😆

kapture 2019-02-08 at 20 57 03

examples/HoverExample.re
```diff
@@ -0,0 +1,147 @@
...
+      <View
+        style=Style.[
+          flexDirection(`Row),
+          justifyContent(`Center),
+          alignItems(`Center),
+          width(500),
+          height(500),
+          backgroundColor(state.parentBackground),
+        ]
+        onMouseEnter={_ => dispatch(SetParentBackground(Colors.lightGray))}
```
bryphe: Excellent example! Thanks for your thought in how to showcase / test this, @samatar26 👍

src/UI/Mouse.re
```diff
@@ -125,6 +251,21 @@ let dispatch =
         ();
       };
 
+      let mouseMove = isMouseMoveEv(eventToSend);
```
bryphe: Regarding capture - I think we'll want to move this block inside the `!handleCapture` block below.

The idea is, if `Mouse.setCapture` is enabled - whoever called it is responsible for handling any mouse events (to the exclusion of other mouse events). So when a 'capture' is active - we can skip sending the mouseEnter/mouseLeave events.

src/UI/Mouse.re
```diff
@@ -107,6 +111,128 @@ let isMouseDownEv =
...
+    storedNodesUnderCursor :=
+      [
+        {handler: deepestNode#handleEvent, id: deepestNode#getInternalId()},
+        ...storedNodesUnderCursor^,
+      ];
+    sendMouseEnterEvents(parent, cursor);
+  };
+};
+
+let rec handleMouseEnterDiff = (deepestNode, _newNodes, cursor: Cursor.t) => {
```
bryphe: Thanks for implementing this diffing logic, and for the notes! 👍

src/UI/Revery_UI.re
```diff
@@ -68,6 +68,35 @@ let start =
       },
     );
 
+  /*
+   Probably not a good idea to subscribe to onMouseMove and calling onMouseEnter everytime
```
bryphe: I didn't quite follow this comment - is it referring to listening to `window.onMouseEnter`?

samatar26: Aaah sorry, it's left over from an older commit  and isn't relevant anymore.

samatar26:
@bryphe I've moved the mouseenter/mouseleave event handling inside the `!handleCapture` block as 
 you've suggested 👍 . I think it's good to go, except that  in the tests I think I'm a little confused as whether to call `Mouse.dispatch` with  `InternalMouseMove`, instead of `InternalMouseEnter/Leave`(which I'm doing now). 

I've tried doing it with dispatching a `MouseMove` event, which in turn should dispatch a `MouseEnter/Leave` event, but I've noticed that doing it that way, there's some shared state between the tests which causes it to fail. I.e. the nodes that the cursor is over in the first test, still exist by the time I run the next one.

samatar26:
@bryphe The tests have actually highlighted something I've missed, which was that I forgot to append a final node of the tree when sending mouseenter events and now all the tests are passing. It should be ready to merge now 👍

samatar26:
I know I said ready to merge, but currently adding onMouseOver and onMouseOut events too 😆. I figured since they're so similar anyway and it will allow us to close the issue once this PR's done.

bryphe:
That's awesome, @samatar26 ! Thanks for adding those too (and doing the merge!) 🎉

Akin909:
@samatar26 might want to get this in sooner rather than later so an API change doesn't leave you having to refactor the PR :wink:

samatar26:
Thanks @Akin909, working on  it as we speak. Can hopefully go in today 👍

samatar26:
@bryphe, @Akin909 this should be ready to go in now 👍

src/UI/Selector.re
```diff
@@ -7,6 +7,7 @@ type selector('a) =
   | FontSize: selector(int)
   | Width: selector(int)
   | FlexGrow: selector(int)
+  | FlexWrap: selector(Layout.LayoutTypes.wrapType)
```
faisalil: why was this change needed? just trying to know more.

samatar26: @faisalil Good question. So there was a PR for adding FlexWrap to our list of styles, but it wasn't added to the `Selector` module which allows us to pull out values from a list of styles and since it wasn't added there it wasn't coming through.

src/UI/Mouse.re
```diff
@@ -336,8 +341,15 @@ let dispatch =
                   */
               sendMouseEnterEvents(deepNode, mouseMoveEventParams);
               bubble(deepNode, MouseOver(mouseMoveEventParams));
-            | [_, ..._xs] =>
-              handleMouseEnterDiff(deepNode, [], mouseMoveEventParams)
+            | [node, ..._tail] =>
```
Akin909: if you're not using tail why not just do
```suggestion
| [node,  _ ] =>
```

src/UI/Mouse.re
```diff
@@ -107,8 +161,135 @@ let isMouseDownEv =
...
+    /**
+     *  1. Get parent if node is not found in currentlyStoredNodes
+     *  2. If no parent, replace currently stored nodes entirely with new tree
+     *  3. If parent found, call function again
+     */
+    let parent = deepestNode#getParent();
+    switch (parent) {
+    | None =>
+      /**
+     *  MouseEnter/Leave events
```
Akin909: nit: can you arrange this properly (sorry this is me being way too nitpicky) 🙏

src/UI/Mouse.re
samatar26: Good point!

src/UI/Mouse.re
```diff
@@ -107,8 +161,135 @@ let isMouseDownEv =
...
+
+      List.iter(
+        newNode => newNode#handleEvent(MouseEnter(evtParams)),
+        [deepestNode, ...newNodes],
+      );
+
+      /**
+       * MouseOver/Out Events
+       */
+      bubble(List.nth(storedNodesUnderCursor^, 0), MouseOut(evtParams));
```
Akin909: an alternative to this is
```suggestion
bubble(List.hd(storedNodesUnderCursor^), MouseOut(evtParams));
```

src/UI/Mouse.re
```diff
@@ -107,8 +161,135 @@ let isMouseDownEv =
...
+    };
+  };
+};
+
 let dispatch =
-    (cursor: Cursor.t, evt: Events.internalMouseEvents, node: Node.node('a)) => {
+    (
+      cursor: Cursor.t,
+      evt: Events.internalMouseEvents,
+      node: Node.node(RenderPass.t),
```
Akin909: Nice this worked 💯

src/UI/Mouse.re
```diff
@@ -127,6 +308,52 @@ let dispatch =
...
+          switch (deepestNode^) {
+          | None =>
+            /**
+             * if no node found, call bubbled MouseOut on deepestStoredNode if there's some stored nodes
+             * And recursively send mouseLeave events to storedNodes if they exist
+             */
+            (
+              switch (storedNodesUnderCursor^) {
+              | [] => ()
+              | [node, ..._tail] =>
```
Akin909: same as above

src/UI/Mouse.re
samatar26: So much cleaner!! Hated passing a `List.nth(list, 0)` everywhere 😄

src/UI/Mouse.re
```diff
@@ -127,6 +308,52 @@ let dispatch =
...
+
+            sendMouseLeaveEvents(
+              storedNodesUnderCursor^,
+              mouseMoveEventParams,
+            );
+
+          | Some(deepNode) =>
+            switch (storedNodesUnderCursor^) {
+            | [] =>
+              /**
```
Akin909: are you using `refmt` it should automatically format your code (including comments I think 🤔)

samatar26: I'm using a `reason-vscode` extensions and it  does format on save, not sure if it's `refmt` or something else.

src/UI/Mouse.re
samatar26: It's still gotta be `[node, ..._]`,  or  else it complains about non-exhaustive  pattern matching.

src/UI/Mouse.re
samatar26: Aaah, do you mean the comment?

src/UI/Mouse.re
samatar26: Comments do seem to format weirdly though 🤔

samatar26: The culprit seems to have been the double star `**`. Removing it seems to have fixed the `refmt`. Good spot!

src/UI/Mouse.re
Akin909: ah k no worries then

src/UI/Mouse.re
Akin909: yeah

src/UI/Mouse.re
```diff
@@ -107,8 +161,135 @@ let isMouseDownEv =
...
+          storedNodesUnderCursor := newNodes @ [node, ...tail];
+        }
+      };
+    };
+
+    let handleMouseOverDiff = listOfNodes => {
+      switch (newNodes) {
+      | [] =>
+        if (deepestNode#getInternalId()
+            != List.nth(listOfNodes, 0)#getInternalId()) {
```
Akin909: `List.hd` here as well

samatar26:
@Akin909 np, it was a lot of fun 💯! And hover styles sounds like an interesting problem, can look into that next!

bryphe:
Wow, so cool to see this come together, @samatar26 ! Thanks for all your work on it (even getting the out/over events in 😍 ). And thanks for all your help reviewing, @Akin909 !

Just tried it out and feels great!  Will bring this in now 👍 

mouse over

> now if only we had `hover` styles

Oh ya, for sure... I'm excited about the possibilities of potentially hooking up the these over/out events with things like spring animations. It could feel really nice in the context of a native app!

Issue #290, opened 2019-01-30
UI - Input: Hook up <Input /> element to Focus


bryphe:
__Issue:__ Our current `<Input />` element, introduced in #205 , directly hooks to the `window.onKeyPress`/`window.onKeyDown` events. This works, but means that `<Input />` elements have to cooperatively manage who has focus, and ignore input events if they don't currently have focus. 

We also have a concept of `Focus` introduced in #199 - but currently the Focus / `<Input />` element are not in sync.

The design I'd like to look investigate is, instead of each `<Input />` listening directly to window events - only the focus manager listens to them at the top-level (the focus manager), and the focus manager will dispatch those events to the focused item. This will be important in terms of managing focus across multiple components (for example, tabbing between inputs/buttons/etc), since the concept of focus is broader than just the `<Input />` elements.

This is sort of like how we handle mouse events right now - we have the `Mouse` module that listens to window-level events, and then the `Mouse` module, based on the position/z-index, decides which component to actually dispatch the event to, instead of every element listening to window mouse events. 

Likewise, for `Focus`, it's similiar, and actually even more straightforward - the focus manager knows which element has focus and can dispatch keydown/keypress/keyup events directly to that element.

A few scenarios this would enable:
- Programmatically calling `focus` on an `<Input />` should enable it to receive key presses
- Items like Buttons should be able to handle key presses (ie, space/enter to commit)
- We should be able to tab through form elements via a tabIndex

faisalil:
Planning to take a stab at this next.

faisalil:
So @bryphe , I spent sometime looking at the Focus and Mouse module to see how they handle the events for mouse events and one simple way I was able to think about to resolve the keyboard events for the Input elements was the following: 
- In the start of the application (in Revery_UI), I can subscribe to the keyboard events. 
- In the handler, the logic is simple: get the current Focused element (Using the Focus.focused global state) and fire the keyboard events handlers on it if it is listening to those. Any thoughts?

faisalil:
The outline solution above was implemented in https://github.com/revery-ui/revery/pull/324

bryphe:
@faisalil - sorry for the slow reply!

I just checked out #324 - looks great - the solution you outlined above is perfect 👍 

The only thing I think we're missing is (potentially) bubbling the key events - but that should be tracked as another issue. Thanks for putting together the scaffolding for this.

faisalil:
[...]

PR #324, opened 2019-02-10
Input: fixing issue of multiple Inputs not working on the same screen

faisalil:
The way this was fixed was the following: 
- Added a central Keyboard module that handles keyboard events. 
- Inside Revery_UI we subscribe to those events and handle them in Keyboard.re
- we handle those by getting the focused node and calling the handler methods on that if they exist. 

This issue was created to address https://github.com/revery-ui/revery/issues/290


src/UI_Components/Input.re
```diff
@@ -82,48 +81,14 @@ let make =
         slots,
       );
 
-    /*
-       TODO: Setting the hook to run only on mount means that the onChange
-       handler is only called with a stale version of state
-       BUT setting the value to ALWAYS causes one handler to to seize
-       control of the event listening
-     */
-    let slots =
```
bryphe: Nice to see this logic get simplified! 👍

src/UI_Components/Input.re
```diff
@@ -65,7 +64,7 @@ let defaultStyles =
 let component = React.component("Input");
 let make =
     (
-      ~window,
+      ~window as _window,
```
bryphe: Can we remove this property now? This was one awkward thing about the `<Input />` today - that we needed to pass a `window` to it. 

With your change, I believe this is decoupled now, and we should be able to remove this completely 👍

examples/Examples.re
```diff
@@ -47,7 +47,7 @@ let state: state = {
     {name: "Border", render: _ => Border.render(), source: "Border.re"},
     {
       name: "ScrollView",
-      render: _w => ScrollViewExample.render(),
```
bryphe: So nice to see this dependency go away! Thanks @faisalil !

bryphe:
Perfect, thanks for the updates! 👍  Merging now.

Issue #287, opened 2019-01-30
UI - ScrollView: Horizontal measurement / scrolling isn't working correctly


bryphe:
__Issue:__ If a `<ScrollView />` is wrapped around a child element that overflows, it won't show a horizontal scroll bar.

Inside the `ScrollView` control, we create a container around the children - and we compare the size of the container and the size of the children to see if we need to show scrollbars.

For vertical scrolling, the height of the container and children is reported accurately, and if the child height is greater than the container height, we show scrollbars.

However, this isn't the case for horizontal scrollbars - we aren't getting the expected values reported for the width of the element.

This is potentially a styling / `flex` issue - probably just a bug in how we are applying styles.

There's a few related spots in the code that we could start looking. I kept in some debug logging to check the child width vs container width:

In addition, there's a descriptive comment here:

glennsl:
This doesn't seem to work for the vertical scrollbar either, though it might be subtly different issue. You can see this in the TodoMVC example, where adding items until it overflows still doesn't have it render the vertical scrollbar. It's only the next time it renders, whether that's from scrolling, adding another item or something else, that it shows the scrollbar. But if that re-render is from adding a new item, the size of the scrollbar thumb is still wrong (until next render).

The issue seems to be that (naturally) the children are rendered after the ScrollView, and therefore when measured in ScrollView returns the "old" measurements. 

The measurements model seems pretty flawed in general. Could it perhaps be made to work as a hook, that's fired post-rendering, and only if measurements have actually changed?

PR #325, opened 2019-02-10
Feature/scale factor

Akin909:
Relates to #226, this PR uses the `glfwGetMonitorPhysicalSize` to calculate the scale factor for a window and use that adjust to screen space (or at least it will eventually)

### DIFF
--- a/src/Core/Monitor.re
+++ b/src/Core/Monitor.re
@@ -18,8 +18,13 @@ let getScaleFactor = () => {
   let physicalSize = Glfw.glfwGetMonitorPhysicalSize(monitor);
   let dpiH = calculateDPI(vidMode.height, physicalSize.height);
   let dpiW = calculateDPI(vidMode.width, physicalSize.width);
-  let avgDPI = dpiW +. dpiH /. 2.0;
-  let scaleFactor = int_of_float(avgDPI) / 96;
+  let avgDPI = (dpiW +. dpiH) /. 2.0;
+  let rawScaleFactor = avgDPI /. 96.;
+  /**
+     int_of_float aggressively shaves off any fractional values so we round the value first
+     e.g. int_of_float 2.93 results in 2
+   */
+  let scaleFactor = int_of_float(floor(rawScaleFactor +. 0.5));
   scaleFactor >= 1 ? scaleFactor : 1;
 };
 
--- a/src/Core/Monitor.re
+++ b/src/Core/Monitor.re
@@ -9,6 +9,20 @@ type size = {
 
 let getPrimaryMonitor = () => Glfw.glfwGetPrimaryMonitor();
 
+let calculateDPI = (modeDimension, physicalDimension) =>
+  float_of_int(modeDimension) /. (float_of_int(physicalDimension) /. 25.4);
+
+let getScaleFactor = () => {
+  let monitor = Glfw.glfwGetPrimaryMonitor();
+  let vidMode = Glfw.glfwGetVideoMode(monitor);
+  let physicalSize = Glfw.glfwGetMonitorPhysicalSize(monitor);
+  let dpiH = calculateDPI(vidMode.height, physicalSize.height);
+  let dpiW = calculateDPI(vidMode.width, physicalSize.width);
+  let avgDPI = dpiW +. dpiH /. 2.0;
+  let scaleFactor = avgDPI /. 96.0;
+  scaleFactor >= 1.0 ? scaleFactor : 1.0;
+};
+
 let getSize = monitor => {
   let monitorSize = Glfw.glfwGetVideoMode(monitor);
   let ret: size = {width: monitorSize.width, height: monitorSize.height};
--- a/src/Core/Monitor.re
+++ b/src/Core/Monitor.re
@@ -19,8 +19,8 @@ let getScaleFactor = () => {
   let dpiH = calculateDPI(vidMode.height, physicalSize.height);
   let dpiW = calculateDPI(vidMode.width, physicalSize.width);
   let avgDPI = dpiW +. dpiH /. 2.0;
-  let scaleFactor = avgDPI /. 96.0;
-  scaleFactor >= 1.0 ? scaleFactor : 1.0;
+  let scaleFactor = int_of_float(avgDPI) / 96;
+  scaleFactor >= 1 ? scaleFactor : 1;
 };
 
 let getSize = monitor => {
--- a/src/Core/Monitor.re
+++ b/src/Core/Monitor.re
@@ -21,10 +21,11 @@ let getScaleFactor = () => {
   let avgDPI = (dpiW +. dpiH) /. 2.0;
   let rawScaleFactor = avgDPI /. 96.;
   /**
-     int_of_float aggressively shaves off any fractional values so we round the value first
-     e.g. int_of_float 2.93 results in 2
+     NOTE: int_of_float aggressively shaves off any fractional values
+     e.g. int_of_float 2.93 results in 2, this means the dpi we calculate
+     is rounded down
    */
-  let scaleFactor = int_of_float(floor(rawScaleFactor +. 0.5));
+  let scaleFactor = int_of_float(rawScaleFactor);
   scaleFactor >= 1 ? scaleFactor : 1;
 };
 
--- a/src/UI/Overflow.re
+++ b/src/UI/Overflow.re
@@ -66,9 +66,7 @@ let _startClipRegion =
   Glfw.glScissor(x, y, width, height);
 };
 
-let _endClipRegion = () => {
-  Glfw.glDisable(GL_SCISSOR_TEST);
-};
+let _endClipRegion = () => Glfw.glDisable(GL_SCISSOR_TEST);
 
 let render =
     (
@@ -81,7 +79,13 @@ let render =
       r: renderCallback,
     ) => {
   if (overflow == LayoutTypes.Hidden || overflow == LayoutTypes.Scroll) {
-    _startClipRegion(worldTransform, dimensions, screenHeight, pixelRatio, scaleFactor);
+    _startClipRegion(
+      worldTransform,
+      dimensions,
+      screenHeight,
+      pixelRatio,
+      scaleFactor,
+    );
   };
 
   r();
--- a/src/UI/Overflow.re
+++ b/src/UI/Overflow.re
@@ -52,15 +52,12 @@ let _startClipRegion =
   let maxX = Vec2.get_x(bbox.max);
   let maxY = Vec2.get_y(bbox.max);
 
-  /**
-     TODO: Should scaleFactor be applied to the values below as well as pixelRatio
-     let multiplier = pixelRatio *. float_of_int(scaleFactor)
-   */
-  let x = int_of_float(minX *. pixelRatio);
+  let multiplier = pixelRatio *. float_of_int(scaleFactor);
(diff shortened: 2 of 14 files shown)


Akin909:
@bryphe I could use a hand here as I don't entirely understand how were using `pixelRatio` and how to adjust screen space with the scale factor. You suggested diving window size by the scale factor? (briefly tried that but things become unreadable), not clear on the flow

src/UI/UiRender.re
```diff
@@ -31,6 +31,12 @@ let render = (container: UiContainer.t, component: UiReact.syntheticElement) =>
   /* Layout */
   let size = Window.getSize(window);
   let pixelRatio = Window.getDevicePixelRatio(window);
+  let scaleFactor = Monitor.getScaleFactor();
+  print_endline(
+    "Dimension =======================" ++ string_of_float(scaleFactor),
+  );
+  let _adjustedHeight = size.height / int_of_float(scaleFactor);
+  let _adjustedWidth = size.width / int_of_float(scaleFactor);
```
faisalil: why are you calculating those if you don't need them?

Akin909: The PR is a WIP @faisalil I raised it quite early to discuss it's nowhere near done

src/Core/Monitor.re
```diff
@@ -9,6 +9,20 @@ type size = {
...
+
+let getScaleFactor = () => {
+  let monitor = Glfw.glfwGetPrimaryMonitor();
+  let vidMode = Glfw.glfwGetVideoMode(monitor);
+  let physicalSize = Glfw.glfwGetMonitorPhysicalSize(monitor);
+  let dpiH = calculateDPI(vidMode.height, physicalSize.height);
+  let dpiW = calculateDPI(vidMode.width, physicalSize.width);
+  let avgDPI = dpiW +. dpiH /. 2.0;
+  let scaleFactor = avgDPI /. 96.0;
+  scaleFactor >= 1.0 ? scaleFactor : 1.0;
```
bryphe: I think that, because the scale factor is not 'precise' - we'll want to round it (for now, perhaps to an integer like `1`, `2`) - but perhaps when we have a more deterministic API to get the scale factor from, we could do fractional scale as well.

bryphe: Thanks for accounting for the less-than-1 case!

src/UI/UiRender.re
bryphe: Yes, this is what I was thinking - and then we'd just these adjusted scale factors 
```
    rootNode#setStyle(
      Style.make(
        ~position=LayoutTypes.Relative,
        ~width=size.width,
        ~height=size.height,
        (),
      ),
    );
```

and here:
```
   Mat4.ortho(
      _projection,
      0.0,
      float_of_int(size.width),
      float_of_int(size.height),
      0.0,
      1000.0,
      -1000.0,
    );
```

bryphe: Changing that orthographic matrix at the bottom sets up our render frame to go from `0, 0` -> `width`, `height` - so if we set the value to a smaller value, it will automatically get scaled up.

bryphe: A couple other places we might need to change, too:
- We might need to apply a transform (dividing by the scale factor) to mouse events
- We'll need to pass this to the `Node` hierarchy (like how we pass the `devicePixelRatio`) - for font rendering, we'll want to multiply the font size by the scale factor when it loads the font (otherwise, the fonts will be blurry)

bryphe:
> (briefly tried that but things become unreadable), not sure I follow how changing the _window_ size will impact the the scale of the elements we draw, I'd have imagined that we want to scale everything up by the factor so whereas we'd by drawing a 10 * 10 box with a SF of 1 we want a 4(10 * 10) with a SF of 4

Yep, that's exactly what we want to do! There's two ways of doing this:
- Multiply everything by SF of 4
- Make the window smaller by SF of 4

The latter approach is what I'm thinking about - if, for example, our window is 400x400, with a SF of 4, but we pretend that it is 100x100 big (using the same size window) - we'll implicitly get that scale. That's what tweaking the `ortho` matrix does. And we also have to set the root node dimensions to the `100` so that layout happens in the right space.

This approach saves us from having to multiply _every_ element. The downside though is we'll have to also transform mouse events (that'd be in the non-scale-factored space - so a mouse event at 396, 396 would need to be transformed to 99, 99 to give us the right hit-test behavior).

One other thing we need - is we need to add this to our `NodeDrawContext` and make sure that the TextNode applies the scale factor when it loads the glyph. The problem is, if we have a SF of 4, and we say fontsize=16, and we get a 16px by 16px glyph, we're going to scale that up to 64px x 64px which will look really blurry. So we actually have to ask the FontRenderer to render a `64` glyph. We do this already for devicePixelRatio, but we'll need to incorporate scaleFactor into the calculation as well.

Hope that helps!

This PR also has some info about how the coordinate spaces relate to each other that might be useful: https://github.com/revery-ui/revery/pull/190

Akin909:
@bryphe With the changes I get

scalefactor

Akin909:
Ok some progress, some elements still don't seem to be size properly but the text is rendering at the right size now

Akin909:
@bryphe not sure but the issue I'm now seeing with the squished side bar I think relates to #287 since actually most of the examples look correct, I've never seen them so big but they look crisper. Although the scroll view seems to be incorrectly sized whereas most other UI elements look normal

Akin909:
Oh might have spoken too soon the `input`,`dropdown` and `tree` aren't rendering their content correctly

src/UI/Overflow.re
```diff
@@ -51,19 +52,18 @@ let _startClipRegion =
...
 
-  let y = int_of_float(pixelRatio *. (float_of_int(screenHeight) -. maxY));
-  let width = int_of_float(pixelRatio *. (maxX -. minX));
-  let height = int_of_float(pixelRatio *. (maxY -. minY));
+  let y = int_of_float(multiplier *. (float_of_int(screenHeight) -. maxY));
+  let width = int_of_float(multiplier *. (maxX -. minX));
+  let height = int_of_float(multiplier *. (maxY -. minY));
 
   Glfw.glEnable(GL_SCISSOR_TEST);
   Glfw.glScissor(x, y, width, height);
```
bryphe: It might be simpler to leave the previous calculations as-is, and then multiply by the scaleFactor here:
```
Glfw.glScissor(x * scaleFactor, y * scaleFactor, width * scaleFactor, height * scaleFactor);
```

src/UI/UiRender.re
```diff
@@ -68,14 +71,15 @@ let render = (container: UiContainer.t, component: UiReact.syntheticElement) =>
...
+      float_of_int(adjustedWidth),
+      float_of_int(adjustedHeight),
       0.0,
       1000.0,
       -1000.0,
     );
 
-    let drawContext = NodeDrawContext.create(pixelRatio, 0, 1.0, size.height);
+    let drawContext =
+      NodeDrawContext.create(pixelRatio, 0, 1.0, size.height, scaleFactor);
```
bryphe: The `size.height` needs to be replaced with `adjustedHeight` here

Akin909: That fixes most of the examples, tree component still doesn't render although I'm beginning to suspect that maybe its off screen based on how I initially positioned it

src/UI/Overflow.re
Akin909: hmmm doing that seems not to hide the "overflown" content anymore I'll poke and try and figure out what's going on

Akin909: Actually to be clearer the examples list renders fine with that arrangement but the `scrollview` example does not

bryphe: Hmm, I tested it in this commit and it seemed to work - would you mind trying out my changes to `Overflow.re` here? https://github.com/revery-ui/revery/commit/5a29b01b8e47fd0030c1bbb88ea05ffa5b6cf7c9

bryphe: (If you do `git fetch origin` you should be able to `git checkout 5a29b01 src/UI/Overflow.re` to pick them up on your branch)

src/UI/UiRender.re
bryphe: Ah ya... It looks like the Tree example uses `Window.getSize`, which we'd have to divide by `scaleFactor` to get the real values.

Alternatively - I think it'd make sense just to 'center' the example by using a flexbox

Akin909:
okay @bryphe think it actually works this time :+1:

src/Core/Monitor.re
```diff
@@ -9,6 +9,20 @@ type size = {
...
+let calculateDPI = (modeDimension, physicalDimension) =>
+  float_of_int(modeDimension) /. (float_of_int(physicalDimension) /. 25.4);
+
+let getScaleFactor = () => {
+  let monitor = Glfw.glfwGetPrimaryMonitor();
+  let vidMode = Glfw.glfwGetVideoMode(monitor);
+  let physicalSize = Glfw.glfwGetMonitorPhysicalSize(monitor);
+  let dpiH = calculateDPI(vidMode.height, physicalSize.height);
+  let dpiW = calculateDPI(vidMode.width, physicalSize.width);
+  let avgDPI = dpiW +. dpiH /. 2.0;
```
bryphe: Ah, one more bug I saw trying the latest - seems like this should actually be `(dpiW +. dpiH) /. 2.`. Without the parentheses - we'll get an artificially large result. Does it still work OK for you after making this change?

bryphe:
Just tried out the latest, @Akin909 - awesome ! Thanks for all the work on this!

I saw one issue - it seemed there was a bug in the `avgDPI` calculation - I added a comment above. This was causing the scale factor to be quite large on my Surfacebook (4x instead of 2x). 

Other than that - looks good 👍 I'm excited about this change since Oni2 will be usable on my Surfacebook without a magnifying glass, lol

src/Core/Monitor.re
Akin909: @bryphe nice spot, it was on the big side, I've looked at it again and noted something, locally I calculate a `dpi` of 2.94 as a float but on converting to using `int_of_float` I get 2, which seems drastic to me as that more or less completely changes the value.

So I applied rounding to the value before converting it. So locally I get `3`. This "feels" more like other apps to me, although my settings say my display is scaled by `2`... I stuck with the rounding since changing a value from 2.93 -> 2 seemed incorrect unless that might be the practice in general (maybe that's what my gnome setup does under the hood :thinking:)

Akin909: There's definitely a noticeable difference scaling going from `2` to `3` so I'd like to get as accurate view as possible so when I'm sizing things they don't end up being completely wrong on someone else machine

Akin909: Ok change of tune I figured that when we switch to native derivation of the `scaleFactor` at least in my case the value will be `2`, and from running revery on the web, it seems that the scaling more closely matches `2` rather than `3` :man_shrugging: lemme know what you think otherwise this is ready to merge

bryphe: Looks good to me, @Akin909 ! I got a scale factor of '2' on my surfacebook which is much better 👍 

Note that these DPI calculations are an approximation - I think the next step is to actually integrate the scaling APIs the platform provides (for example, Windows has an API to get the scaling amount - sounds like Linux does, too!), so we can get a more deterministic value. But that can happen later - this work greatly improves the state of the world for Revery, and having the 'scaleFactor' plumbed through is the hard part 👍

PR #328, opened 2019-02-11
[WIP] Dependency: brisk-reconciler->1b8f321

bryphe:
This upgrades Revery to use the latest `brisk-reconciler`, which changes (streamlines) the API a bit:

- A separate `React.element` call is no longer needed thanks to @jchavarri 's work - going through and updating these!
- Components now return a tuple of `(hooks, <JSX />)`  thanks to @wokalski 's work 

TODO:
- Rename `slots` -> `hooks`
- Remove additional `make`

### DIFF
--- a/examples/AnalogClock.re
+++ b/examples/AnalogClock.re
@@ -43,7 +43,7 @@ module AnalogClock = {
 
   let component = React.component("AnalogClock");
 
-  let make = () => {
+  let createElement = (~children as _, ()) => 
     component(slots => {
       let (state, dispatch, slots) =
         React.Hooks.reducer(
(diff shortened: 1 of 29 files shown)


bryphe:
Setting state seems to be broken after converting (for the `Examples` build - both the `ScrollView` and `Clickable` are broken, because it seems the `ref` we're trying to store here:
```
    let (clickableRef, setClickableRefOption, slots) =
      React.Hooks.state(None, slots);
```

Isn't being persisted. I see that we are setting it, and that we're getting marked as stale, but when we go to render again, we still have `None`. Need to dive in a bit more and see if I can get more a minimal repro.

wokalski:
@bryphe I’m on it

wokalski:
@bryphe I updated tests to test setState (it appears we only used `reducer`s in the tests.) It should work now!

bryphe:
Awesome, thank you @wokalski - much appreciated!

bryphe:
Closing in lieu of #331

bryphe:
FYI @wokalski - tested with #331 and everything is working! Thank you! 🎉

Issue #330, opened 2019-02-12
Undefined references on Linux


zxey:
I get undefined references to couple of function that glfw3 needs.
The concerning functions are located in Xinerama and Xcursor libraries.
This is unfortunately a problem with reason-glfw which does not specify correct linker flags.

The log when building revery:
```
info esy build 0.5.6 (using package.json)
info building reason-fontkit@2.1.0@d41d8cd9
error: build failed with exit code: 1
  build log:
    # esy-build-package: building: reason-fontkit@2.1.0
    # esy-build-package: pwd: /home/richard/.esy/3__________________________________________________________________/b/reason_fontkit-2.1.0-c3a70403
    # esy-build-package: running: 'dune' 'build' '--root' '.'
             gcc src/fontkit.o
    fontkit.cpp: In function 'hb_font_t* get_font_ot(const char*, int)':
    fontkit.cpp:75:12: warning: ignoring return value of 'size_t fread(void*, size_t, size_t, FILE*)', declared with attribute warn_unused_result [-Wunused-result]
           fread(data, length, 1, file);
           ~~~~~^~~~~~~~~~~~~~~~~~~~~~~
        ocamlopt bin_native/test_fontkit.exe (exit 2)
[...]

PR #332, opened 2019-02-13
Dependency: Upgrade reason-fontkit and reason-glfw

bryphe:
This upgrades `reason-glfw` to 3.2.1015, and `reason-fontkit` to 2.2.0, which picks up @zxey 's fix for #330 .

Thank you @zxey for your investigation and fix!


bryphe:
Fixes #330

PR #336, opened 2019-02-13
API: onDimensionsChanged event

bryphe:


bryphe:
Tested this in Oni2's branch (needed it for editor resizing) - seems like it works 👍

PR #340, opened 2019-02-15
Build: Publish release packages 

bryphe:


Et7f3:
I think this branch can be safely removed @Akin909 ?

Issue #339, opened 2019-02-14
Symbol Lookup Error with Harfbuzz


zbaylin:
I'm running
- Fedora 28
- Esy 0.5.6

and am unable to run any revery project (including the quick start) because of a symbol lookup error with Harfbuzz.

Here is the output I get when I run `esy run` in the quick-start project:
```
/home/baylinz/Development/revery-quick-start/_esy/default/store/i/revery_quick_start-c2246e4c/bin/App: symbol lookup error: /home/baylinz
/Development/revery-quick-start/_esy/default/store/i/revery_quick_start-c2246e4c/bin/App: undefined symbol: hb_blob_create_from_file  
```
Upon further investigation, `hb_blob_create_from_file` is a Harfbuzz function [documented here](http://www.manpagez.com/html/harfbuzz/harfbuzz-1.8.6/harfbuzz-hb-blob.php).

Both `esy install` and `esy build` run without error. This might (in fact, it's probably likely) to be an error on my end, but I am unsure how to proced from here. Any help is greatly appreciated!

zbaylin:
Ok I have fixed it by upgrading to Fedora 29, but I think there should be some note in the README that it requires Harfbuzz 1.7.7+. I can make a PR for this.

PR #345, opened 2019-02-17
Add harfbuzz to list of dependencies and a note about the minimum required version

zbaylin:
This is to prevent any issues in the future, like #339 with Fedora 28. @bryphe noted that this could also be fixed with linker flags, but I think this is a good (temporary) solution.


OhadRau:
This is under the Ubuntu section, I think it would benefit from being in a Fedora specific section @zbaylin

zbaylin:
@OhadRau I added a Fedora section to the README with a note about harfbuzz

PR #348, opened 2019-02-18
Fix #347: High-dpi display text rendering issues

### LINKED ISSUE #347: Regression on example app in high-dpi displays
Thanks @faisalil for the screenshot! 

image

It looks like we are wrapping text aggressively in high-DPI displays. This might be due to the 'scaling' we do (where we render the glyphs at a larger size to account for the pixel ratio)

bryphe:
There were a couple of issues with text-rendering on high-dpi displays. These stem from the fact that if the `pixelRatio` is `2`, `3`, we actually request a glyph thats 2x or 3x bigger than the screen-space coordinates - otherwise, we'll end up with a font glyph that is low-resolution. On a Macbook retina display, for example, if our screen-space Window is 800x600, it might actually have a framebuffer size of 1600x1200 or 2400x1800. 

Then, if we render a glyph for font size 14px, we actually want to render it at either a 28px or 42px size, so we get the full glyph resolution. This is challenging for layout because it means the font metrics we get are actually scaled up, which we actually have to scale back down by dividing by the pixel ratio.

__Fixes:__
1) The `lineHeightPx` method wasn't taking into account the measured height of the font (which could be different than the actual pixel height) 
2) The `measureWidth` was using the up-scaled font size, so we need to convert that back down 
3) With the above fixes for 1/2, the pixelRatio was applied again for the size of the measured box - so we need to remove those applications

### DIFF
--- a/src/UI/TextNode.re
+++ b/src/UI/TextNode.re
@@ -195,6 +195,6 @@ class textNode (text: string) = {
   };
   pri _getLineHeightPx = pixelRatio => {
     let style = _super#getStyle();
-    style.lineHeight *. float_of_int(style.fontSize) *. pixelRatio;
+    style.lineHeight *. float_of_int(style.fontSize) /. pixelRatio;
   };
 };
--- a/src/UI/TextNode.re
+++ b/src/UI/TextNode.re
@@ -146,7 +146,7 @@ class textNode (text: string) = {
         let (lines, maxWidthLine) =
           TextWrapping.wrapText(
             ~text,
-            ~measureWidth=str => FontRenderer.measure(font, str).width,
+            ~measureWidth=str => int_of_float(float_of_int(FontRenderer.measure(font, str).width) /. pixelRatio),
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #347

Akin909:
This is great, with this I should be able to progress with #325

Akin909:
NICE :100: incorporated your PR @bryphe and #325 now works :heart_eyes: I can see `revery` properly (side note the derived `scaleFactor` is probably not exactly right as things are a little big but it is a huge improvement

bryphe:
@Akin909 - that's awesome! 🎉 😎 I'll take a look at #325 in a bit - thanks for your work on it!

faisalil:
tested it. Its working now!

bryphe:
Awesome! Thanks @faisalil for confirming!

saitonakamura:
Thank you, @bryphe , for fixing my bugs! :)

PR #349, opened 2019-02-18
Cleanup: Remove Cube from Geometry module

bryphe:
And clean up the verbose `Revery.Geometry.Geometry` module


bryphe:
(`Cube` is a leftover from when Revery was a game engine, so moving it elsewhere 😎 )

PR #360, opened 2019-02-21
Add the option for a custom window icon

zbaylin:
This is a follow up from bryphe/reason-glfw#97.
I implemented it in revery by adding an `icon: option(string)` field to the `windowCreateOptions` in `Window.re`. The default is `None`. An example of this in practice can be found in the examples project, i.e. `esy x Examples`.


bryphe:
Thanks @zbaylin for your work on this - awesome!

so cool to have the icon now for the example app:
image

PR #363, opened 2019-02-23
Update scroll bounce logic to use reducer

### LINKED ISSUE #346: Regression on scroll bounce after immutable hooks change in brisk-reconciler
The scroll bouncing effect is broken in `master`. The underlying cause I believe is the same as https://github.com/revery-ui/revery/issues/342. 

There is an animation for the scroll bouncing where a callback is passed that updates the scroll top position with every new value:

https://github.com/revery-ui/revery/blob/e14abb351c6a7b79659ce50fc3b41cb8af709718/src/UI_Components/ScrollView.re#L165

This model, which assumes that the updater reference stays constant, doesn't work anymore with immutable hooks, because for every new value the updater function `setScrollTop` will be "obsolete", and it will update a memory position that is not the right one anymore.

I tried to use the effects like in #343, but I'm not sure how it should be implemented. @wokalski @bryphe Do you have any hints on how this could be solved? It feels to me like the immutable hooks in brisk-reconciler introduce a quite important divergence in the way React hooks work that could make quite a lot existing patterns and idioms obsolete. What are your thoughts about this?

jchavarri:
Fixes #346.

@bryphe Let me know what you think :)

### DIFF
--- a/src/UI_Components/ScrollView.re
+++ b/src/UI_Components/ScrollView.re
@@ -161,9 +161,6 @@ let make =
               repeat: false,
               easing: Animated.cubicBezier(0.23, 1., 0.32, 1.),
             };
-            let updateValue = v => {
-              dispatch(ScrollUpdated(int_of_float(v)));
-            };
             let playback =
               tween(
                 floatValue(float_of_int(actualScrollTop)),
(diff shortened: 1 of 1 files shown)


bryphe:
Looks great @jchavarri ! The bounce is back 😎 

Thanks for switching this to the `reducer` hook. I need to remember to use `reducer` over `state` (and would be good to add to our docs, at some point).

PR #373, opened 2019-02-26
Subset of glyph atlas changes 

bryphe:
This is a subset of the changes in #265 , to allow for the new image / texture Bigarray representation.

Verified working in both native and JSOO builds 👍


bryphe:
Phew, got this in! Thanks for your work on it, @manuhornung  👍

PR #377, opened 2019-03-02
[WIP] Fix #109: Implement basic subpixel rendering

### LINKED ISSUE #109: Text Rendering / Clarity: Investigate subpixel rendering strategies
__Issue:__ On Low DPI displays, the text rendering is not as clear as it could be.

One common technique for dealing with this is _subpixel rendering_, which exploits the fact that LGB displays [tend to have strips in RGB order](https://en.wikipedia.org/wiki/Subpixel_rendering).

It's a strategy for increasing the resolution of font rendering by taking advantage of this pixel geometry.

@cryza did some amazing work in Oni to set up a full-WebGL based subpixel rendering strategy: https://github.com/onivim/oni/pull/2120

The idea, as I understand it, was to render the same glyph 4 times (for each subpixel offset case), and then pick the appropriate one based on the pixel offset of the actual glyph. It would be great to have a similar strategy here, in combination with #108 - we'd have some really sharp font rendering 😄

bryphe:
This change adds subpixel rendering for low DPI displays to improve the quality of the rendered text (important for editors like Onivim 2). The basic strategy is to render the font glyph at 3x resolution, and then use a kernel to sample each color channel individually. 

The chosen strategy only works with solid color backgrounds, which is an acceptable trade-off for now (if there is a non-solid color background, we'll fallback to a non-subpixel strategy).

The initial kernel is very simple - sample each channel subpixel against the high-resolution glyph. There's better choices of kernel potentially (that require more texture reads), we can experiment and see what works well.

This fixes #109 

__TODO:__

- [x] Conditionally enable subpixel rendering _only_ when low DPI _and_ background color is solid
- [ ] Investigate using `FT_Render_Mode_LCD` with FreeType, along with various filters
- [ ] Remove hardcoding of scaling
- [ ] Handle pixel alignment - make glyph quad always pixel-aligned, but adjust subpixel offset for text (test in example)
- [ ] Try out different colors in example
- [ ] Add option in example to enable / disable subpixel rendering
- [x] Should we be using `GL_NEAREST` instead of `GL_LINEAR` for sampling? _Nope_
- [x] Experiment with alternate kernels - the 5-tap filter here looks promising: https://www.grc.com/cttech.htm
- [ ] Investigate filter options with Freetype (https://www.freetype.org/freetype2/docs/reference/ft2-lcd_filtering.html)


bryphe:
Will close this out to be superseded by the Skia effort

Issue #319, opened 2019-02-09
Playground: Live-editing code


bryphe:
Today, our [playground](https://www.outrunlabs.com/revery/playground/) is static - the code shown alongside the examples is _read-only_. This is unfortunate, because it is so tempting to edit it!

@thangngoc89 showed the Reason world how this works with [sketch.sh](https://sketch.sh/)- based off the incredible work on the `js_of_ocaml` toplevel - a demo of it in action here: http://ocsigen.org/js_of_ocaml/toplevel.html#version=4.06.0

It seems like we'll need to expose an entry point for the playground with [`JsooTop`](https://github.com/ocsigen/js_of_ocaml/blob/c2a5c19163b757153ec5d408af773789fca022cb/toplevel/lib/jsooTop.mli) provided by the `js_of_ocaml-toplevel` package. 

Each example is a module that provides a `render: unit => component` function. We'd need to figure out how to execute that 'function' in the context of the top-level, and wire up the example to read from it.

One concern is that it is easy to write code like `while(true) { ... }` that could halt / stop the browser. `sketch.sh` sandboxes this by running the code in a worker, but unfortunately I don't think that is (trivially) workable for us. Perhaps there is a way we could extract that `render` function to a worker?

Some additional links I found interesting while learning / pursuing this space:
- Documentation for Jsoo top level: https://ocsigen.org/js_of_ocaml/3.1.0/api/JsooTop
- @thangngoc89 's sketch blog posts:
   - [Part 1: Introduction](https://khoanguyen.me/sketch/)
   - [Part 2: The Engine](https://khoanguyen.me/sketch/part-2-the-engine/)

@thangngoc89 's posts are a must read - they go over other critical pieces, like how to interpret errors, how to support ReasonML syntax, etc.

Some other items to think about:
- We'll need to 'invert' the flow for running the examples. Right now, we compile the examples and run them - but in this new model - we wouldn't need to 'pre-compile' the examples - we'd compile them in the browser and dynamically call that `render` method some way .
- We'll need to figure out how to handle errors. If we can get line numbers - the monaco editor supports error decorations that can be helpful.
- We'll need to update our Monaco editor initialization logic here: https://github.com/revery-ui/revery/blob/7ff96ebab393c1d76f7cbda4d82900c8a7bb0ace/playground/src/index.html#L76 so that it isn't read-only, and that we respond to code changes.

Build instructions for the playground are available here: https://github.com/revery-ui/revery/tree/master/playground

[...]

PR #388, opened 2019-03-05
Fixed Playground: Live-editing code #319

nik72619c:
Fixed the read-only issue with the revery playground editor


bryphe:
Thanks @nik72619c ! Note that this won't be too useful until we implement the rest of the 'live' piece - where we process the update, send to to the [Jsoo top-level]( https://ocsigen.org/js_of_ocaml/3.1.0/api/JsooTop), and can see the results in real-time.

PR #390, opened 2019-03-07
Fix #357: TextNode: render text when dimensions are specified

### LINKED ISSUE #357: UI - Text: If both 'width' and 'height' styles are specified, text does not render
__Issue:__ If you have a text element like this:

```
let textStyle =
  Style.[
    color(Colors.white),
    width(100),
    fontFamily("Roboto-Regular.ttf"),
    fontSize(16),
    height(100),
    textWrap(TextWrapping.NoWrap),
  ];

            <Text
              style=textStyle
              text={"Test"}
            />
```

The text won't be displayed.

__Defect:__ The `TextNode` has a `_lines` ref that needs to be populated for rendering. It assumes that the `measure` function will get called - and that's where we set the `_lines` for it:
https://github.com/revery-ui/revery/blob/767d51ae2eb444f25d539aaf9b1d6de53739cf3f/src/UI/TextNode.re#L175

However, in cases where Layout decides not to measure (like, if the `width` and `height` are both specified), the `_lines` ref won't get set - so the text will remain empty.

A 'simple' fix would be just to use the `text` property in the case where `_lines` isn't set - however, that won't properly handle the wrapping scenarios. We might need to use the wrapping logic in render too, in the case that `_lines` hasn't been set.

romgrk:
screenshot from 2019-03-07 15-19-37

A few questions:
 - Should the fixed text be cropped to its dimensions?
 - Should I add a test for this? If so, how?

### DIFF
--- a/src/UI/TextNode.re
+++ b/src/UI/TextNode.re
@@ -13,6 +13,7 @@ class textNode (text: string) = {
   as _this;
   val mutable text = text;
   val mutable gamma = 2.2;
+  val mutable _isMeasured = false;
   val _lines: ref(list(string)) = ref([]);
   inherit (class viewNode)() as _super;
   pub! draw = (parentContext: NodeDrawContext.t) => {
@@ -30,6 +31,11 @@ class textNode (text: string) = {
 
     let lineHeightPx =
       Text.getLineHeight(~fontFamily, ~fontSize, ~lineHeight, ());
+
+    /* when style.width & style.height are defined, Layout doesn't call the measure function */
+    if (!_isMeasured) {
+      _this#measure(style.width, style.height) |> ignore;
+    };
 
     List.iteri(
       (lineNum, line) =>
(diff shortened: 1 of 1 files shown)


bryphe:
This looks great, @romgrk ! Thanks for fixing this - I like the way you refactored the code 👍 

Regarding your questions:

> Should the fixed text be cropped to its dimensions?

I think the behavior you show in your screenshot is reasonable - especially since @Akin909 implemented text overflow in #392 , and we have the ability to add `overflow(Hidden)` on the parent.

> Should I add a test for this? If so, how?

Really appreciate you thinking about this! 💯 Unfortunately there is a blocking bug on our CI machine - i tracked in #400 - the Azure CI environments don't support creating an OpenGL context (at the moment). Since the test would need to touch the `draw` method, which makes OpenGL calls, testing that portion of the stack isn't doable until we get #400 sorted out. But I think what you have right now is reasonable.

I'll bring in this as it looks good overall - thanks for the help, @romgrk ! Really appreciate the contribution.

PR #392, opened 2019-03-08
Feature/text overflow

### LINKED ISSUE #389: UI Styles: Text-overflow: ellipsis +/- clip +/- specifiable character
Currently we have line wrapping with text that is too long, we should also offer the user the opportunity to have overflowing text cut-off and render an ellipsis at that point similar to [css](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow).

We could go with this as an initial starting point or optionally allow the same variability to specify the character to render for the clipped text

Akin909:
fixes #389 

text-overflow

### DIFF
--- a/src/UI/Style.re
+++ b/src/UI/Style.re
@@ -79,6 +79,7 @@ type t = {
   padding: int,
   paddingVertical: int,
   paddingHorizontal: int,
+  textOverflow: TextOverflow.t,
   overflow: LayoutTypes.overflow,
   borderTop: Border.t,
   borderLeft: Border.t,
@@ -95,6 +96,7 @@ type t = {
 
 let make =
     (
+      ~textOverflow=TextOverflow.Overflow,
       ~backgroundColor: Color.t=Colors.transparentBlack,
       ~color: Color.t=Colors.white,
       ~width=Encoding.cssUndefined,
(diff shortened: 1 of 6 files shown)


src/UI/Style.re
```diff
@@ -317,7 +320,11 @@ type coreStyleProps = [
 
 type fontProps = [ | `FontFamily(string) | `FontSize(int)];
 
-type textProps = [ | `LineHeight(float) | `TextWrap(TextWrapping.wrapType)];
+type textProps = [
+  | `LineHeight(float)
```
Akin909: feel like there should/is a way to encode a constraint here so that both `TextWrap` and `TextOverflow` aren't allowed at the same time

src/UI/TextNode.re
```diff
@@ -59,74 +59,115 @@ class textNode (text: string) = {
...
+
+    let width = measure(formattedText);
+    let isOverflowing = width >= maxWidth;
+
+    let handleOverflow =
+      TextOverflow.handleOverflow(~maxWidth, ~text=formattedText, ~measure);
+
+    let truncated =
+      switch (textOverflow, isOverflowing) {
+      | (Ellipsis, true) => handleOverflow()
```
bryphe: Awesome! Thanks for thinking through these cases, @Akin909 💯

bryphe:
Sorry I just brought in #390 which introduced a merge conflict with this. Let me know if you need any help resolving it.

Issue #316, opened 2019-02-08
esy build fails on Ubuntu 18.04.1 LTS


sslotsky:
Amazing project, very excited to build some apps with it! Sadly, build fails with the error below. I've installed all packages listed in the README. I searched for the error and saw a suggestion to install `libgtk2.0-dev` or `libgtk-3-dev`. Neither of these resolve the error but perhaps I'm missing something from my `$PATH`?

```
sam@sam-ThinkPad-T480s:~/revery$ esy build
info esy build 0.5.6 (using package.json)
      ocamlc src/Native/dialog_gtk.o (exit 2)
(cd /home/sam/revery/_esy/default/store/b/revery-5c7dd45f/default/src/Native && /home/sam/.esy/3______________________________________________________________________/i/ocaml-4.7.1004-a6ae015e/bin/ocamlc.opt -g -I /home/sam/.esy/3______________________________________________________________________/i/opam__s__lwt-opam__c__4.1.0-62ec7e5a/lib/lwt -I /home/sam/.esy/3______________________________________________________________________/i/opam__s__ocamlfind-opam__c__1.8.0-eecbed0c/lib/bytes -I /home/sam/.esy/3______________________________________________________________________/i/opam__s__result-opam__c__1.3-1fdcd9f3/lib/result -I /home/sam/.esy/3______________________________________________________________________/i/reason_gl_matrix-0.9.9302-3a6eaf9d/lib/reglm -I /home/sam/.esy/3______________________________________________________________________/i/reason_glfw-3.2.1013-1dfe7dca/lib/reglfw -o dialog_gtk.o dialog_gtk.c)
dialog_gtk.c:2:10: fatal error: gtk/gtk.h: No such file or directory
 #include <gtk/gtk.h>
          ^~~~~~~~~~~
compilation terminated.
error: command failed: 'dune' 'build' '--root' '.' '-j4' (exited with 1)
esy-build-package: exiting with errors above...
error: build failed with exit code: 1
  
esy: exiting due to errors above
```

Akin909:
@sslotsky thanks for reporting this :+1: , I recently added the initial bits for using gtk for native widgets which has gtk as a dependency. I assumed that most linux distros bundled gtk especially ubuntu which uses gnome by default :thinking: 

Can I ask what happens when you run `pkg-config --cflags gtk+-x11-3.0` if it prints out a list of directories do those exist if so the build should be working :face_with_head_bandage:, also installing gtk-3 didn't work?

[...]

PR #393, opened 2019-03-10
Add notes in readme about ubuntu install issues.

JasoonS:
This fix includes the advice from two issues:

https://github.com/revery-ui/revery/issues/316#issuecomment-462505096

and 

https://github.com/revery-ui/revery/issues/339 (which is not only an issue with Fedora but also ubuntu <= 18.04 - https://www.ubuntuupdates.org/package_metas?utf8=%E2%9C%93&q=harfbuzz)


bryphe:
Thanks @JasoonS for adding this!

PR #394, opened 2019-03-10
Fix 2 simple errors that occur in build script: `where` not found and `Source and destination must not be the same`

JasoonS:
image

image


bryphe:
Excellent! Thanks for your help cleaning this up, @JasoonS ! Looks great 👍

JasoonS:
It's my pleasure, thanks for pushing this project forward!

PR #406, opened 2019-03-18
Make Input component controlled

tatchi:
Hey,

I was playing with the "TodoMVC" example and tried to reset the input when a todo is added. I thought that resetting the `inputValue` state in the `AddTodo` action would make it. But I noticed that the `inputValue` state was not even passed to the `Input` component.

It seems that the `Input` component manages his own state. I changed that to make it a controlled component which makes more sense to me.

If I'm not mistaken, you have both in react:
* Uncontrolled if you don't pass a `value` prop
* Controlled if we pass a `value` prop

I don't know if it really makes sense to have both possibilities ? Even if what I did make sens for you? 

If yes, I've seen that `Checkbox` is also `uncontrolled` (and probably others too) and I would be happy to change them as well :)


Akin909:
@tatchi thanks for the contribution. The Input could definitely use some love. I think `controlled` is definitely a must have but having it be able to be both also provides a lot of flexibility.

I briefly looked into if I could do both but pulled the plug on that due to time constraints. I wonder if you looked into whether or not both were possible i.e. `controlled` and `uncontrolled` modes, so a user doesn't necessarily have to manage the state of the input if they don't care like is the case with the browser?

tatchi:
Hi @Akin909, thanks for your response.

> I wonder if you looked into whether or not both were possible i.e. `controlled` and `uncontrolled` modes, so a user doesn't necessarily have to manage the state of the input if they don't care like is the case with the browser?

No, I didn't try to make both cases possible since I don't really see any use case for `uncontrolled` input. I mean, if you add an `Input` you would surely like to access the value of the input (by storing it in the parent) and process it at a certain point.

But it might be me that miss something. Anyway, I was waiting for someone else's input before going further and I would be happy to try to implement both scenarios.

I won't be able to work on that during the coming days since I'm leaving on holiday for 1 week. But will look at that upon my return.

Akin909:
@tatchi my thinking is to match React and the browser APIs where possible when possible as those are things a lot of people interact with/develop and informs their expectations. I think uncontrolled is a far more common use case for something like this so happy to go with what you've got. I'm going to be working on this soon to do some tweaking so might be good to get this in sooner rather than later so you don't have to rebase it.

Think if you run `esy format` locally you should be able to get the tests passing 👍

tatchi:
Hi @Akin909,

I have been able to further work on it during my flight. I think I came up with a basic implementation that seems to work.

Don't hesitate to make any comments and if you can wait for one week, I would be happy to continue to work on it. Otherwise, don't hesitate to reuse it and continue from that.

Akin909:
@tatchi sorry for the merge issue I needed to make some changes to the component so we could use elsewhere onivim/oni2#167, but lemme know if you'd like some help reconciling these changes

tatchi:
> @tatchi sorry for the merge issue I needed to make some changes to the component so we could use elsewhere [onivim/oni2#167](https://github.com/onivim/oni2/pull/167), but lemme know if you'd like some help reconciling these changes

No problem, will look at that upon my return from holiday and will let you know if I need help :) Thanks

tatchi:
Hi @Akin909,

Let me know what do you think about my implementation. I had to change some things compared to your changes so don't hesitate if you have any suggestion

src/UI_Components/Input.re
```diff
@@ -23,59 +18,46 @@ type changeEvent = {
...
-  | CursorPosition(pos) => {
-      ...state,
-      cursorPosition:
-        getSafeStringBounds(state.inputString, state.cursorPosition, pos),
-    }
-
-  | UpdateText({newString, cursorPosition}) =>
-    state.isFocused
-      ? {cursorPosition, isFocused: true, inputString: newString} : state
-  | Backspace({newString, cursorPosition}) =>
```
Akin909: @tatchi these changes seem like they completely put the responsibility exclusively on the parent component to implement backspace functionality. I think whatever solution we go with it'd be really good to have this work both ways i.e. `uncontrolled` and `controlled` so it could be used either way.

Could a solution be to specify in this component the value as `value?` so that this component checks an `option` but the parent can pass a string or not, if there is `some(parentValue)` then we take the lead from the parent (but set internal state based on the parent)?

src/UI_Components/Input.re
```diff
@@ -96,77 +78,104 @@ let defaultStyles =
...
-              altKey: event.altKey,
-              ctrlKey: event.ctrlKey,
-              shiftKey: event.shiftKey,
-              superKey: event.superKey,
-            });
-          }
-        );
+        switch (valueAsProp) {
+        | Some(v) =>
+          onChange(
```
Akin909: These could be piped together I think rather than called altogether
```suggestion
removeCharacter(v, state.cursorOffset)
|> createChangeEvent
|> onChange
```

src/UI_Components/Input.re
```diff
@@ -203,62 +214,106 @@ let make =
...
         onKeyDown=handleKeyDown
         onKeyPress=handleKeyPress>
-        <View style=viewStyles>
-          ...{
-               hasPlaceholder
-                 ? [cursor, placeholderText] : [startText, cursor, endText]
-             }
-        </View>
+        {hasPlaceholder
+           ? <View style=viewStyles>
```
Akin909: @tatchi switching to JSX rather than elements seems to lead to what looks like duplication with some minor variation, there's also a lot more going on here which seems a bit harder to follow. Why not use the variables that were here before?

tatchi:
Hey,

I have been working recently on this but I'm struggling with the `cursorPosition` logic while we are in a `controlled` mode.

When we work in an `uncontrolled` mode, we handle the input value internally and decide what we do on every keystroke. For instance, when the user types a new character, we add this new character to the input value and we increment the `cursorPosition` by 1 (since we know that we've only added one character).

But while in a `controlled` mode, the input value is handled in the parent component, and we can't really expect what will happen when a new character is typed. In this situation, we fire the `onChange` callback with the new input value but the user might omit to update the parent's state.
Or he might also add twice the character, or add the character after a specific timeout.

Even if these scenarios are probably not conventional, I'm not sure we can assume what the user is going to execute in the `onChange` callback (and therefore what will be the next input value given as prop) and always increment the `cursorPosition`.

I tried to check what React does in these situations and from what I can see, the cursor goes at the end of the input text. You can test it here: https://codesandbox.io/s/r7420m2oko (put the cursor in the middle of the input and add one character).

This is why I tried a different approach: instead of manually updating the `cursorPostion` after every keystroke, I created an `effect` hooks that compare the current input value with the preview one (that I keep as ref) and based on the current `cursorPosition`, I try to guess what happened:

* A character added
* A character deleter through the `backspace`
* A character deleter through the `delete` key

Then based on that I update the `cursorPosition`. The logic is not perfect yet but that seems to work except when typing quickly on the keyboard. In this case, the cursor is located somewhere in the input text where it should be at the end.

Anyway, I'm not sure what should be the correct approach. Maybe these scenarios are not relevant for now and we should go to the point.

What do you think?

Akin909:
@tatchi thanks for the excellent write up of the issue you've face I'll have a closer look over the weekend to see if I have any suggestions if not I'm happy for this to be the next iteration and we can look into a fix down the line 👍

PR #407, opened 2019-03-19
Expose focus module in revery

Akin909:
Expose focus module in `revery` so apps like `Oni2` can use `Focus.focused` to see if any of revery's ui is focused


bryphe:
Looks good to me, @Akin909 ! I'll bring this in and publish a new version of Revery so we can pick it up in Oni2 👍

bryphe:
Published revery@0.14.1 w/ this fix

PR #408, opened 2019-03-20
Feature/ refactor input

Akin909:
input

This PR changes a few things in the Input:
* fixes positioning of the cursor
* adds ability to navigate text with the arrow keys
* autofocus - add the ability to specify autofocus if true the element is immediately in focus
* event - return an event type containing info about the pressed key

### DIFF
--- a/examples/InputExample.re
+++ b/examples/InputExample.re
@@ -44,6 +44,7 @@ module Example = {
               backgroundColor(Colors.paleVioletRed),
               color(Colors.white),
               margin(20),
+              height(50),
               boxShadow(
                 ~xOffset=-5.,
                 ~yOffset=2.,
--- a/examples/InputExample.re
+++ b/examples/InputExample.re
@@ -33,14 +33,14 @@ module Example = {
         <View style=containerStyle>
           <Input
             placeholder="Insert text here"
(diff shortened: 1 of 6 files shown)


examples/InputExample.re
```diff
@@ -33,17 +33,20 @@ module Example = {
...
+            onChange={({value, _}) => setValue({first: value, second})}
           />
           <Input
             placeholder="custom input"
             placeholderColor=Colors.plum
             cursorColor=Colors.white
-            onChange={(~value) => setValue({first, second: value})}
+            autofocus=true
+            onChange={({value, _}) => setValue({first, second: value})}
+            onKeyDown={event => Console.log(event)}
```
bryphe: Is this debug logging still needed?

bryphe:
I'll bring this in so we can unblock the menu / quickopen work in Onivim 2 😎  Thanks @Akin909 !

examples/InputExample.re
Akin909: oh sorry @bryphe it was actually a not very clear way to show off the functionality 🤣 a better thing would have been to render the event or use deriving to show the event in the example

Issue #353, opened 2019-02-19
UI Styles: Border Radius


Akin909:
It would be nice to be able to draw circular or rounded shapes in revery along the same lines as the `border-radius` style rule in css. I think this would probably take a similar form as @OhadRau's border styles PR #123 

I haven't yet looked into the how in `OpenGL`, but will update this issue if I get time later to research it.

#### Resources
 1. https://stackoverflow.com/questions/5369507/opengl-es-1-0-2d-rounded-rectangle

OhadRau:
Yeah, I looked into this a little bit at the time that I was doing the border styles. The thing that makes this a little bit harder is the way it interacts with the edges of borders: https://jsfiddle.net/Ldykermt/. While we can't just naively use triangles for the corner like in normal borders, the triangle fans should allow this to work as long as we define the midpoint correctly.

I'd be happy to take a stab at this but my current time commitment is very low because of school work.

This document might also be useful: https://www.w3.org/TR/css-backgrounds-3/#propdef-border-radius.

nik72619c:
is this free to work on @Akin909 ?

Akin909:
@nik72619c I think @zbaylin has been working on this for a bit already

glennsl:
A major problem with the current implementation is that it only curves the outside of the border. The inside of the border and the content remains square. This seems to preclude most common scenarios for border-radius. The only use that seems to work reasonably well is when the border is wider than the radius.

PR #409, opened 2019-03-21
Add border radius

zbaylin:
Pursuing #353, I added an Arc primitive, a border radius shader, and a border radius style attribute. Couple issues as of right now:
- No antialiasing
- Since children components aren't rendered with the same shader, the attribute doesn't propagate downward

I think this is definitely fixable, and I can keep looking into it in the near future.


zbaylin:
Ok I have since added both non-border border radius and anti-aliasing. Let me know if there is anything else you want me to add!

PR #410, opened 2019-03-23
Fixes to enable live-editable Playground

bryphe:
The actual code for the playground itself can be found here: https://github.com/revery-ui/revery-playground.

The playground has a unique architecture - part of the app is run in a worker (the editable code!) and it pushes frames and updates to the renderer.

This change does the following:
- Removes the `playground` folder (which now lives in the `revery-playground` repo)
- Fixes a bug in JSOO where fonts were being reloaded every frame. This was because we were pushing up strings from the worker, and they would get a unique hash cache - so our `FontCache` would always show as a miss.
- Add a seam in the NodeFactory where each side of the playground could inject a custom implementation. On the worker, we just add a 'Proxy' node that forwards updates to the Renderer process. In the Renderer process, we create 'real' nodes, but maintain a mapping of worker ids -> renderer ids, so we can pipe updates from the worker to the proper node.

One additional complexity of this architecture is that the _layout_ step requires information that only the renderer has (font metrics, etc). So when we get an update from the worker, we run layout on the renderer thread, and send the measurements back to the worker. 

These measurements are needed because we need to send UI events to the worker to be handled by the editable code (mouse events require knowing the dimensions / metrics).


src/Draw/ImageRenderer.re
```diff
@@ -46,6 +44,7 @@ let getTexture = (imagePath: string) => {
       let imageLoadPromise = Image.load(relativeImagePath);
 
       let success = img => {
+          print_endline ("successfully loaded image");
```
bryphe: Remove this

src/UI/Primitives.re
```diff
@@ -8,6 +8,22 @@
  * https://github.com/bryphe/reason-reactify
  */
 
+type nodeFactory = {
```
bryphe: Move to separate module?

src/UI/Node.re
```diff
@@ -43,8 +43,7 @@ class node (()) = {
   val _lastDimensions: ref(NodeEvents.DimensionsChangedEventParams.t) =
     ref(NodeEvents.DimensionsChangedEventParams.create());
   val _isLayoutDirty = ref(true);
-  val _miniLayout: ref(Dimensions.t) =
-    ref(Dimensions.create(~top=0, ~left=0, ~width=0, ~height=0, ()));
+  val _forcedMeasurements: ref(option(Dimensions.t)) = ref(None);
```
bryphe: Add comment

bryphe: Can we move this to the Playground in the proxyNode override? 🤔

examples/Calculator.re
```diff
@@ -33,7 +31,12 @@ module Column = {
           backgroundColor(Colors.darkGrey),
           flexGrow(1),
         ];
-      (hooks, <View style> ...children </View>);
+      (
+        hooks,
+        <View style onKeyPress={_ => print_endline("KEY PRESS")}>
```
bryphe: Remove

PR #419, opened 2019-03-28
Absolute cursor position

### LINKED ISSUE #415: UI - Input: Cursor should not shift text content
The issue seems to be that the cursor `View` is part of the normal content flow, so when it moves it also shifts the characters around. I checked oni2 and it doesn't have this problem since it [uses absolute positioning for the cursor](https://github.com/onivim/oni2/blob/master/src/editor/UI/EditorSurface.re#L188-L216) and manually calculates the offset. It seems like something similar should be done here for the `Input` component.

tatchi:
My attempt to fix #415


src/UI_Components/Input.re
```diff
@@ -213,18 +213,28 @@ let make =
...
+          startStr,
+        );
       <View
         style=Style.[
           width(2),
-          marginLeft(hasPlaceholder ? inputTextMargin : 0),
+          marginTop((defaultHeight - dimension.height) / 2),
           height(inputFontSize),
+          position(`Absolute),
+          marginLeft(dimension.width + inputTextMargin),
```
bryphe: The cursor looks a bit snug here:
image

Could we add an extra pixel here to give it a little extra breathing room? Here's a system dialog:
image

(Note the extra pixel gap between the 'i' and the cursor)

tatchi:
Hi @bryphe,

Thanks for your feedback. I added the extra 1px and indeed it seems better now! :)

Let me know if you still have any suggestions.

bryphe:
Excellent, looks great @tatchi - thanks for your work on this. Bringing it in now!

Issue #411, opened 2019-03-23
Multi-threaded, pipelined rendering architecture


bryphe:
I've been researching / thinking a bit about how we could migrate to a multi-threaded, pipelined rendering solution.

In today's model, we do a lot of things synchronously that we could be doing in parallel:
- Reconciliation
- Layout
- OpenGL rendering

In particular, even though this would benefit primarily the __native__ builds, I was thinking about this in the context of our JSOO playground. Currently, the compilation is time-consuming, and it's not a great experience to have to wait for it. We should delegate it to a web worker, but the challenge I had was then:
- How do I get the compiled 'render function' out from the top-level? Unfortunately, `Marshal.to_string` doesn't work on a function. I also tried an alternative of marshalling the `React.syntheticElement` that gets created from the render function - but this too had functional values.

Our current, non-pipelined world looks like this:
![image](https://user-images.githubusercontent.com/13532591/54871632-a1f14700-4d74-11e9-8126-794618af48dd.png)

If, however, we could split out the 'reconciliation + layout' steps, and the 'OpenGL rendering' step, that would open us for a few cool things:
![image](https://user-images.githubusercontent.com/13532591/54871636-acabdc00-4d74-11e9-976e-9f26ca123245.png)

We could start processing the next frame while we are rendering the current frame - opening us up for a higher FPS. Note that I keep the reconcilation + layout together for those diagrams..., but perhaps those could be split apart as well. (Care has to be taken with the 'Node' hierarchy because the layout elements are mutable, but the initial processing / diff of  updates could potentially be handled in a separate thread).

The 'draw' tends to be a bottleneck, and by splitting that out to be handling in a separate thread, it could potentially be a big win. This could be facilitated by the Nodes themselves not actually making OpenGL API calls - but an _immutable description_ of the scene. __This would also enable us to serialize these calls from a web worker, so that the worker does most of the logic, but the 'main thread'  handles the actually OpenGL API calls__.

bryphe:
I went down this route a bit in the hopes of getting the live playground interactive - but I hit one more blocker. The `TextNode` requires font access via `fontkit.js`, which doesn't work from a Worker thread today.

PR #421, opened 2019-03-30
[Prototype] Intermediate Representation for Scene

bryphe:
This is some prototyping for #411 to support a threaded renderer, and ability to support running part of Revery apps in a worker thread in JS.

Instead of `Node`'s making direct calls against OpenGL in their `draw` methods - they'd return a set of instructions (an intermediate representation) - an immutable scene tree - that is serializable across native threads / JS workers.

This allows us to run take that immutable scene tree, and hand it off to a renderer thread to do the actual work of making the OpenGL calls (or in the future - Skia calls 😉 )

### DIFF
--- a/src/Draw/IntermediateRepresentation.re
+++ b/src/Draw/IntermediateRepresentation.re
@@ -1,0 +1,63 @@
+/*
+ * IntermediateRepresentation.re
+ *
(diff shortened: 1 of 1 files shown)


bryphe:
Close for now - had to pursue an alternate strategy to get the playground working. This will be nice to have for a threaded draw experience though (can revisit once we have Skia in-place!)

PR #424, opened 2019-03-31
feat: make input-cursor behaviour more consistent

### LINKED ISSUE #422: UI: Cursor animates with opacity and animates while typing
Hey there, I'd like to start with say that this is purely a matter of preference!

-- 

Right now the cursor animates the opacity instead of "hard"-toggling the opacity from 0 to 1, it also keeps animating while typing.

The cursor-animation also doesn't reset when focus switches which can give a bit of a "flickery" feel for lack of better words!

**Proposal:**

Hard toggling the opacity and perhaps add a boolean for "isTyping" to keep the cursor at full opacity while a user is typing. I believe this is the standard behaviour (try in the browser for example)!

Also, I'd be happy to take this if people agree! 🙂

lessp:
**Intent:**

To make the Input-cursor more consistent with other implementations this PR does the following:

- The cursor now "hard-toggles" from full opacity to transparent instead of animating.
- The cursor does not animate while typing or changing the cursor-position.
- The cursor-toggling timer resets on focus and key-events to give a more consistent behaviour while switching input fields for example. 
This differs from the previous implementation where you would get inconsistent ticks depending on where in the timer-animation you would switch focus (if that makes sense 😄)

**Other:**

I went a bit back and forth of how and where to handle the logic (I previously kept more of it in the state, but finally settled on having most of it in one place), so any improvements are more than welcome! 🙌 

revery-cursor

Closes #422


src/UI_Components/Input.re
```diff
@@ -208,6 +233,26 @@ let make =
...
+      let isTyping =
+        lastKeyPress
+        |> (
+          fun
+          | Some(timeAgo) =>
+            timeAgo >= Time.increment(timeAgo, cursorToggleBreakpoint)
+          | None => false
+        );
+
+      (isTyping, isFocused)
```
Akin909: @lessp this looks great thanks for working on this :+1:

minor comment: why return `isTyping` here since in the next block we just ignore it, in which case why calculate it at all (I can't see if maybe it's used elsewhere)?

lessp: Thanks @Akin909, you're right, I can't see it being used anymore! 

Since the the timer is now reset to 0 on every keypress/focus-event and the caret is visible while the timer is <= 0.5  there's no need to keep track of the lastKeyPress anymore, and we still get the consistent behaviour of it ticking for the half second until being invisible again, I believe! 🤔🙌

bryphe:
Just tested this out - between this and #419 the input is really improving 😍  

Thanks for your work and attention to detail on this @lessp - and thanks for reviewing @Akin909 !

Will bring this in after the CI checks for the merge finish 👍

lessp:
Sweet! Happy to help out! 👏

PR #429, opened 2019-04-02
Input:  handle delete key

tatchi:
Another improvement to the `Input` component. This time I tried to handle the delete key.


src/UI_Components/Input.re
```diff
@@ -175,24 +194,24 @@ let make =
...
+          }
+        )
       | Key.KEY_BACKSPACE =>
-        dispatch(CursorPosition(-1));
-        removeCharacter(inputString, cursorPosition)
+        removeCharacterBefore(inputString, cursorPosition)
         |> (
           update => {
-            dispatch(Backspace(update));
+            dispatch(UpdateText(update));
```
tatchi: Didn't really understand why we have a special action for the `Backspace` case.

PR #434, opened 2019-04-05
Fix bug with text wrapping past max width

msvbg:
`TextWrapping.wrapText` can throw an exception when trying to wrap text that's too wide for the container. To be fair, I don't totally understand the code here so I may have messed something up. But the supplied test case throws an exception without this change to TextWrapping.re, so that's something at least.

On another note, it may be worthwhile to look at Jane Street's Base instead of the Ocaml standard library, which throws exceptions like `String.sub` does here.


msvbg:
I am happy to announce that this now passes CI, after using esy 0.5.6.

OhadRau:
Looks like this PR kinda got ignored 😅 Hopefully the work for this is superseded by #685 but gonna keep this open temporarily in-case a similar bug exists in new text-wrapping lib since it's using String.sub as well.

PR #436, opened 2019-04-05
Fix border radius shader

msvbg:
The border radius shader has broken types. I wonder how it compiled before.


bryphe:
Looks great @msvbg - thanks for fixing this!

>  I wonder how it compiled before.

Seems like some systems / drivers are more 'forgiving'... we should have a CI test that exercises this though (perhaps a software implementation like Mesa could be set up to run in a strict mode). We're missing that validation at the moment.

PR #442, opened 2019-04-07
Fix rendering of glyph clusters

msvbg:
Clusters like ligatures currently render incorrectly. Here is a picture of the problem:

sample

In the first case, I'm assuming that the font I'm using cannot render Z͒̀͆ͯaͣ̈̋̐ͤͥ͗l͗ͮ͑ͮ͑ͥ͌g̔o̐͗͌͊ͮ̒.ͬ̒̃̃́ properly. But the version to the right looks far better than the version to the left! In the second case, the `fi` ligature in the typeface messes with the font rendering to cause the `r` to render twice at the end.

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "af496c194d032a30efd3d34a64f0ac0d",
+  "checksum": "95684439452d4263a90a9bb3b4ec638a",
   "root": "revery@link-dev:./package.json",
   "node": {
     "revery@link-dev:./package.json": {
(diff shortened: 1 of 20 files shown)


Akin909:
@msvbg thanks for picking this up and for the PR. I think the CI is failing because you need to run `esy format` on the code to so `refmt` gets autoformats it here and in you're other PR as well :+1:

msvbg:
Thanks @Akin909. Will try this.

msvbg:
@Akin909 It seems that it's still breaking. The CI log is not telling me much, either.

bryphe:
Just tested this out @msvbg - looks great! Thank you for the contribution. 

It looks like the build failure is due to a problem with the lock files - I reverted the lock file changes and it seems like the hygiene check passes. I'll push that to your branch, and then bring this once the build is green.

PR #443, opened 2019-04-07
Implement animation direction

msvbg:
Same as #433, but submitting from a new branch. As in seen in CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction. Hopefully CI will pass this time.


test/UI/AnimationTest.re
```diff
@@ -242,4 +252,68 @@ test("Animation", () => {
...
+        let _playback = Animated.start(myAnimation);
+
+        expect(myAnimation.isReverse).toBe(before);
+        TestTicker.simulateTick(Time.Seconds(1.));
+        expect(myAnimation.isReverse).toBe(after);
+      },
+    );
+  };
+
+  directionTest("animation that does not alternate", `Normal, false, false);
```
bryphe: Thank you for adding these tests, @msvbg ! Much appreciated.

msvbg:
Thanks for sorting out the lockfile issues! I will have a look at my esy setup to see if it's out of date. It's very possible.

msvbg:
Indeed, I was on an old version of esy. Turns out that I had esy installed with both npm and yarn, and the npm version took precedence over the newer version installed with yarn. Should be good now, hopefully.

Issue #156, opened 2018-12-26
Automated Testing: Image Snapshot Tests


bryphe:
Thinking about #145 - for some of these very visual cases, we have no current test coverage. It's important to be able to make changes _safely_ and _confidently_ - so I always think when there is a regression - how can we improve our 'safety net' to catch these?

We're getting to a level of features with background color, text rendering, borders, shadows that it becomes tough to validate all of these in a PR change!

What I'd like to add to our infrastructure is a set of image-based verification tests, that can validate some of these basic scenarios. These would render a simple scene or component, save it as an image, and then compare that image to a snapshot.

This isn't a new idea; tools like [Telerik](https://www.telerik.com/blogs/using-the-new-image-verification-feature) have supported it for a while. Doing googling shows a [RosettaCode](https://rosettacode.org/wiki/Percentage_difference_between_images) problem for an algorithm for this 😄 

The challenge with such tests is making sure they are reliable and easy-to-update. For reliability, it often helps to have a threshold (% of pixels with the same value), or use per-platform snapshots (there might be differences in anti-aliasing, for example). These test suites should be pretty limited and focused on the core set of rendering primitives we have, because they have a maintenance cost. But they can help protect us against regressions.

__Open questions:__
- Is there a library we could use for image-verification tests? Something that could do a bitmap comparison and ideally show some sort of diff (ie, an overlay that shows different pixels in red).
- Which scenarios should we cover?
- When an image-verification test fails, how do we make it easy to act on and know what failed? 
- What additional APIs do we need to take a screenshot of the GLFW window, to compare against the snapshot? It looks like the `glReadPixels` API mentioned in this [StackOverflow Post](https://stackoverflow.com/questions/33757634/creating-screenshot-in-opengl-not-working?rq=1) could help.

__Alternatives:__
One alternative to _image-based snapshot testing_ is _OpenGL API snapshot testing_ - essentially, put a proxy in place for all the `glXXX` calls, that record the inputs. This can ensure we end up with the same set of GL calls. This is more robust then the image-verification approach, but it also is a much higher maintenance cost - any internal refactoring performance improvements that would've passed the image verification test would also flag as a failure for these tests. So I'd lean towards the image verification test, for now.

OhadRau:
[...]

PR #445, opened 2019-04-08
Refactoring: Remove state management API

bryphe:
Revery included a very simple Redux-inspired state-management solution. However, there's no reason to force users into a certain state-management paradigm (we should let the user bring their own state management solution.

For Onivim 2, I'm experimenting with [isolinear](https://github.com/bryphe/isolinear), which is an Elm-inspired approach that also handles effects.


msvbg:
This commit breaks my code (per `git bisect`). The window is completely black until I move the window, at which point rendering resumes.

bryphe:
Ah, sorry about that @msvbg ! 

I suspect it is due to the removal of the `needsRender` tracking variable - it's set to `true` for the first render, so it guarantees there will always be an initial render. I tested against this change against the example app and Onivim 2, but I suspect it worked in the former due to an animation (marking the window as dirty) and in the latter due to the separate state management we do that would trigger a render. I was able to reproduce with the example by changing the default example to a non-animatable one _and_ removing the `Window.setPos` call (any change in position / size will mark the window as dirty).

We can add the flag back to ensure there is a first render. I'll put up a shortly. Thanks for reporting!

bryphe:
This would be a perfect case to have a tool like #156 to avoid regressions like this.

msvbg:
Indeed, considering how visual Revery is it would make a lot of sense to have that.

PR #456, opened 2019-04-13
Bugfix - Regression: PR #445 caused a blank screen on startup

bryphe:
__Issue:__ For apps that don't have an animation or initial event to trigger a dirty window, the screen will be blank until something triggers a dirty state.

__Defect:__ Previously, before #445 , we had a `needsRender` value on the app instance that was set to `true` on startup, and anytime an action was dispatched against our ad-hoc state management. When that was removed, if nothing triggered a dirty event, a render would never be triggered.

__Fix:__ Always render the first frame. Bring back a limited version of  `needsRender` in the form of `isFirstRender`

### DIFF
--- a/src/Core/App.re
+++ b/src/Core/App.re
@@ -9,6 +9,7 @@ type t('s, 'a) = {
   mutable windows: list(Window.t),
   mutable idleCount: int,
   onIdle: idleFunc,
+  isFirstRender: ref(bool),
 };
 
 let framesToIdle = 10;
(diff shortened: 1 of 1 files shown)


bryphe:
@msvbg , does this fix the issue you hit with #445 ?

msvbg:
Yes, this appears to fix the issue :)

bryphe:
Thanks for the confirmation @msvbg ! 👍

PR #461, opened 2019-04-18
Font improvements

msvbg:
Cluster rendering now appears to be working properly. 8870f65 makes kerning a tiny bit off with certain font sizes on high DPI screens, but since Revery will be moving to Skia I'm thinking it's probably not worth spending more time working out details of font rendering.

### DIFF
--- a/src/Draw/FontRenderer.re
+++ b/src/Draw/FontRenderer.re
@@ -39,6 +39,31 @@ let _getTexture = ((font: Fontkit.fk_face, glyphId: int)) => {
   texture;
 };
 
+let filterGlyphClusters = shapedText => {
+  let lastCluster = ref(-1);
+  let increasing = ref(true);
+  Array.to_list(shapedText)
+  |> List.filter(s
+       // When the cluster values stop increasing monotonically, we have reached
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for the PR, @msvbg , and working through the cluster issues!

> 8870f65 makes kerning a tiny bit off with certain font sizes on high DPI screens, but since Revery will be moving to Skia I'm thinking it's probably not worth spending more time working out details of font rendering.

Agree 100% 👍 

Unfortunately there is a regression in the JS build with this change:
image

(You can run the JS build via `esy '@js' install` and `esy '@js' run`. The problem is there is no cluster information for the web implementation, so it only renders the first character. I think we should ignore the cluster and always render characters in that scenario (for now). We can check that via `Environment.webGL`

msvbg:
Oh no! I'd be happy to sort this out, or we can just close this PR since the Skia work is progressing so fast. I'm guessing this PR will be obsolete in a month or so anyway!

bryphe:
Sorry I missed the comment @msvbg ! Sounds reasonable - seems like Skia will handle this for us shortly 👍

Issue #322, opened 2019-02-09
Native Integration: Tray API


bryphe:
It would be great to have a cross-platform API for integrating with the system 'tray' (the system's notification area). Electron has shown a potential API for this use-case, along with events and methods that have thought-through applicability to all platforms.

This API surface would be a good starting point:
https://electronjs.org/docs/api/tray

The API surface is fairly large, so we should start by thinking of a minimal subset - perhaps addressing the following scenarios:
- Create a tray icon
- Destroy a tray icon

We'd also likely need a context menu for OSX.

lessp:
This may or not be a good starting point for implementing this on Windows:
<https://www.codeproject.com/Articles/18783/Example-of-a-SysTray-App-in-Win32>

PR #463, opened 2019-04-20
add Feature/menu for Revery_Native

Et7f3:
add Menu module to Revery:
image

I don't know why but I can't call OCaml functions as callback
https://github.com/Et7f3/revery/blob/feature/menu/src/Native/menu_win32.c#L52
Maybe I don't use the right C function.
I have added printf to debug.
I'm new to interfacing C and OCaml.


src/Native/Menu.re
```diff
@@ -0,0 +1,41 @@
+open Reglfw.Glfw;
+
+[@noalloc] external menuSupported: unit => bool = "revery_menuSupported";
+
+type menu;
+
+external create_menu: unit => menu = "revery_create_menu";
```
Akin909: @Et7f3 just a heads up, somewhat confusingly the convention in `reason` is `camelCase` whereas in `ocaml` its `snake_case` so I think these functions should be `createMenu` etc.

Et7f3: Sorry about that. I came from OCaml world. Should the native function be called revery_createMenu ? ok snkae case is ok because it's C ? In my school we use snake_case.

Et7f3:
I have seen that my dispatch cause the crash it seem that I don't use well reference 🤔 🤷‍♂️ . I have used reason try to obtain this function bug maybe it is not the reason paradigm.
I have added lot of printf and it get printed until I access assocCallback^ and try to do pattern-matching.
I have not tryed with switch syntax but I have no hope.

@Akin909 I have changed the naming convention

src/Native/Menu.re
Akin909: yeah :+1: the `C` code should be whatever is idiomatic in this case `snake_case`

Et7f3:
I try to implement https://github.com/revery-ui/revery/issues/250 and https://github.com/revery-ui/revery/issues/322 (because it is the same system under windows)

my current progress:
- [x] new Menu()
- [ ] `Menu.setApplicationMenu(menu)` (partially)
- [ ] `Menu.getApplicationMenu()`
- [x] `Menu.buildFromTemplate(template)`
- [ ] `menu.popup(options)`
- [ ] `menu.closePopup([browserWindow])`
- [ ] `menu.append(menuItem)`
- [ ] `menu.getMenuItemById(id)`
- [ ] `menu.insert(pos, menuItem)`
- [ ] `Event: 'menu-will-show'`
- [ ] `Event: 'menu-will-close'`
- [ ] `menu.items`
- [ ] `Event: 'menu-created'`

Et7f3:
`getApplicationMenu` return application menu but when multi window will be avaible we could have `getWindowMenu` also on windows
we don't have `setApplicationMenu` but we have `assignMenu` that assign menu to a window

Et7f3:
If someone want to explore tray here is a hint https://docs.microsoft.com/fr-fr/windows/desktop/api/shellapi/nf-shellapi-shell_notifyicona. Don't know if I should open a new pull request.

bryphe:
Thanks for the note about the tray, @Et7f3 ! We actually have another issue tracking that (https://github.com/revery-ui/revery/issues/322) - would be fine to have as a separate PR 👍

src/Native/menu.cpp
```diff
@@ -0,0 +1,369 @@
...
+#elif __linux__
+    printf("WARNING - Not implemented: revery_create_menu_gtk");
+#else
+    printf("WARNING - Not implemented: revery_create_menu");
+#endif
+
+    CAMLreturn(ret);
+}
+
+CAMLprim value revery_create_sub_menu()
```
bryphe: Thanks for creating our interface for this!

src/Native/Menu.re
```diff
@@ -0,0 +1,216 @@
...
+  "revery_add_sub_menu_popup_menu"; // should be the same on all OS
+
+let assocCallback = ref([]: list(unit => unit));
+/* TODO: make it private */
+/* TODO: use HashTbl to save memory and don't store all callback when not needed */
+
+let menuList = ref([]: list(menu));
+
+let menuDispatch = i => {
+  Printf.printf("we will dispatch: %d\n", i);
```
bryphe: We should remove these prior to bringing in

src/Native/menu_win32.c
```diff
@@ -0,0 +1,212 @@
...
+    UnhookWindowsHookEx(g_hook_handle);
+}
+
+value revery_add_string_item_menu_win32(value vMenu, int uid, const char * pMessage)
+{
+    bool ret = AppendMenu(Menu_val(vMenu).menu_handle, MF_STRING, uid, pMessage);
+
+    if (!g_hook_handle)
+    {
+        g_hook_handle = SetWindowsHookExW(WH_GETMESSAGE, WndProc, (HINSTANCE)NULL, GetCurrentThreadId());
```
bryphe: Thanks for figuring this part out 👍 nice find!

src/Native/menu_win32.c
```diff
@@ -0,0 +1,212 @@
...
+            printf("%d is clicked\n", LOWORD(m.wParam));
+
+            /* First time around, look up by name */
+            if (g_popup_dispatch == NULL)
+                g_popup_dispatch = caml_named_value("popup_dispatch");
+            caml_callback(*g_popup_dispatch, Val_int(LOWORD(m.wParam)));
+        }
+    }
+
+    CAMLreturnT(LRESULT, CallNextHookEx((void *) NULL, msg, wParam, lParam));
```
bryphe: Didn't know about `CAMLreturnT` - cool!

bryphe:
Thanks for all the work on this, @Et7f3 ! So cool to have that menu in our example app 😍 

It looks like there are a few things we need:

- [ ] Latest `master` needs to be merged (for the lock files - it might be easiest to just delete them and re-generate them - let me know if you need help)
- [ ] Remove debug logs (The `printfs` like 'We have created menu')

Overall looks good to me though - great foundation for us to continue building on 👍 Some opportunities for additional PRs down the road based on our discussions:

- JavaScript implementation for the API
- Other related APIs (like the tray API or right-click menus)
- Supporting other platforms (Cocoa, GTK, etc)

Very cool 😎

Et7f3:
👀 `right-click menus` is already implemented see here https://github.com/revery-ui/revery/blob/4bfc981ecbddbe7267370a47efeb5d3406c833d1/examples/Examples.re#L236-L237 (Or I haven't understand)

thanks for `glfwGetCursorPos` tips: it is used here https://github.com/revery-ui/revery/pull/463/commits/77ba0c8eb4cde5334ab4caddb6ed54c41ed4715c

Very-last details I have to do:


- [ ] keep window handle in the menuInfo and remove it when the menu is removed of the window
- [ ] transform append so it redraw menu on the window if needed
- [ ] insert
- [ ] remove
- [ ] produce rei file
- [ ] implement all missing js primitives no-ops
- [ ] add comment to explain what to change when multiple windows will be supported
- [ ] erase printf
- [ ] merge lock file (should be last thing)

This should be done by monday evening if I have time.

checkbox subitem for next PR
and we can change callback by using Hashtbl and manipulating callback value of each menu: next PR to protect it (so we can't erase useful callback)

src/Native/Menu.re
Et7f3: fix in 3c95b4b.

Et7f3:
I tried to merge and it seem I should better start from scratch. I was not aware of brisk-reconciler: So I have a lot of code to emulate that + the new ppx want `Brisk_reconciler.element` and non-generic `'a` like the old one + incomplete Sdl2 binding: They are no more `glfwGetCursorPos` That was used in popup-menu.

PR #469, opened 2019-04-23
Make clickable react only to left button

### LINKED ISSUE #468: Button should react only to left click
- Operating System: Windows 10
- Revery Version: last master
- OCaml version: not important
- Native, Bytecode, or JS build: not important

- Steps to reproduce: Open Button example and right click it

- Actual Result: Click triggered
- Expected Result: No click triggered

- Search terms used: Button, left click, right click

saitonakamura:
Are we ok with the fact that clickable should react to left button only?

Fixes #468

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -12,7 +12,7 @@ open Revery_Math;
 type clickFunction = unit => unit;
 type clickFunctionWithEvt = NodeEvents.mouseButtonEventParams => unit;
 let noop = () => ();
-let noopEvt = evt => ();
+let noopEvt = _evt => ();
 
 let isMouseInsideRef = (ref: node, mouseX: float, mouseY: float) => {
   let clickableDimensions: BoundingBox2d.t = ref#getBoundingBox();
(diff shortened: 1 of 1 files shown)


Et7f3:
I think we should make this a option because you disable the right click for contextual menu for instance. @saitonakamura

saitonakamura:
@Et7f3 I would suggest special callback for the right click, like `onRightClick`, and maybe `onWhateverClick` callback as a safehatch for more advanced scenarios (e.g. middle click). It would be more declarative rather than `reactToOtherThanLeftClicks` of something like that.

saitonakamura:
I've added support for `onRightClick` and `onAnyClick`. Right now there's a buggy behavior when you press multiple buttons simultaneously, but I'm not sure it worth fixing. Browsers tend to be strange for this cases too. Maybe there's a reason or a spec for it, but not that I know of

saitonakamura:
I don't genuinely know why build is failing on Macos

bryphe:
> I don't genuinely know why build is failing on Macos

Looks like a transient build failure... I kicked it off again 🤞

saitonakamura:
Seems like it's successful this time, let's merge?

bryphe:
Sounds good! Thanks @saitonakamura for the contribution!

PR #475, opened 2019-04-27
Enable DropDown and RadioButtons components to have generic value.

Et7f3:
In a school project, I have a function that translate string to Variant because DropDown only accept string as value.
But value are not dispayed => we could put anything.
So I have made this small patch to allow everybody to put whatever their want in value field.


bryphe:
Hi @Et7f3 ,

Sorry for the slow reply from me! This change looks great - thank you for the PR! 💯

PR #481, opened 2019-05-05
make checkbox controlled

### LINKED ISSUE #384: Checkbox inner state doesn’t update when outer state changed
When playing around examples I found strange behavior in Todo example.
1. Add 2 todos
2. Mark one done
3. Switch to only done todos
4. Switch to undone todos 

Result: undone todo marked as done, internal checkbox state doesn’t reflect external change

tatchi:
This PR should fix #384 by making the `Checkbox` component fully controlled.

The old `uncontrolled` behaviour is not supported anymore since I don't really see any use case where we would like to have a `Checkbox` that manages its own internal state that couldn't be accessed outside.

Furthermore, I think that managing these both scenarios add a non-negligible complexity for no real benefits.

Let me know what you think, but we can probably start with that and still make it more flexible if we really need it in the future?


examples/TodoExample.re
```diff
@@ -141,7 +142,8 @@ module Example = {
         <View style=Style.[flexDirection(`Row)]>
           <Checkbox
             checked={task.isDone}
-            onChange={checked => dispatch(ChangeTaskState(task.id, checked))}
+            onChange={() => dispatch(ToggleTaskState(task.id))}
+            // onChange={checked => dispatch(ChangeTaskState(task.id, checked))}
```
Akin909: can we remove this comment if were no longer using this

tatchi: Oops sorry I forgot to remove it 😅

tatchi:
Thanks for the review @Akin909, I did remove the comment :)

alexmironof:
awesome! works like a charm :+1:

PR #490, opened 2019-05-11
[WIP] Additional fixes to support Playground

bryphe:


bryphe:
I don't believe this is actually required now with the incremental compilation in the revery playground: https://github.com/revery-ui/revery-playground/pull/35

PR #503, opened 2019-05-18
API: Tick hook

wokalski:
This PR replaces the `Ticker` component with a `tick` hook. `Ticker` was used as a side-effect-only component which doesn't compose well with other effects (which are normally performed via hooks.)

In https://github.com/revery-ui/revery/commit/07e86599bcece57535dd7828c7568375173c32c8 I introduced some `open`s and module aliases to make rei files in hooks more readable.


bryphe:
Thanks @wokalski !

PR #513, opened 2019-06-19
[WIP] Bug: Typing quickly in textinput gives incorrect output

bryphe:
__Issue:__ The cursor position is not reliable when typing quickly - it can lag behind and cause text to be inserted in the incorrect position. 

Example:
typing-fast-breaks

__Defect:__ Our cursor-positioning logic is based on a render hook, which is dependent on re-renders and framerate. 

__Fix:__ We should adjust the cursor position at the same time we adjust the text.


bryphe:
@glennsl fixed this in a much better way in #644

PR #515, opened 2019-06-20
Fix #501

### LINKED ISSUE #501: Events are captured by clipped elements
- Operating System: macOS High Sierra 10.13.6
- Revery Version: 0.22.0
- OCaml version: 4.7.0

Events seem to be dispatched for elements that are clipped by a container having `overflow` set.

- Steps to reproduce:

For an easy repro, go to the `ScrollView` example, scroll down a bit and try to click on the checkbox, you will see the checkbox state is not toggled. In the gif below, I wrapped the `Image` element in the `ScrollView` example with a `Clickable` to see when it was being clicked:

```reason
<ScrollView style=outerBox bounce>
  <Clickable onClick={_ => print_endline("Clicked")}>
    <Image
      src="outrun-logo.png"
      style=Style.[width(512), height(256)]
    />
  </Clickable>
  ....
</ScrollView
```

clipped-events

- Actual Result: The events are dispatched to the clipped image.

- Expected Result: Clipped elements should not receive events. The click event should be dispatched to to the toggle element in the repro above.

- Search terms used: events, clip, overflow

SirFurness:
Fixes #501 

It introduces a clipped bounding box for nodes formed by the intersection of the node's bounding box and its parent's clipped bounding box. It then uses this clipped bounding box for the hitTest so that clipped parts of elements will not be interacted with.

### DIFF
--- a/src/Math/BoundingBox2d.re
+++ b/src/Math/BoundingBox2d.re
@@ -40,8 +40,8 @@ let intersects = (b0: t, b1: t) => {
   !noOverlap;
 };
 
-let intersect = (b0: t, b1: t) => {
-  if(intersects(b0, b1)) {
+let intersect = (b0: t, b1: t) =>
+  if (intersects(b0, b1)) {
     let box0_minX = Vec2.get_x(b0.min);
     let box1_minX = Vec2.get_x(b1.min);
     let minX = max(box0_minX, box1_minX);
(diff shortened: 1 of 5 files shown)


src/Math/BoundingBox2d.rei
```diff
@@ -9,6 +9,8 @@ let create: (Vec2.t, Vec2.t) => t;
 
 let intersects: (t, t) => bool;
 
+let intersect: (t, t) => t;
```
bryphe: Thanks for adding this method, @SirFurness !

src/Math/BoundingBox2d.re
```diff
@@ -40,6 +40,34 @@ let intersects = (b0: t, b1: t) => {
   !noOverlap;
 };
 
+let intersect = (b0: t, b1: t) =>
```
bryphe: Could you please add a simple test case here? https://github.com/revery-ui/revery/blob/master/test/Math/BoundingBox2dTests.re

SirFurness:
Glad to hear the fix looks good! I added in some basic tests for the function, @bryphe.

bryphe:
Awesome! Looks great @SirFurness . Thank you for adding the tests!

Bringing this in now 👍

PR #516, opened 2019-06-28
Fix #514

### LINKED ISSUE #514: Scrollbar not updating correctly
- Operating System: macOS Mojave 10.14.5
- Revery Version: 0.23.0
- OCaml version: 4.7.0

- Link to github repo: https://github.com/SirFurness/ScrollbarIssue

- Steps to reproduce: Install and run the repo (the readme explains a bit more)

- Actual Result: Currently, when children are added to a ScrollView, the scrollbar does not change. Then when one does scroll, the scrollbar will teleport to its new location and change size.
- Expected Result: When children are added to a ScrollView, the size and location of the scrollbar should change immediately.

- Additional Information: Besides being jarring, this can lead to certain difficulties. In Oni2, if one scrolls to the bottom of the file explorer menu thing and opens a folder, one cannot scroll down to view the contents because the scrollbar has not yet updated its position. So one needs to scroll up and then scroll down.

- Search terms used: ScrollView, scrollbar, updating

r281GQ:
force rerender in scrollview on children update (which is everytime I suppose if brisk treats props immutable)

I was tinkering with onivim2 and noticed this #514 .

That is my first time with brisk, so it is just a though experiment. The container view updates perfectly since its height is managed by the reconciliation. As I understand the code, we acquire the ref, to measure the the height and than pass it to the vertical sidebar. This is where things get out of sync.

Since the container view is not re-mounted, the callback with the `ref `and then setState does not fire.

When we scroll however, we trigger a re-render and the height is recalculated.

The thing I don't get is the `ref` should be just a memory reference to the container object. So we should have the most up to date value on the ref all the time. But that is not the case. This dirty approach solves the problem, by forcing a re-render, but looking at the code it should take care of itself. 

What do you guys think?

### DIFF
--- a/src/UI_Components/ScrollView.re
+++ b/src/UI_Components/ScrollView.re
@@ -38,6 +38,8 @@ let make =
       children: React.syntheticElement,
     ) =>
   component(slots => {
+    let (dummy, setDummy, slots) = Hooks.state(0, slots);
+
     let (actualScrollTop, dispatch, slots) =
       Hooks.reducer(~initialState=scrollTop, reducer, slots);
     let (outerRef: option(Revery_UI.node), setOuterRef, slots) =
@@ -52,6 +54,17 @@ let make =
       TranslateX((-1.) *. float_of_int(actualScrollLeft)),
       TranslateY((-1.) *. float_of_int(actualScrollTop)),
     ];
+
+    let slots =
+      Hooks.effect(
+        Revery_UI.React.Hooks.Effect.If((!==), children),
+        () => {
+          setDummy(dummy + 1);
+
+          None;
+        },
+        slots,
+      );
 
     let (horizontalScrollBar, verticalScrollBar, scroll) =
       switch (outerRef) {


bryphe:
Thanks for the PR & notes, @r281GQ ! Great investigation.

> The thing I don't get is the ref should be just a memory reference to the container object. So we should have the most up to date value on the ref all the time. 

I suspect what is happening here is the following:

- We render the `<ScrollView />` with new `children`
- Layout occurs after the render
- Measurement for the node gets updated (ie `outer#measurements();`  would now return something new)

...but without a trigger of some kind - we don't know we need to re-render (in the case where our dimensions changed) - and so we don't actually pick up that new value.

The `setState` you do here queues up a change, so that we get an additional render that picks up the new value returned by `outer#measurements`.

(There's also a potential we're hitting the stale-setter issue in here: https://github.com/briskml/brisk-reconciler/issues/26 if somehow the `ref` is being called later)

I think there could potentially be a nicer way to express this, but we're missing it in Revery today - if we had a way to compute that had a resize event, like:
```
<SizeObserver onSizeChanged={(width, height) => ...set new outer width / height...}>
...
</SizeObserver>
```

Having this could potentially simplify by removing the need for a ref/effect/state to trigger the re-render - but your approach is clever to work around this limitation. 

The proposed approach seems reasonable to me, and as a next step we could look at something like the `SizeObserver` to streamline this.

PR #517, opened 2019-07-03
Clickable: Expose onRightClick and onAnyClick

thangngoc89:
I took a stab  at https://github.com/onivim/oni2/issues/322 but found out that it's blocked because onAnyClick isn't exposed anywhere. This PR aims to fix that


bryphe:
Ah, nice catch @thangngoc89 ! Thanks for adding this 👍

PR #521, opened 2019-07-12
[WIP] Experiment: Custom title bar on OSX

bryphe:
This is an experiment in creating a frameless window but with the default OSX titlebar buttons - inspired by Electron (and actually, the code is ported from Electron's implementation)


bryphe:
I'll close this out! This is out-of-date, as it was against GLFW - and I think @zbaylin 's work in https://github.com/revery-ui/reason-sdl2/pull/26 and APis will Revery will address this scenario for us now 🎉

PR #529, opened 2019-07-30
Performance: Improve idle performance

bryphe:
__Issue:__ Today, when a Revery app is 'idle' (ie, no user input or animations incoming), CPU usage is higher than it should be.

__Defect:__ Revery is constantly running the application loop and re-rendering (like a game). Howevever, for applications, we don't want to be running the application loop this frequently. When there is no user input, and no animations or processing, we should block on user input.

__Fix:__ When idling, use `glfwWaitInputEvent` to block until user input. Add a method on `App` to force a re-render (via `glfwPostEmptyEvent`).

Needed for https://github.com/onivim/oni2/issues/304

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@esy-ocaml/reason": "3.4.0",
-    "reason-glfw": "^3.2.1024",
+    "reason-glfw": "^3.2.1025",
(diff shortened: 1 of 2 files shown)


bryphe:
Implemented in https://github.com/revery-ui/revery/pull/606

PR #555, opened 2019-09-18
GLFW -> SDL2 migration

bryphe:
We're moving from GLFW -> SDL2 for a few reasons:
- Better IME support across platforms in SDL2
- Better keyboard layout / composition support
- Better scaling / high-DPI support

In addition, there are several APIs we'd be interesting in picking up down the road... For example - the Audio API will be useful, and the native messagebox APIs could replace our native alert dialogs.


package.json
```diff
@@ -39,15 +38,18 @@
...
   },
   "resolutions": {
     "rebez": "github:jchavarri/rebez#46cbc183",
     "@opam/cmdliner": "1.0.2",
     "@opam/js_of_ocaml": "github:ocsigen/js_of_ocaml:js_of_ocaml.opam#db257ce",
     "@opam/js_of_ocaml-compiler": "github:ocsigen/js_of_ocaml:js_of_ocaml-compiler.opam#db257ce",
     "@brisk/brisk-reconciler": "github:briskml/brisk-reconciler#dd933fc",
-    "pesy": "0.4.1"
+    "pesy": "0.4.1",
+    "reason-fontkit": "link:../reason-fontkit",
```
Et7f3: Maybe upload thesse two repo so the ci can download them.

examples/Examples.re
```diff
@@ -293,7 +295,7 @@ let init = app => {
   } else {
     let xPosition = (dimensions.width - windowWidth) / 2;
     let yPosition = (dimensions.height - windowHeight) / 2;
-    Window.setPos(win, xPosition, yPosition);
+    Window.setPosition(win, xPosition, yPosition);
   };
 
   let _ = UI.start(win, <ExampleHost win />);
```
Et7f3: Don't forget to rename to let _ignore so it compile to 4.8.1

package.json
bryphe: Done 👍

bryphe: Need to upgrade `reason-sdl2`, there was a dependency bug I fixed - https://github.com/revery-ui/reason-sdl2/pulls/3

I'll publish a new package w/ that and update the lockfiles

examples/Examples.re
bryphe: Ah, thanks for the reminder @Et7f3 ! I just merged in your change.

szbergeron:
Just a note, revery seems to pull from the system sdl rather than esy-sdl as well, so it may be necessary to add something similar to the patch for reason-sdl that set the link path

szbergeron:
@bryphe the src/Core/dune file seems to have artifacts from a manual merge, might cause issues during build

bryphe:
Thanks @szbergeron for the review!

> Just a note, revery seems to pull from the system sdl rather than esy-sdl as well, so it may be necessary to add something similar to the patch for reason-sdl that set the link path

I'm upgrading reason-sdl2 to pick up the latest `esy-sdl2`, which has that fix: https://github.com/revery-ui/reason-sdl2/pull/7

> @bryphe the src/Core/dune file seems to have artifacts from a manual merge, might cause issues during build

Should be fixed in the latest -

package.json
bryphe: Should be fixed now!

PR #556, opened 2019-09-19
Make revery buildable with OCaml 4.8.1

Et7f3:
We have warning 5 because we are ignoring the error
`let _ = `

I changed to
`let _ignore = `
So we ignore but without warning so it build.
I have also added a TODO.


Et7f3:
I have added a resolution to rench because it fail without it.

PR #566, opened 2019-09-28
[Experiment] brisk-reconciler component PPX

bryphe:


ducaale:
I have seen PPX for hooks example in https://github.com/briskml/brisk-reconciler/pull/42#issuecomment-544195151. can that also be used here?

```
let%component make = (~prop) => {
  let%hook (state, setState) = Hooks.state("");
  <Text value=state />;
};
```

bryphe:
Closing as #576 handles this, and much more

PR #567, opened 2019-09-28
Skia integration

### LINKED ISSUE #353: UI Styles: Border Radius

bryphe:
This is just linking in `reason-skia` and starting to hook up the Skia canvas to Revery's pipeline - still a lot left to do for the full integration!

This is dependent on a PR in `reason-skia`: https://github.com/manuhornung/reason-skia/pull/6

__Latest screenshot:__
latest-skia

__Remaining work:__
- [x] Load images via `SKData`
- [x] Handle transforms - stub out `sk_matrix44_t`, and copy our transform matrices over to use for the canvas
- [x] `Overflow` - bring in and use `clipRect`
- [x] Text rendering: Calculate baseline correctly
- [x] OSX: Fix crash with typeface
- [x] Move OpenGL -> Canvas, port example
- [x] Use image width / height
- [x] Validate `zoom` works as expected
- [x] Validate high-DPI rendering (ie, retina display) is correct
- [x] Fix warnings
- [x] Bring back border rendering using the Skia canvas primitives
- [x] Bring back box shadow rendering using the Skia canvas primitives
- [x] Move from `Rectangle.t` to just the `Skia.Rect.t` primitive
- [x] Move from `Reglm.Mat4.t` to just the `Skia.Matrix44.t` primitive
- [x] Use `Skia.Rect.t` as underlying primitive for `Skia.BoundingBox2d.t`
- [x] Hook up rotation transforms again
- [x] Fix rendering of FontAwesome glyphs
- [x] Text rendering: Use harfbuzz to shape and pass glyph ids, instead of sending a string
- [x] Use Skia primitives for font measurement, instead of `Fontkit`
- [x] Test and validate with Onivim 2

Closes #353


manuhornung:
This is looking awesome! I'll try to find some time to bring all the PRs in during the next few days so we can properly hook everything up

PR #570, opened 2019-10-01
Bugfix: Change font rendering to use GL_NEAREST

kitten:
**This depends on: https://github.com/revery-ui/reason-sdl2/pull/9**

In `Draw/Text.re` there's already some code that tells FreeType2 to render given glyphs with a font-size that takes the current pixel ratio into account. Hence FreeType2 is always used to render text at the exact size that it needs to be displayed at.

Yet in `Draw/FontRenderer` we use `GL_LINEAR` (& also w/o mipmap filtering) which enables bilinear filtering and causes some artifacts to be rendered. This is especially noticeable on 2xDPI screens since the pixel density is just low enough for the text to look a little "smudgy".

By using `GL_NEAREST` we prevent this filtering from happening altogether and render the font texture as-is (https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glTexParameter.xml)

>  Returns the value of the texture element that is nearest (in Manhattan distance) to the center of the pixel being textured.

I've attached a screenshot that should demonstrate what rendering will look like after applying these changes. Not sure if the texture filtering is enabled on purpose though 😅 

Screenshot 2019-10-01 at 15 32 35

This is at font-size 8px on a Macbook Pro 15" (2x pixel ratio). The problem is especially noticeable on the `a`, how the nearest-filtering smudges on some tight curves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,8 @@ Thank you to all our backers! 🙏 [[Become a backer](https://opencollective.com
 - [flex](https://github.com/jordwalke/flex) by @jordwalke
 - [briskml](https://github.com/briskml)
     - [brisk-reconciler](https://github.com/briskml/brisk-reconciler) - the "native React" implementation.
-- [reason-glfw](https://github.com/bryphe/reason-glfw)
-    - [GLFW](https://www.glfw.org)
+- [reason-sdl2](https://github.com/revery-ui/reason-sdl2)
+    - [SDL2](https://www.libsdl.org)
     - [stb-image](https://github.com/nothings/stb)
 - [reason-fontkit](https://github.com/bryphe/reason-fontkit)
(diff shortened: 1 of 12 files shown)


bryphe:
Wow, great catch @kitten ! Appreciate the PR and detailed notes / investigation.

Looks like we don't have the `GL_NEAREST` interpolation flag exposed right now - I'll bring that in and update the deps, so we can then pick up this PR.

kitten:
@bryphe Awesome! The tests on that PR are currently failing, but I'm not quite sure why https://github.com/revery-ui/reason-sdl2/pull/9/checks

bryphe:
Build is green now! 

Thanks for the contribution and the help updating `reason-sdl2`, @kitten 🎉 🎉 Awesome catch!

Issue #489, opened 2019-05-11
Proposal: API - Move from 'style' to Compositional Components


bryphe:
This came from some discussions with @manuhornung and others around Flutter's model vs our `style` model. IMO, one thing Flutter did really well in their design is their compositional component model (moving away from the uber-`style` grab bag of properties).

To clarify, what I mean by a compositional model, is instead of having a `Style.[...]` with arbitrary properties (like positioning, transform, etc), Flutter has more scoped widgets - like a `Transform` widget, or `Positioned` widget. This is nice in terms of discoverability - each of those widgets takes a relatively small and concise set of properties.

In addition, the `Style.[..]` model is coupled to a single layout system, whereas Flutter uses component composition to express a variety of layout models - this is very convenient for building UI apps.

The `Style.[..]` model has gotten us really far, but there are a few downsides I've seen so far:
- __Discoverability:__ The use of Polymorphic Variants isn't as well supported with auto-complete (as far as I can tell), whereas specific properties would be more discoverable
- __Discoverability:__ It's challenging to know all the potential properties, and how they interact, when using `Style.[..]`.
- __Performance:__ We always have to `fold` our style-list into a style-object: https://github.com/revery-ui/revery/blob/0eb56f437da69b58033bc6156c35b1db145f172f/src/UI/Style.re#L501 - I wonder if there could be a cheaper way to handle this? Merging is always challenging: https://github.com/revery-ui/revery/blob/0eb56f437da69b58033bc6156c35b1db145f172f/src/UI/Style.re#L589
- __Performance:__ Our `Node` class ends up having to do a lot of different things - recompute transforms, even if there is are no extra transforms, check the overflow property, etc.

I think the compositional model could help with some of the __discoverability__, by having a _more constrained set of properties per-component_, and help with the __performance__ by giving each component a more _constrained set of responsibilities_ (potentially minimizing extra transform calculations, overflow calculations, etc).

If people are on-board with moving to the compositional model vs the uber-style model, I'd propose tackling it incrementally, by removing style properties and replacing them with compositional components:

- [ ] The `transform` property could become the responsibility of `Transform.Rotate`/`Transform.Scale`, etc components (#488)
- [ ] The `position` and `top`/`left`/`right`/`bottom` could become the responsibility of `Stack` and `Positioned` components (#483)
- [ ] The `backgroundColor` property could become the `color` property of a `Container`
- [ ] The `color` property could become the `color` property of a `Text`
[...]

PR #576, opened 2019-10-01
refactor!: update to latest brisk-reconciler

lessp:
~~So, pushing what I've got, I began converting the examples. Most of the should be OK now hopefully. Got a bit stuck on the `UI_Primitives` type-errors.~~

This PR introduces the changes from [brisk-reconciler](https://github.com/briskml/brisk-reconciler/commit/0a2e4765be13a47f377c836ab3eeee6cebb5d78e) with the new syntax:

e.g.

```ocaml
let%component make = () => {
  let%hook (counter, setCount) = React.Hooks.state(0);

  <Clickable onClick={_ => setCount(prevCount => prevCount + 1)}>
    <Text style=Style.[...] text={string_of_int(counter)} />
  </Clickable>
};
```


lessp:
Btw, this is a PR for merge into: https://github.com/revery-ui/revery/pull/566

examples/DefaultButton.re
```diff
@@ -5,7 +5,7 @@ open Revery.UI.Components;
 module DefaultButtonWithCounter = {
   let%component make = () => {
     let%hook (count, setCount) = Hooks.state(0);
-    let increment = () => setCount(count + 1);
+    let increment = () => setCount(_ => count + 1);
```
glennsl: You're supposed to do `setCount(count => count + 1);` here. That's why it was made a function ;) It's unlikely to cause a problem if you don't, especially in this case where it worked absolutely fine before, but when it does cause a problem it's a tricky one to find because it requires breaking the abstraction. It also might be very hard to reproduce. See briskml/brisk-reconciler#26 for details.

lessp:
Ok, so with a few files commented out:

`UI_Components`
- Button.rei
- Checkbox.rei
- Clickable.rei
- ClipContainer.rei
- ClipContainer.rei
- Container.rei
- Dropdown.re
- Positioned.rei

`UI_Hooks`
- Effect.rei_
- Reducer.rei_
- Ref.rei_
- State.rei_
- Tick.rei_

`examples`
- DropdownExample.re `noop`
- FontsExample.re `noop`
- RadioButtonExample.re `noop`

-------

This builds and runs, with the core changes happening in `UI/Reconciler.re` and `UI/Container.re`

examples/DefaultButton.re
lessp: Yep! I just mangled through all these and replaced with `_`!

Et7f3:
I don't why I have a empty windows when executing `esy run`

lessp:
Merged with master, some examples doesn't run:

- Fonts
- Input~
- RadioButton

Et7f3:
You should better rebase onto master instead of merging (you will avoid this kind of merge error).

lessp:
The `Input`-component doesn't seem to be able to get refocused. I did some minor debugging and backtracked it to `NodeEvents` I think, but didn't spend a lot of time on it!

lessp:
The issue with the `Input`-component is fixed, turns out it was a missing default-value for `tabindex` in `Clickable` that was lost when I converted that.

I have updated all the interface-files I commented out previously. I'm still experiencing this:

> EDIT: Also, when an application starts, I've sometimes got stuck in a screen where it's not rendered completely. Not sure about that one!

Would be great if someone else could confirm!

lessp:
Also, running locally I'm getting the following error when trying to run tests. I think the same is happening in CI. Not sure what it means:

```
❯ esy @test run
esy-build-package: source path
                   /Users/ter/dev/diverse/reason/revery/_esy/bench/store/b/revery-cf2a28d6/install/default/lib/ReveryTest/META
                   does not exist
  
esy: exiting due to errors above
```

js.json
```diff
@@ -1,29 +1,29 @@
...
+      "@opam/js_of_ocaml": "*",
+      "@opam/js_of_ocaml-compiler": "*",
+      "@opam/js_of_ocaml-lwt": "*",
+      "http-server": "*",
+      "@opam/merlin-extend": "0.4"
+    },
+    "resolutions": {
+      "rebez": "github:jchavarri/rebez#46cbc183",
+      "rench": "github:revery-ui/rench#4554280",
+      "@brisk/brisk-reconciler": "github:briskml/brisk-reconciler#a0fe3a9",
```
Et7f3: not latest commit.

src/UI/Container.re
```diff
@@ -1,21 +1,30 @@
...
-        let updates = React.RenderedElement.render(node, element);
+        let updates =
+          React.RenderedElement.render(
+            {
+              node,
+              insertNode: React.insertNode,
+              deleteNode: React.deleteNode,
+              moveNode: React.moveNode,
+            },
+            React.empty,
```
Et7f3: Where is element ? is it normal it is React.empty ?

src/UI_Components/Button.rei
```diff
@@ -11,10 +11,9 @@ Simple out-of-box button component
 */
 let make:
   (
-    ~children: 'a,
-    ~title: string,
+    ~title: String.t,
     ~onClick: unit => unit=?,
-    ~color: Revery_Core.Color.t=?,
+    ~color: Revery_Core__.Color.t=?,
     ~fontSize: int=?,
```
Et7f3: Why ? and String.t

src/UI_Components/Clickable.re
```diff
@@ -23,34 +23,31 @@ let isMouseInsideRef = (ref: node, mouseX: float, mouseY: float) => {
...
+let%component make =
+              (
+                ~style=[],
+                ~onClick: clickFunction=noop,
+                ~onRightClick: clickFunction=noop,
+                ~onAnyClick: clickFunctionWithEvt=noopEvt,
+                ~componentRef=?,
+                ~onBlur=?,
+                ~onFocus=?,
+                ~tabindex=0,
```
Et7f3: why 0? Many node will have zero as index...

src/UI_Components/ExpandContainer.re
```diff
@@ -2,10 +2,9 @@ open Revery_Core;
 open Revery_UI;
 open Revery_UI_Primitives;
 
-[@component]
-let make = (~children, ~color=Colors.transparentWhite, (), hooks) => {
+let make = (~children, ~color=Colors.transparentWhite, ()) => {
```
Et7f3: shouldn't be `let%component make` or it is stateless so no need for this extention ?

src/UI_Hooks/Animation.rei
```diff
@@ -1,13 +1,62 @@
...
+  let transition:
+    (
+      float,
+      ~delay: Revery_Core.Time.t=?,
+      ~duration: Revery_Core.Time.t=?,
+      Revery_UI.React.Hooks.t(
+        (
+          ref(Revery_UI.Animated.animation),
+          Revery_UI.React.Hooks.Reducer.t(int),
+          Revery_UI.React.Hooks.Effect.t(Revery_UI.React.Hooks.Effect.onMount)
```
Et7f3: Why so long name 😂 `Reducer.t(int)` instead of `Revery_UI.React.Hooks.Reducer.t(int)`

test/UI/ReconcilerTests.re
```diff
@@ -1,41 +1,39 @@
...
+open TestFramework;
+
 /*
  * This module tests the integration of the Brisk reconciler - `React` -
  * with our Node tree.
  */
 
 module TestRefComponent = {
-  let component = React.component("TestRefComponent");
+  [@component]
```
Et7f3: Old syntax.

src/UI/Container.re
lessp: Ahh, nice catch! This fixed the rendering issue where it would not render the content before you had given the window focus!

src/UI_Components/Button.rei
lessp: Perfect, thanks!

src/UI_Components/Clickable.re
lessp: Good question, I'm not sure, but my feeling was that this is sometimes needed for an element to be able to receive focus. 

I had forgotten it initially, when added back it allowed the `Input`-component to again receive focus. 

Anyway, the default-value here is not a part of this commit, so might be a different discussion!

<https://github.com/revery-ui/revery/blob/master/src/UI_Components/Clickable.re#L114>

src/UI_Components/ExpandContainer.re
lessp: Exactly, it's stateless, so can't use the extension here.

src/UI_Components/Clickable.re
Et7f3: I have applied diff from HEAD of master 4a37510cc8ca436e69fc8e77e9930126ca633b33 and this branch.

src/UI_Hooks/Animation.rei
Et7f3: Fixed in 9516740

js.json
lessp: Fixed in: https://github.com/revery-ui/revery/pull/576/commits/1db02312dfa80c49b01bffd907340ebf81f091b2

src/UI/Container.re
lessp: Fixed in: https://github.com/revery-ui/revery/pull/576/commits/4c68021239363ea3fe8da0e344a34ff9ff0bfa2d

src/UI_Components/Button.rei
lessp: Fixed in: https://github.com/revery-ui/revery/pull/576/commits/9516740a91cefdd4a55a0cbdfd7e33e221c39498

test/UI/ReconcilerTests.re
lessp: Fixed in: https://github.com/revery-ui/revery/pull/576/commits/245f46f8b651b227c43973700a5f4992ad55ba67

src/UI_Primitives/Image.re
```diff
@@ -17,25 +17,23 @@ let getStyles: (option(int), option(int), Style.t) => Style.t =
...
+                      ~onMouseUp=?,
+                      ~onMouseWheel=?,
+                      ~ref=?,
+                      ~resizeMode=ImageResizeMode.Stretch,
+                      ~opacity=1.0,
+                      ~width=?,
+                      ~height=?,
+                      ~src="",
+                      ~style=Style.emptyImageStyle,
+                      ~children=React.empty,
```
Et7f3: Image can have children 👀

Et7f3: Same for other Primitives like OpenGL

lessp: I don't think this changes that, we just need t provide an empty children as a default which is needed IIRC!

lessp:
Wondering if we should change the merge target to master instead of #566?

src/UI_Primitives/Image.re
Et7f3: Scrollview, Row, Container, Center must have one argument so default arg to `React.empty` is not fine for me. Ticker shouldn't have child.

Et7f3: I though not exposing argument and defining a local children variable. What do you think ? I can fix that.

lessp: That might make sense, but it differs from how master looks currently: <https://github.com/revery-ui/revery/blob/master/src/UI_Components/Input.re#L356>

lessp: I've updated all containers except `Container` as to explicitly take children @Et7f3. 
Do you want me to push these, or do you want to create a separate issue for that?

Et7f3: Ah I see so create another issue. IIRC @bryphe wanted some compile time assertion to be verified.

Et7f3:
When reviewing it I compared this branch and master and it was clean. I change it.

src/UI_Primitives/Image.re
Et7f3: for reference https://github.com/revery-ui/revery/issues/489#issuecomment-546668675_

bryphe:
So cool to see this green and the `let%hooks` land! 🎉 Thanks for all the work / help with this!

> Wonder if we should change the merge target to master instead of #566?

Sounds good to me! We can just close #566 in lieu of this more complete change 😄

wokalski:
The only problem is that let%hook doesn't work in reis currently. And it's hard to make it work there.

lessp:
> The only problem is that let%hook doesn't work in reis currently. And it's hard to make it work there.

Would you like to get this in before merging, or should we add it at a later stage? @wokalski @Et7f3

wokalski:
No, it’s very hard. There are lower hanging fruits with more ROI.

src/UI_Components/Positioned.re
```diff
@@ -1,22 +1,21 @@
 open Revery_UI;
 open Revery_UI_Primitives;
 
-let component = React.component("Positioned");
+module X = Revery_UI.Brisk_jsx_runtime;
```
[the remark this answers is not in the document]
lessp: I wouldn't think so! I removed them. I wonder if they were used in the initial refactor in https://github.com/revery-ui/revery/pull/566. Anyway, good catch!

Removed in: 56e6c33 and a47c9dc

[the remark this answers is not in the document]
wokalski: Brisk_jsx_runtime is unnecessary as of the defunctorized brisk-reconciler!

bryphe:
Awesome! Let's bring this in - very exciting! 🎉 Thanks @wokalski @lessp @Et7f3

PR #578, opened 2019-10-01
Add an animation controller

zbaylin:
This will allow control of the animation from the UI, which was not present before.

I've already added pause and restart functions. We will probably need

- `resume`
- `reverse`
- etc.


zbaylin:
Okay! I finally fixed it so that it will restart even if the animation has completed by adding a completion generator function. 

This will really help with my Material Design library, because now we can trigger animations, i.e. on hover for buttons.

src/UI/Animation.re
```diff
@@ -237,4 +228,81 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
+        // Set the animation's start time to the new time
+        animation.startTime = newStartTime;
+        // Reinsert the last active animation
+        activeAnimations := List.append([activeAnim], activeAnimations^);
+        ();
+      | None => ()
+      };
+    pub getAnimation = () => animation;
+    pub setCompleter = (c : unit => ((unit => unit))) => completer = Some(c);
+  };
```
glennsl: Using a class for this is not very idiomatic in Reason. I also don't think it provides anything in this case over using a plain object (which can be created independently of classes) because it doesn't use inheritance. Object would still not be idiomatic, however. The idiomatic approach would be to just use plain functions that's passed the state as an ordinary argument, e.g. `Animation.pause(animation)`. It seems like it would be better to implement these methods as plain functions alongside `start`.

glennsl:
In Oni2 I've created an [`AnimatedView`](https://github.com/onivim/oni2/pull/858/commits/841a54174c8574791da170e311bfe56239e6500e) component that provides a declarative API using a render function.  Note that this uses the `Animation` module in Oni2, not the one in revery, but they seem pretty similar. It also doesn't actually doing anything yet if any of the props changes, but it might be an idea worth pursuing here as well.

src/UI/Animation.re
zbaylin: I agree with you that a class is idiomatic in Reason -- it was the first time I had written a class in OCaml! But I used it for the following reasons:

- Primarily, persistent mutable state. The animation controller class stores not only an animation itself, but the last `activeAnimation`. The de facto solution for this has been to iterate over `activeAnimations` to find any that contain a matching `animation`, but since we need to keep access to this after the animation "expired", this solution wouldn't work.
- Classes are used throughout revery. Every single `node` is represented as a class, and there are other situations where classes are used too.

I would definitely like to remove the class if possible, so if you have any ideas on how to refactor this let me know!

glennsl: The `node` hierarchy uses classes because of inheritance and nominal subtyping. But since this doesn't use that, you can just use a plain object instead. All you need to do is change

    class animationController (_animation: animation) = ...

to

    let animationController = (_animation: animation) => ...

then you have an ordinary function returning an ordinary object, which works exactly like an object obtained from a class type except you don't need to instantiate it with `new`.

I'm still not convinced that you need an object though. Could you elaborate on what you mean by the animation "expiring" and being needed past that?

src/UI/Animation.re
```diff
@@ -237,4 +228,81 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
+          List.filter(
+            a => a.animation.id == _animation.id,
+            activeAnimations^,
+          ),
+          0,
+        )
+      ) {
+      | x => Some(x)
+      | exception (Failure(_)) => None
+      };
```
glennsl: This could be reduced to `List.find_opt(a => a.animation.id == _animation.id, activeAnimations^)'

src/UI/Animation.re
```diff
@@ -237,4 +228,81 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
+    pub updateLastActive = () => {
+      let l =
+        List.filter(a => a.animation.id == animation.id, activeAnimations^);
+      switch (List.nth(l, 0)) {
+      | x => lastActive = Some(x)
+      | exception (Failure(_)) => ()
+      };
```
glennsl: `List.find_opt` can be used here as well. In addition, `List.nth(l, 0)` is the same as `List.hd(l)`, but you should almost avoid using these in favor of pattern matching directly on the list:

```
switch (l) {
  | [x, ..._] => lastActive = Some(x)
  | _ => ()
}
```

src/UI/Animation.re
zbaylin: Animations are "ticked" in a separate type, `activeAnimation`. However, if the animation finishes (see https://github.com/revery-ui/revery/blob/86bf21617d97f72e4baec346c23ff9abbacc24c5/src/UI/Animation.re#L233-L234), the `activeAnimation` record is filtered out and garbage collected. 

The `animationController` caches the last active animation in `lastActive`, which is an `option(activeAnimation)`. The reason we can't just create new `activeAnimation`s is because they contain relevant functions, particularly how to update and complete the animation.

Thinking now, I guess we could create a memoized function that returns a cached `activeAnimation` (we already have `Memoized.make`), but I still think an object suits this better.

src/UI/Animation.re
```diff
@@ -237,4 +228,81 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
+        let complete = switch (completer) {
+          | Some(x) => Some(x())
+          | None => None
+        };
+        let newActiveAnim = {
+          animation,
+          update: activeAnim.update,
+          complete
+        }
+        activeAnimations := List.append([newActiveAnim], activeAnimations^);
```
glennsl: There is a special syntax for appending (or rather prepending, which is what this actually does) a single element to a list:

    [newActiveAnim, ...activeAnimation^]

does exactly the same as

    List.append([newActiveAnim], activeAnimations^)

 except it doesn't allocate a temporary list to hold `newActiveAnim`

zbaylin: I originally had that, but I noticed that this was here: https://github.com/revery-ui/revery/blob/86bf21617d97f72e4baec346c23ff9abbacc24c5/src/UI/Animation.re#L170

I can change both if that is considered better practice.

src/UI/Animation.re
zbaylin: fixed in latest commit

src/UI/Animation.re
zbaylin: fixed in latest commit

src/UI/Animation.re
glennsl: Yeah I think I see what you mean. `pause` could just return the active animation, or a `resume` closure that references it. `restart` could also be a closure returned from `start`, or it could require that `update` is passed in again.

Lots of ways to solve this, but I think I'd prefer using closures since it prevents you from resuming something that isn't paused, and from restarting something that hasn't been started. It models the problem better.

src/UI/Animation.re
glennsl: Yeah, this is exactly what `List.append` does under the hood anyway.

Lists in OCaml/Reason are essnetially implemented as

    type list('a) = Cons('a, list('a)) | Nil

with the empty list being `Nil`, a one element list being `Cons(1, Nil)`, two elements `Cons(1, Cons(2, Nil))` etc.

    [newActiveAnim, ...activeAnimations^]

is therefore essentially just syntax sugar for

    Cons(newActiveAnim, activeAnimations^)

src/UI/Animation.re
zbaylin: Using closures would mean you would need to have external state management, likely in the component, which is essentially just replacing where the state is managed (in the current case, the class/object). For instance, if you returned the active animation from pause, you would have to store that value somewhere in your component. This gets cumbersome if you have multiple animations (having to manage two things instead of one), and it makes the `activeAnimation` record visible to the user (as opposed to only being managed in the `animationController` and in `activeAnimations`. 

Also, it gets cumbersome to return every action as a closure from the hook i.e.
```reason
let (anim1Value, restartAnim1, pauseAnim1, resumeAnim1, ..., hooks) = Hooks.animation(...);
```
And if you only returned certain closures from the animation, it would be unclear that others exist, like `Animation.pause`, etc.

src/UI/Animation.re
zbaylin: Okay I changed both references to use the spread.

src/UI/Animation.re
glennsl: That isn't exactly what I mean, but you basically describe the difference between functional and OO paradigms, and functional is what's idiomatic to Reason. But you wouldn't typically use this kind of low-level API directly anyway. Instead you would use a declarative API like the `AnimatedView` component I referred to earlier.

OhadRau: Kinda skimmed through this and it seems rather minor, but I'm trying to understand the two arguments here. I think what @zbaylin might be saying here is that it doesn't necessarily make sense to hold onto the generated closures (resume, pause, etc.) because they won't always be valid. E.g. we don't want a `resume` stored as part of the component when `resume` isn't a valid operation, and by storing all these in a single object we're giving control of "what can happen when" to the object. While OOP doesn't always make sense, this personally seems kinda nitpicky to say one or the other. Could you clarify what the API would look like the way you imagine it @glennsl?

zbaylin: If you hide the animation control in a component, you have no API for changing the behavior of the animation from an external component, which is desirable. Also your example reworks how the animation hook works in general, by changing what is responsible for starting the animation and storing it's state (the user vs. the hook).

Also, there is no way of restarting here if the animation is finished. It wouldn't make sense to store it in `Running` because you're unaware if the animation is running, or if it is finished.

The way I have it now has the least amount of breaking changes to the animation API, and incorporates restarting etc. The only real difference to what you are describing and what I have is storing the state of the animation (`Paused, Active`, etc.), which you could easily do the way I have it now. In this way the API you describe is almost identical to mine -- you're calling a pause or resume function that doesn't take an animation object, because it's assume which one you're referring to.

glennsl: > If you hide the animation control in a component, you have no API for changing the behavior of the animation from an external component, which is desirable.

That's why there is a low-level and high-level API. You could also just lift the state up a level, or several, and pass it down through a prop.

> Also your example reworks how the animation hook works in general, by changing what is responsible for starting the animation and storing it's state (the user vs. the hook).

I'm not sure what gives you that idea. You could still have either, or both.

> Also, there is no way of restarting here if the animation is finished. It wouldn't make sense to store it in Running because you're unaware if the animation is running, or if it is finished.

I skipped `restart` here mostly for simplicity, but I also don't think a general `restart` function is a particularly good design, or very useful. You _ought_ to know what state the animation is in, and then act on that. That's the basic idea behind this.

> The only real difference to what you are describing and what I have is storing the state of the animation

Yeah, pretty much. Like I said, it's basically functional vs OOP. One is idiomatic (and a better design in general I think, which is why I use a functional language), the other is not. In React, my approach is analogous to a [controlled component](https://reactjs.org/docs/forms.html#controlled-components) and yours to an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html). The former is recommended because it gives you more control.

bryphe: From an API perspective - I agree with @glennsl, I would prefer to stay away from objects / classes in our public API. The syntax is not common, or even well-documented, in Reason. 

> but I'm concerned because in the bigger picture there are cognitive costs to using an entirely different API design using features that are uncommon and therefore usually poorly understood by relative newcomers.

Agree with this 👍 

> Classes are used throughout revery. Every single node is represented as a class, and there are other situations where classes are used too.

Note that  we primarily use classes _internally_ - we __rarely__ expose them to the end-user! It's simply an implementation detail (there is one exception of the top of my head - the `ref` returns a node instance... but I'd actually prefer to remove this and find a more functional model) 😄 This is because they are cumbersome and the syntax is non-obvious.

In fact, I'd actually prefer to replace our node-class-hierarchy with a functional, immutable tree. The classes were convenient containers of mutable state (for our current drawing strategy), and the nominal sub-typing was in really a crutch for me while I was still getting the hang of the 'functional' approach to this domain. I hope that with the Skia work we can keep less state around, and perhaps remove most of that mutable state, and re-design this.

> Primarily, persistent mutable state. The animation controller class stores not only an animation itself, but the last activeAnimation. The de facto solution for this has been to iterate over activeAnimations to find any that contain a matching animation, but since we need to keep access to this after the animation "expired", this solution wouldn't work.

I think we might not need to actually store state. Using the animation 'id' should be enough as handle - which is immutable, and could be contained in the closure.

> Also, it gets cumbersome to return every action as a closure from the hook i.e

I actually prefer this to the class-approach. And I think the number of functions could be limited. For example, maybe exposing `resume` as an action from the start doesn't make sense - but it could be returned from `pause`.

Something like: 
```
let (anim1Value, restartAnim1, pauseAnim1, hooks) = Hooks.animation(...)
```

We could just stick with `restart` and `pause` for now. `pause`, when called, would return a `resume` method. This would give us the tools to build other abstractions on top (simpler hooks, components, etc)

This is similar to how people are using ReactJS hooks for animation - for example, [`react-spring`](https://www.react-spring.io/docs/hooks/use-springs) has examples like this:
```
const [springs, set, stop] = useSprings(number, index => ({opacity: 1}))

// Update springs with new props
set(index => ({opacity: 0}))
// Stop all springs
stop()
```

Another way I look at this - we want to make Revery as accessible as possible to users coming right from ReactJS. If we can stick to that 'functional' paradigm - this will map more neatly to React+hooks, as opposed to forcing users to learn the awkward class syntax on top of everything else

zbaylin: Okay I switched over to a more functional approach like @bryphe mentioned. Hopefully this will suffice.

zbaylin:
All right, everything passes CI now!

bryphe:
Thanks @zbaylin !

I just noticed one more thing - sorry I missed this before!
image

It looks like the 'Pause' and 'Restart' buttons are a little funky - would you mind trying a smaller fontsize for those? Thanks for adding it to the example to test out!

zbaylin:
image

Does just increasing the size of the buttons work?

bryphe:
Sure, thanks @zbaylin !

PR #580, opened 2019-10-02
Escape path to allow copying into paths with spaces.

CrossR:
This also needs the same fix applying over in `reason-font-manager`, so assuming this fix is correct should merge that first and update the deps here as well. There is also the same fix required in Oni2.

https://github.com/revery-ui/reason-font-manager/pull/4


bryphe:
Thanks for the fix here, @CrossR ! Looks great 👍

PR #587, opened 2019-10-05
Mouse - API: Allow capture to complete when mouse leaves window

bryphe:


src/UI/Mouse.rei
```diff
@@ -39,6 +39,7 @@ let setCapture:
     ~onMouseLeave: mouseMoveHandler=?,
     ~onMouseOver: mouseMoveHandler=?,
     ~onMouseOut: mouseMoveHandler=?,
+    ~onMouseLeaveWindow: unit => unit=?,
```
kyldvs: Should this be added to the `registerListeners` API as well then?

bryphe: Good point @kyldvs (and I'll add the corresponding `onMouseEnterWindow` event too)

bryphe: Added in latest and tested it out in the hover example 👍

PR #594, opened 2019-10-11
chore: use rebez and rench as deps instead of resolutions

lessp:
Would it work to use these as `dependencies` instead of resolutions?


bryphe:
Unfortunately I don't think we can bring these in as `devDeps` because they're actually used at runtime (aka, they are real `deps`). We could publish NPM packages for them, though, so we don't need the resolution 👍

lessp:
> Unfortunately I don't think we can bring these in as `devDeps` because they're actually used at runtime (aka, they are real `deps`). We could publish NPM packages for them, though, so we don't need the resolution 👍

Sorry, looks like I was tired while submitting this! 😄

What I really meant was if we could replace the `*` in `"dependenices"` (and remove them in resolutions) to point directly to the Github-repo, meaning that consumers would not have to specify the resolutions for these as well!

lessp:
I think this is good to merge, unless I'm missing any potential issues this could create! It removes the noise to have to specify the resolutions for `rench` and `rebez` when consuming Revery!

bryphe:
Oh nice, thanks a lot for fixing this up @lessp ! 🎉 

>  It removes the noise to have to specify the resolutions for rench and rebez when consuming Revery!

Definitely! Ideally it's just a dependency on `revery` and you're good to go 👍 

Looks good to me - bringing this in now!

PR #597, opened 2019-10-11
Bugfix: Command+Q doesn't quit the application

bryphe:
__Issue:__ This is a regression from the SDL2 upgrade. We were expecting that we get a 'WindowClosed' event prior to a 'Quit' event - this is true in the case of pressing the 'x' button in the window, but not the case when quitting the app external to a window - like via `Command+Q` in OSX.

__Fix:__ When we get a Quit event, try to close all the windows. In the 'WindowClosed' before, there won't be any windows around - so the logic will behave as before. However, in the `Command+Q` case, the logic will now attempt to close all open windows.

__BREAKING:__ This also hooks up the `App.quit` API to use the `canQuit` logic.

### DIFF
--- a/src/Core/App.re
+++ b/src/Core/App.re
@@ -44,7 +44,7 @@ let _tryToClose = (app: t, window: Window.t) =>
   };
 
 let _tryToCloseAll = (app: t) => {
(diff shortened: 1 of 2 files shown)


bryphe:
Need to find to have integration tests for cases like this 🤔

PR #618, opened 2019-10-26
chore: bump rebez so as not to leak pesy

lessp:
I'm getting this with `kitty` if I don't put `pesy` in resolutions. I noticed over at the [rebez-repo](https://github.com/jchavarri/rebez/pull/5) that `pesy` was put as a devDependency which seems to solve the issue. Just needs a bump!

Screenshot 2019-10-26 at 22 09 45


bryphe:
Great! Thanks for catching this @lessp 🎉

Issue #467, opened 2019-04-22
Scollview don't display the same when underfilled


Et7f3:
- Operating System: Windows 10
- Revery Version: 0.18
- OCaml version: 4.7
- Native, Bytecode, or JS build: Native

- Link to github repo: Examples from revery

- Steps to reproduce: empty the viewlist on the side, execute and scroll 

- Actual Result:
![image](https://user-images.githubusercontent.com/29592775/56519539-9a26ed00-6542-11e9-8f40-7bca695a47e9.png)
or Expected Result

- Expected Result:
![image](https://user-images.githubusercontent.com/29592775/56519592-baef4280-6542-11e9-9bef-24dfbf4cdfd7.png)

maybe add a option to display element when under filled:
ForceTop
ForceBotton
ForceMiddle

- Additional Information: I have scrolled with a asus touchpad

- Search terms used: none

CrossR:
I think this is impacting the file explorer view in Oni2.

When a scroll view is underfilled, scrolling down causes the entire view to move to the bottom. Scrolling up will cause it to return.

We could add a simple check to the scroll calculation that if the vertical scroll bar isn't visible (ie no content is off screen), then the top of the view remains the same rather than being updated. I gave that a quick test and it seemed to work, but I'm not sure if / what the expectations are around the `bounce` attribute are for a scroll view that is fully on screen. Should it still bounce a little?

PR #621, opened 2019-10-27
Fix ScrollView freaking out when scrolling without overflow

glennsl:
This fixes an issue with ScrollView where scrolling with the scroll wheel when not overflowing causes the viewport to rapidly "invert" or "jump" from top to bottom. This is because `maxHeight` is calculated to be negative when the content is smaller than the ScrollView. Fixed by clamping to 0.

### DIFF
--- a/src/UI_Components/ScrollView.re
+++ b/src/UI_Components/ScrollView.re
@@ -60,7 +60,7 @@ let make =
         let childMeasurements = inner#measurements();
         let outerMeasurements = outer#measurements();
 
-        let maxHeight = childMeasurements.height - outerMeasurements.height;
+        let maxHeight = max(0, childMeasurements.height - outerMeasurements.height);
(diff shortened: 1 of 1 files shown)


bryphe:
So nice to have a fix for this! Thank you @glennsl - this is one of those annoying, nagging issues that has been around for a while. I believe this should address, or at least improve, https://github.com/onivim/oni2/issues/616

glennsl:
Could this also have fixed https://github.com/revery-ui/revery/issues/467? I see it's referenced in the Oni2 issue, but the description itself seems like it might be an issue where it gets pinned to the bottom rather than flicking between top and bottom. It's not unlikely that this could have an effect on that issue in any case though.

bryphe:
Ah, yes, I believe this will fix #467 as well 👍

PR #624, opened 2019-10-30
Et7f3/optimise/object

Et7f3:
This PR reduce indirections `val _something = ref(value)` => `val mutable _something = value` + change switch on bool to if (I have seen while reducing indirections).

Is there a bench tool to see memory usage: I would like to see memory impact on oni2 for instance ☺.

### DIFF
--- a/src/Core/Performance.re
+++ b/src/Core/Performance.re
@@ -39,9 +39,7 @@ let isBenchmarking =
 
 let bench: (string, performanceFunction('a)) => 'a =
   (name, f) =>
-    switch (isBenchmarking) {
-    | false => f()
-    | true =>
+    if (isBenchmarking) {
       nestingLevel := nestingLevel^ + 1;
       let startTime = Unix.gettimeofday();
       let startCounters = GarbageCollector.counters();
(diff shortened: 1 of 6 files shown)


bryphe:
Cool! Thanks for looking into this, @Et7f3 !

Is there a bench tool to see memory usage: I would like to see memory impact on oni2 for instance ☺.

We have some cursory benchmarks that measure both runtime & memory (both for Onivim 2 and Revery!):
- `esy '@bench' install`
- `esy '@bench' run`

I don't think our existing benchmarks would cover these scenarios, though - might need to add benchmarks to exercise those code paths - kind of like: https://github.com/revery-ui/revery/blob/3beb78f1fff6867891a1edf799803eab8da43539/bench/lib/LayoutBench.re#L43

bryphe:
I'm curious too - I wonder if there is a difference between `mutable` and `ref` from a performance perspective? 🤔 

It does make the code cleaner, though!

Et7f3:
an old discussion with octachron in #beginners chan of OCaml discord server.
image
https://discordapp.com/channels/436568060288172042/585511202759770135/592153447768391691

Et7f3:
an old discussion with octachron in #beginners chan of OCaml discord server.
image
https://discordapp.com/channels/436568060288172042/585511202759770135/592153447768391691

bryphe:
Oh, neat - TIL! That's great to know!

Et7f3:
More than I hoped.😍

PR #625, opened 2019-10-31
Animation: Implement 'Hooks.transition' hook

### LINKED ISSUE #121: Animation: Implement 'Hooks.transition' hook
Today, we have a `Hooks.animation` hook here:
https://github.com/revery-ui/revery/blob/master/src/UI_Hooks/Revery_UI_Hooks.re and https://github.com/revery-ui/revery/blob/master/src/UI_Hooks/Animation.re

That is used as follows: 
```
      let (rotationY, pauseRotationY, restartRotationY, hooks) =
        Hooks.animation(
          Animated.floatValue(0.),
          Animated.options(
            ~toValue=6.28,
            ~duration=Seconds(4.),
            ~delay=Seconds(0.5),
            ~repeat=true,
            (),
          ),
          hooks,
        );
```
(from https://github.com/revery-ui/revery/blob/master/examples/Hello.re)

I think it'd be convenient to have a `Hooks.transition` hook, that would work as follows:
```
let currentValue = Hooks.transition(1.0, { duration: Seconds(1) });
```

The idea is that you could use this along with some other events, for example:
```
let (opacity, setOpacity) = Hooks.state(1.0);

let transitionedOpacity = Hooks.transition(opacity, { duration: Seconds(1) });

let onMouseDown  = () => setOpacity(0.5);
let onMouseUp = () => setOpacity(1.0);

<view style={Style.make(~opacity=transitionedOpacity, ())} ... />
```

This would enable a smooth transition between the opacity values, as opposed to just directly switching from 0.5 <-> 1.0. The `Hooks.transition` hook could leverage `Hooks.state` under the hood to keep track of the last value. If the last value is different, it could start an animation and use the animated value. Otherwise, it could just return the current value.

Et7f3:
fix #121
We call it like `Hooks.state`
https://github.com/revery-ui/revery/blob/Et7f3/useTransition/examples/Hello.re#L11-L12 and it return a setter
https://github.com/revery-ui/revery/blob/Et7f3/useTransition/examples/Hello.re#L85
I have added some
```
      Printf.printf("transitionedOpacity: %f\n", transitionedOpacity);
      flush(stdout);
```
because opacity is a no-op on my windows (didn't try on other os)
If we have more hooks maybe add a Hooks demo for them.

Also It miss the last update. 😢


src/UI_Hooks/Animation.rei
```diff
@@ -1,6 +1,19 @@
 open Revery_UI.Animated;
 open Revery_UI.React.Hooks;
 
+let animation':
```
glennsl: This isn't a great name (obviously), and doesn't seem like a very useful hook to expose in general either. Might it be better to move the `transition` hook in here instead?

src/UI_Hooks/Transition.re
```diff
@@ -0,0 +1,24 @@
+/* Hooks specific to Revery */
+open Revery_Core;
+open Revery_UI.Animated;
+
+let transition =
+    (toValue, ~delay=Time.Seconds(0.0), ~duration=Time.Seconds(1.), slots) => {
+  let repeat = false;
```
glennsl: This seems like unnecessary indirection. IMO, at least, it would be clearer to just set `~repeat=false` directly.

src/UI_Hooks/Transition.re
```diff
@@ -0,0 +1,24 @@
...
+    Animation.animation'(
+      floatValue(toValue),
+      options(~toValue, ~duration, ~delay=Time.Seconds(0.0), ~repeat, ()),
+      slots,
+    );
+  let setAnim = toValue => {
+    let animation =
+      tween(value, options(~toValue, ~duration, ~delay, ~repeat, ()));
+    // only for cleaning purpose we don't restart it
+    let _ignore = pauseAnim();
```
glennsl: Wildcard let bindings is bad practice, since it's a bit of a footgun. I think there might even be a compiler warning for this in 4.09. The reason is that if an argument is added to the end of `pauseAnim`, this will be a partial application that doesn't produce the side-effects you expect. and is therefore a very hard bug to track down.

But if you add a type annotation that can't happen.

    let _: unit => unit = pauseAnim();

src/UI_Hooks/Transition.re
```diff
@@ -0,0 +1,24 @@
...
+    let newActiveAnim = {animation, update: None, complete: None};
+    activeAnimations := [newActiveAnim, ...activeAnimations^];
```
glennsl: I see this is done in the animation hook as well, but it is a really, really bad idea to mutate global refs from other modules. `activeAnimations` shouldn't even be exposed from `Animated`, but instead should expose functions that mutate it indirectly. Uncontained side-effects can be extremely hard to track down, and by going through functions you can for example do checks that maintain internal invariants, and easily log every change.

src/UI_Hooks/Animation.rei
Et7f3: Maybe rename it `animation_internal` because I use it to re-implement `transition`. I decided to expose because maybe user want to implement such hook and need this facilitie. For the naming I have used what I have seen in opium https://github.com/rgrinberg/opium/blob/master/opium/lib/app.ml#L225

examples/Hello.re
```diff
@@ -8,7 +8,10 @@ let logo = {
 
   (~children as _: list(React.syntheticElement), ()) =>
     component(hooks => {
-      let (logoOpacity, _setOpacity, hooks) = Hooks.state(1.0, hooks);
+      let (transitionedOpacity, setTransitionedOpacity, hooks) =
+        Hooks.transition(1., ~duration=Seconds(1.), hooks);
+      Printf.printf("transitionedOpacity: %f\n", transitionedOpacity);
+      flush(stdout);
```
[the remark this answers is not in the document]
Et7f3: Thanks. It is much shorter. Some C reflex.

src/UI_Hooks/Transition.re
Et7f3: I use it two time 🤷‍♂ but yeah I agree.

src/UI_Hooks/Transition.re
Et7f3: `let _` is different of `let _ignore` the value returned by `pauseAnim` is a function to relaunch animation that is not the desired behaviour and don't want to copy-paste.

src/UI_Hooks/Animation.rei
glennsl: I don't think `animation_internal` is much better. A good name is descriptive. And if an exposed function is hard to describe, that's usually a good sign that it shouldn't be exposed.

In this case I think it would be better to just move the `transition` hook into `Animation`, and thereby  avoid exposing internals. Unless of course there's a really good reason I don't know about for having it in a separate module.

src/UI_Hooks/Transition.re
glennsl: How is it `let _` and `let _ignore` different? They'll both discard any type of value.

src/UI_Hooks/Transition.re
Et7f3: fixed in https://github.com/revery-ui/revery/pull/625/commits/c24130b3ea17a76eb7ad9decf551d97300975e84

src/UI_Hooks/Transition.re
Et7f3: I didn't know annotation can disable partial application warning. Thanks.
fixed in https://github.com/revery-ui/revery/pull/625/commits/b5485f07046bf2e8becf7886f21381965fbc6be1

src/UI_Hooks/Animation.rei
Et7f3: > Unless of course there's a really good reason I don't know about for having it in a separate module.

Currently we have one hook in each file even they are one-line function, see ref, state, reducer. I have seen in pr history to find out why we have made this change: I found that https://github.com/revery-ui/revery/pull/494/commits/6eb9011c7d06c8a93e7865ccd68cee450e985f79.
I think it is good to expose it, mainly if the end user want to add such hook.
I can hide `animation'` only from outside when we are rexporting variable.

Et7f3:
> Also It miss the last update. 😢

Now I explicitly set the value when the animation end: https://github.com/revery-ui/revery/pull/625/commits/b5485f07046bf2e8becf7886f21381965fbc6be1#diff-cd777f7799e2787af430fc84ce54bf45R22. Maybe there is a cleaner way. Ihave tried with reducer like animation hook and it don't helped.

src/UI_Hooks/Animation.rei
glennsl: From that, it seems like they're just in separate modules for the sake of the API. You can still get the same API if you implement it as `Animation.transition` and then re-export it as `Transition.transition`. You could even implement it as `Animation.Transition.transition` and re-export in `Transition.re` with `include Animation.Transition`.

Et7f3: fixed inn https://github.com/revery-ui/revery/pull/625/commits/a7d95327f80d1a8184fdc791139c629ac6024a11

zbaylin:
This is awesome and solves a lot of the a-functionality the current API uses (partly my fault 😨). One thing I might add would be an `immediate` parameter to the `setX` function, i.e. `setOpacity(~immediate=true, 1.0)`.

Also, I know it's just convention, but it might make more sense to use the name `transitionXTo` (i.e. `transitionOpacityTo(1.0)`) rather than `setX`, which sounds more immediate.

I really hope this can get merged soon, because it will help **a lot** with my efforts on revery-material.

Great work @Et7f3 !!

bryphe:
Awesome work @Et7f3 ! (and thanks @glennsl and @zbaylin for reviewing it!)

Such a nice primitive to have now 🎉

PR #628, opened 2019-11-03
UI - Hooks: Animation - useSpring hook

### LINKED ISSUE #131: Animation: Springs - 'useSpring' hook?
Inspired by [react-spring](https://github.com/react-spring/react-spring) - springs are excellent tools for creating an interactive and animated UI. Another great description comes from the [react-motion](https://github.com/chenglou/react-motion) repo.

# Proposal

We could enable easy spring-based animations via a `useSpring` hook. There's been thinking about this already here: https://medium.com/@drcmda/hooks-in-react-spring-a-tutorial-c6c436ad7ee4

Our hook could look like:
```
let currentVal = useSpring({ currentValue, destinationValue, springConfiguration });
```

`springConfiguration` could have configurable properties:
- `stiffness`
- `damping`

And, like [react-motion](https://github.com/chenglou/react-motion), it would be helpful to have a good set of presets for this.

Under the hood, we'd need to use `setState` to store the current `position`, `velocity`, and `acceleration`. We can use [Hooke's Law](https://en.wikipedia.org/wiki/Hooke%27s_law) to determine the acceleration based on the force (`F=kd=ma`).

We'll also need to update this every `tick` - we'll need to generalize our `Animation` framework a bit to allow for arbitrary tick functions (right now, we always call the `tickAnimation` function).

bryphe:
This is a prototype of a `useSpring` hook - was just hacking on it during the plane ride. Fix #131 

The usage looks like this:
```
      let%hook curr = Hooks.spring(0.5, Hooks.Spring.Options.create(logoOpacity));
```

Where `Hooks.Spring.Options` exposes some parameters:
- `stiffness` - spring stiffness (spring constant `k`) 
- `damping` - damping (constant inverse force proportional to velocity)
and an `initialValue`

Along with `Hooks.animation` and `Hooks.transition`, this is another tool for our animation toolbox in Revery.

Some use cases I'd like to explore with it in Onivim:
- Spring-based animation for cursor size (moving between a full cursor an a smaller one in operator-pending mode - we used this in Onivim v1)
- Spring-based animation for cursor movement (might be distracting, but could be useful when pairing to give additional visual context about the velocity and direction of the cursor).

__TODO:__
- [x] Move the core spring logic out of hooks and alongside the animation stuff
- [x] Add presets like react-motion: https://github.com/chenglou/react-motion/blob/9cb90eca20ecf56e77feb816d101a4a9110c7d70/src/presets.js
- [x] Add `rei` for core spring logic
- [x] The `initialVal` is kind of mixed up between the `Hooks.Spring.Options.create` - we should be passing `logoOpacity` as the first argument of `Hooks.spring`.
- [x] The final settling needs adjustment - we might need to tweak the threshold of when we consider the animation 'complete' (add parameter to options)
- [x] Disable ticker when 'settled'
- [x] Clean up module hacks
- [x] Example: Add sliders for stiffness / damping to experiment with different values
- [x] Remove duplicate hooks
- [x] Switch to use absolute time instead of hardcoded `0.016`


src/UI_Hooks/Spring.re
```diff
@@ -0,0 +1,118 @@
...
+      damping,
+      springConstant,
+    };
+  };
+
+  type t = {
+    x: float,
+    v: float,
+    a: float,
+  };
```
glennsl: Could these field names be made more descriptive? I would guess `x`, `v`, and `a` refer to the current value, the velocity and the acceleration, but it's not entirely obvious. Partly that's because later in the module `v` seem to mean "value", not "velocity", but also a different kind of value than `x` here (initial vs current value, I guess?).

src/UI_Hooks/Spring.re
```diff
@@ -0,0 +1,118 @@
...
+    Effect.effect(
+      If((!=), opts),
+      () => {
+        dispatch(ResetOptions(opts));
+        None;
+      },
+      hooks,
+    );
+
+  let slots =
```
glennsl: `slots`? Should it not be `hooks`? Hope we can get the `%hook` ppx soon :)

src/UI_Hooks/Spring.re
bryphe: Good point! I'll adjust these to `position`, `velocity`, and `acceleration`, respectively

src/UI_Hooks/Spring.re
bryphe: Updating to use the new `%hook` PPX - so this should be addressed as part of that 👍

PR #630, opened 2019-11-04
[WIP] Clean up animation API and implementation

glennsl:
I initially just wanted to add a well-defined interface to the Animation API, but ended up simplifying the implementation quite a bit too, which is of course a thing you can do when you have a well-defined interface. This also resulted in much simpler hooks.

There's still much that can be done here, like modelling possible animation states more cleanly and simplifying the chaining, but I didn't want to stray too far from the original API and wanted to get some feedback on whether I've already gone too far before going even further.

### DIFF
--- a/src/UI/Animation.re
+++ b/src/UI/Animation.re
@@ -21,19 +21,13 @@ type animationDirection = [
 ];
 
 module Make = (AnimationTickerImpl: AnimationTicker) => {
-
-  module AnimationId =
-    UniqueId.Make({});
-
   type animation = {
-    id: int,
     delay: float,
     mutable startTime: float,
     duration: float,
     initialValue: float,
     toValue: float,
     mutable value: float,
-    mutable time: float,
     repeat: bool,
     easing: Easing.t,
     direction: animationDirection,
@@ -113,14 +107,12 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
   let tween =
       (initialValue: float, animationOptions: animationOptions) => {
     let animation = {
-      id: AnimationId.getUniqueId(),
       delay: Time.to_float_seconds(animationOptions.delay),
       duration: Time.to_float_seconds(animationOptions.duration),
       toValue: animationOptions.toValue,
       repeat: animationOptions.repeat,
       value: initialValue,
       startTime: Time.to_float_seconds(AnimationTickerImpl.time()),
-      time: Time.to_float_seconds(AnimationTickerImpl.time()),
       initialValue,
       easing: animationOptions.easing,
       direction: animationOptions.direction,
--- a/src/UI/Animation.re
+++ b/src/UI/Animation.re
@@ -33,38 +33,20 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
     initialValue: float,
     toValue: float,
     mutable value: float,
+    mutable time: float,
     repeat: bool,
     easing: Easing.t,
     direction: animationDirection,
     mutable isReverse: bool,
-  };
-
-  type activeAnimation = {
-    animation,
-    update: option(float => unit),
-    complete: option(unit => unit),
+    mutable running: bool,
+    mutable update: option(float => unit),
+    mutable complete: option(unit => unit),
   };
 
   type playback = {
     pause: unit => unit,
     stop: unit => unit,
   };
-
-  let activeAnimations: ref(list(activeAnimation)) = ref([]);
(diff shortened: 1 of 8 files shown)


src/UI/Animation.re
```diff
@@ -78,8 +51,6 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
     direction: animationDirection,
   };
 
-  let floatValue: float => animationValue = (v: float) => {current: v};
```
Et7f3: I don't know why we have this. Maybe @bryphe have some other thing in mind here e6e61d011d7afdc4de9c527da64c1aea85df8bb4.

src/UI/Animation.re
```diff
@@ -21,54 +21,27 @@ type animationDirection = [
...
+    mutable update: option(float => unit),
+    mutable complete: option(unit => unit),
   };
 
   type playback = {
     pause: unit => unit,
     stop: unit => unit,
   };
 
-  let activeAnimations: ref(list(activeAnimation)) = ref([]);
```
Et7f3: So we don't track them anymore 🤔 only brisk-reconcillier will keep them ?

src/UI/Animation.rei
```diff
@@ -0,0 +1,60 @@
+open Revery_Core;
+open Revery_Math;
```
Et7f3: unuseed ? even in Animation.re ?

src/UI/Animation.re
```diff
@@ -21,54 +21,27 @@ type animationDirection = [
 ];
 
 module Make = (AnimationTickerImpl: AnimationTicker) => {
-  type animationValue = {mutable current: float};
-
-  module AnimationId =
-    UniqueId.Make({});
-
   type animation = {
-    id: int,
```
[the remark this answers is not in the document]
glennsl: It wasn't used for anything, and since animations are now independent I don't see what it might be useful for in any case.

src/UI/Animation.re
```diff
@@ -130,63 +91,84 @@ module Make = (AnimationTickerImpl: AnimationTicker) => {
...
 
   let getTime = () => AnimationTickerImpl.time();
 
+  let restart = (animation) => {
+    animation.startTime = Time.to_float_seconds(getTime());
+    animation.value = animation.initialValue;
+    animation.running = true;
+  };
+
+  let pause = (animation) => {
```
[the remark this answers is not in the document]
glennsl: And it would be even cleaner with a variant to more accurately model the possible states I think.

src/UI/Ui.re
```diff
@@ -136,7 +136,7 @@ let start = (window: Window.t, element: React.syntheticElement) => {
     );
 
   Window.setShouldRenderCallback(window, () =>
-    uiDirty^ || Animated.anyActiveAnimations()
```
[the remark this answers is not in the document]
glennsl: The idea is that animation should be managed on a component level instead, typically driven by hooks.

src/UI/Animation.re
```diff
@@ -21,54 +21,27 @@ type animationDirection = [
...
 module Make = (AnimationTickerImpl: AnimationTicker) => {
-  type animationValue = {mutable current: float};
-
-  module AnimationId =
-    UniqueId.Make({});
-
   type animation = {
-    id: int,
     delay: float,
     mutable startTime: float,
```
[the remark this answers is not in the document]
glennsl: Yeah, I felt that was a bit iffy as well. It would be cleaner if using a variant to model state instead I think. Then it could be an immutable field of a `Running` state, but of course the state itself would be a mutable field of the animation.

src/UI/Animation.re
glennsl: Yes, it would typically be gooks driven is the idea.

src/UI/Animation.rei
glennsl: Indeed. But in `Animation.re` `interpolate` is at least still used.

glennsl:
It just dawned on me that chaining isn't going to work currently. I was putting that off and forgot that it would still require a background task to drive it, so I need to refactor the chaining too for this to make sense.

glennsl:
Ok, so I went a bit further in #631. That's a pretty drastic departure from the current API though, so consider it an experiment more than anything. Keeping this open in case this approach is considered better.

glennsl:
#631 worked out great I think, so I'm closing this in favor of that.

PR #631, opened 2019-11-05
Animation as a pure function of time

glennsl:
This supercedes #630.

The idea here is to model animation as a pure function of time, and to manipulate time instead of manipulating the animation. There are many benefits to this approach:

- Simpler implementation. No need for a functor, and no need to carry around and account for every bell and whistle in the animation itself. It's just a function, and (pure) functions can easily be composed.
- Easier to reason about since the animation never changes. Only time does. For example, a chained animation can't be started prematurely from the other side of the codebase.
- Cleanly separates specification from interaction.
- Composes better. The `Chain` module is reduced to a 4-line function (`andThen`), for example. Animations can also be `zip`ed, to run animations in parallel, and `map`ed, which `tween` and `ease` are convenient forms of.
- Much easier to test.
- Animations can be reused, run in parallel or thrown away and recreated on every render.
- It's data-driven, which is more react-y, not action-driven. An action can easily be turned into data when needed using `Hooks.state`. To turn data into action you need to use `Hooks.effect`.

I'm pretty happy with this now. The low-level API could have a bit better naming and ergonomics perhaps, and it also needs mroe tests and ocoumentation. The implementation could also be made much simpler if `Time` was just a float (see #632). There is A LOT of back and forth with `Time.toSeconds` and `Time.ofSeconds`.

### TODO
- [x] Alternating repeat
- [x] Pause timers when animation completes (managed by the hooks)
- [x] Better Time API? (See #632)
- [x] Tests
- [x] Documentation

### Usage
There are several different usage patterns in the examples, to try out different styles. In no particular order:

#### `transition` hook
Exactly the same as before.
```reason
let%hook (transitionedOpacity, transitionOpacityTo) = Hooks.transition(1.);
```

#### Manual timer (no longer used, but still possible) and `zip`
```reason
let (isTimerActive, setTimerActive) = Hooks.state(true);
let (time, resetTimer) = Hooks.timer(~active=isTimerActive, ());

let ((rotationX, rotationY), _) =
  Animation.(
    zip((
      animate(Time.seconds(9.))
      |> tween(0., 6.28)
      |> repeat
      |> delay(Time.seconds(1.)),
      animate(Time.seconds(4.))
      |> tween(0., 6.28)
      |> repeat
      |> delay(Time.seconds(0.5)),
    ))
    |> apply(time)
  );
```

#### `animation` hook with zipped animation and playback control
`setShouldRotate(false)` and `setShouldRotate(true)` is effectively `pause` and `resume`, and `resetRotation` will restart the animation.
```reason
let%hook (shouldRotate, setShouldRotate) = Hooks.state(true);
let%hook ((rotationX, rotationY), _animationState, resetRotation) =
  Hooks.animation(rotationAnimation, ~active=shouldRotate);
```

#### Chaining
```reason
let bounceAway =
  animate(Time.milliseconds(100.))
  |> ease(Easing.cubicBezier(0.23, 1., 0.32, 1.))
  |> tween(float(origin), float(origin + force));

let bounceBack =
  Animation.animate(Time.milliseconds(800.))
  |> ease(Easing.cubicBezier(0.23, 1., 0.32, 1.))
  |> tween(float(origin + force), float(origin));

bounceAway |> andThen(~next=bounceBack) |> map(int_of_float);
```

### Triggered animation
This animation is started when `setBouncingState(Bouncing(force))` is called, runs the animation until it completes and then sets `bouncingState` back to `Idle`.
```reason
let%hook (bouncingState, setBouncingState) = Hooks.state(Idle);
let%hook (actualScrollTop, _bounceAnimationState, resetBouncingAnimation) =
  switch (bouncingState) {
  | Idle => Hooks.animation(Animation.const(actualScrollTop))

  | Bouncing(force) =>
    Hooks.animation(
      bounceAnimation(~origin=actualScrollTop, ~force), ~onComplete=() =>
      setBouncingState(_ => Idle)
    )
  };
let setBouncingState = state => {
  resetBouncingAnimation();
  setBouncingState(state);
};

```


src/UI/Animation.re
```diff
@@ -1,232 +1,109 @@
...
-      List.filter(a => !isComplete(t, a), activeAnimations^);
+    {resume: resume};
   };
 
-  Event.subscribe(AnimationTickerImpl.onTick, t =>
-    tick(Time.to_float_seconds(t))
-  );
-};
+  {pause: pause};
+};
```
glennsl: This isn't actually used anywhere, and could be removed. But maybe it could be useful...? There is a possibility that I'm a hoarder, but it also serves to demonstrate how the animations could be action-driven.

src/UI_Components/ScrollView.re
```diff
@@ -45,6 +45,44 @@ let make =
...
+        | Delayed => actualScrollTop
+        | Running(value) => int_of_float(value)
+        | Complete(_) =>
+          setBouncingState(Idle)
+          actualScrollTop
+        };
+
+      | _ =>
+        actualScrollTop
+      };
```
glennsl: This is an awful mess.

src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -1,7 +1,109 @@
...
+  let (time, resetTimer, hooks) = timer(hooks);
+  let ((startValue, targetValue), setTargetValue, hooks) =
+    state((startValue, startValue), hooks);
+
+  let value =
+    Animation.(
+      animate(duration, ~delay)
+      |> tween'(startValue, targetValue)
+      |> (animate => animate(time))
+    );
```
glennsl: This is actually recreated on every call. The old one managed the animation as state, but that's not needed here.

src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -1,7 +1,109 @@
...
+  let (time, setTime, hooks) =
+    reducer(~initialState=Time.getTime(), t => t, hooks);
+  let (startTime, setStartTime, hooks) = Ref.ref(time, hooks);
+
+  let onTick = _dt => setTime(_t => Time.getTime());
+
+  let hooks =
+    mountIfEffect(If((!=), active), () =>
+      if (active) {
+        setStartTime(Time.ofSeconds(Time.toSeconds(Time.getTime()) -. (Time.toSeconds(time) -. Time.toSeconds(startTime))));
```
glennsl: So much `Time`...

src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -1,7 +1,109 @@
...
+      condition,
+      () => {
+        mountCleanup();
+        handler();
+      },
+      hooks,
+    );
+
+  hooks;
+};
```
glennsl: I'll probably fix https://github.com/briskml/brisk-reconciler/issues/27 later this week, so this can be removed.

src/UI_Components/ScrollView.re
glennsl: Also, note that this doesn't call `scrollUpdated` anywhere. It just "overrides" `actualScrollTop` temporarily. I.e. data-driven, not action-driven.

src/UI/Animation.re
```diff
@@ -1,232 +1,109 @@
...
+  | Running(float) // Returns a value in the interval [0., 1.]
+  | Complete(Time.t); // Returns elapsed time
+
+let getLocalTime = (delay, duration, startTime, time) => {};
+
+/**
+ * `time` is assumed to start at 0
+ */
+let animate =
+    (~delay=Time.Seconds(0.), ~repeat=false, duration, time) => {
```
glennsl: I think `delay` and `repeat` can be made separate function, which would greatly simplify this, and also provide a pretty nice and consistent API I think.

src/UI/Animation.re
```diff
@@ -1,232 +1,109 @@
...
-    update: option(float => unit),
-    complete: option(unit => unit),
+type t = Time.t => float;
+
+type result =
+  | Delayed
+  | Running(float) // Returns a value in the interval [0., 1.]
+  | Complete(Time.t); // Returns elapsed time
+
+let getLocalTime = (delay, duration, startTime, time) => {};
```
glennsl: Oops. Note to self: Remove this.

src/Core/Tick.re
```diff
@@ -104,9 +104,16 @@ module Make = (ClockImpl: Clock) => {
...
 
+  exception Stop;
+
   let interval = (f: callback, frequency: Time.t) => {
     let id = TickId.getUniqueId();
 
+    let f = t =>
+      try (f(t)) {
+      | Stop => _clear(id, ())
+      };
```
glennsl: This makes it possible to cancel the `interval` from inside the callback by using `raise(Tick.Stop)`, which is nice. And it's an unobtrusive feature. No-one needs to know this exists until they need it.

src/UI_Hooks/Revery_UI_Hooks.re
Et7f3:
```suggestion
        open Time;
        setStartTime(ofSeconds(toSeconds(getTime()) -. (toSeconds(time) -. toSeconds(startTime))));
```

examples/Hello.re
```diff
@@ -48,25 +33,23 @@ let logo = {
...
-                onClick={() => {
-                  pauseRotation() |> ignore;
-                  pauseRotationY() |> ignore;
-                  ();
-                }}
-                title="Pause"
+                onClick={() =>
+                  setTimerActive(!isTimerActive);
+                }
+                title={isTimerActive ? "Pause" : "Resume"}
```
glennsl: Now also a resume button, with a dead simple implementation. Before this would have required storing the resume function returned from calling the pause function for each animation to be called here. Instead we just flip a bit. Data-driven, not action-driven.

src/UI_Hooks/Revery_UI_Hooks.re
glennsl: Yeah, that would help :) But I'd rather get rid of the need to call `ofSeconds` and `toSeconds` entirely.

test/UI/AnimationTest.re
```diff
@@ -1,318 +1,71 @@
...
+    let combined = Animation.andThen(first, ~next=second);
 
-    TestTicker.simulateTick(Time.Seconds(1.));
-    expect.float(first.value.current).toBeCloseTo(5.);
-    // Simulate the end of the first, so the second can start
-    TestTicker.simulateTick(Time.Seconds(2.));
-    TestTicker.simulateTick(Time.Seconds(3.));
-    expect.float(second.value.current).toBeCloseTo(5.);
+    expect.float(fst(Animation.apply(Seconds(1.), combined))).toBeCloseTo(5.);
+    expect.float(fst(Animation.apply(Seconds(3.), combined))).toBeCloseTo(5.);
```
glennsl: I don't understand this test. It looks to me like it simulates 6 seconds, but that would run it way past completion of the second animation too. Alternatively, if `simulateTick` calls are absolute, not cumulative, the tests make sense. But if it's not cumulative, what's the point of having two `simulateTick` calls in sequence?

src/UI_Components/ScrollView.re
```diff
@@ -46,6 +60,20 @@ let make =
...
+      | Bouncing(force) => bounceAnimation(~origin=actualScrollTop, ~force)
+      };
+
+    let (actualScrollTop, bounceAnimationState, resetBouncing, slots) =
+      Hooks.animation(bounceAnimation, ~active=bouncingState != Idle, slots);
+
+    switch (bounceAnimationState) {
+    | Complete(_) => setBouncingState(Idle)
+    | _ => ()
+    };
```
glennsl: This is significantly better than the previous version, but the interaction between `bouncingState` and `bounceAnimation` and `bounceAnimationState` is still pretty complex. Can it be simplified, or is it a pattern that can be generalized and extracted into a single hook?

src/UI_Components/ScrollView.re
```diff
@@ -130,58 +158,25 @@ let make =
...
-                     bounceBackAnim,
-                   ),
-                 )
-              |> Chain.start(~update=v =>
-                   dispatch(ScrollUpdated(int_of_float(v)))
-                 );
-            setBouncingState(Bouncing(direction, playback));
+            setBouncingState(Bouncing(- delta * 2));
+            resetBouncing();
+            dispatch(ScrollUpdated(isAtTop ? 0 : maxHeight));
```
glennsl: Also kind of a lot going on here just to start the animation.

src/UI_Components/ScrollView.re
```diff
@@ -5,11 +5,8 @@ open Revery_UI_Primitives;
 
 module Hooks = Revery_UI_Hooks;
 
-type direction =
-  | Top
-  | Bottom;
 type bouncingState =
-  | Bouncing(direction, Animated.playback)
+  | Bouncing(int)
```
glennsl: The argument is now just the "force" of the bounce. A negative force means we're bouncing upwards, and that must mean we're at the top, while a positive force means the opposite. Simple and effective :)

src/UI_Components/ScrollView.re
```diff
@@ -45,6 +59,22 @@ let make =
...
+        Hooks.animation(
+          bounceAnimation(~origin=actualScrollTop, ~force),
+          ~onComplete=() => setBouncingState(Idle),
+          slots,
+        );
+      };
+    let setBouncingState = state => {
+      resetBouncingAnimation();
+      setBouncingState(state);
+    };
```
glennsl: Perhaps we could make this into a `statefulAnimation` hook which resets and runs an animation when a given predicate turns true. Sounds a bit complicated though, and maybe this is simple enough as it is.

bryphe:
I really like the approach here to model animations as pure functions of time. The current implementation always felt confusing... with the extra complexity of having to maintain an external animation 'state'. Sure is nice to be able to remove the functor and simplify the implementation!

test/UI/AnimationTest.re
bryphe: Good question... I think the implementation might have depended on the intermediate time to switch states from the first chained anim to the second? (Which could be indicative of a bug!)

Looks like the `simulateTick` are indeed absolute - so I suspect that the implementation depended on an intermediate frame to switch from the first animation to the second. 🤔 Would be curious to see if the original test fails if we remove the `TestTicker.simulateTick(Time.Seconds(2.));` line.l.

In any case - I think your new test case makes perfect sense!

src/UI_Hooks/Revery_UI_Hooks.re
bryphe: I wonder if the `Time` concept is kind of 'overkill'? 

The main reason I experimented it with was sometimes I find APIs confusing that use `float` for time - sometimes it is seconds, sometimes it is milliseconds - so I wanted to encode that in the type system - but in retrospect, perhaps the boilerplate is excessive compared to the benefits? Or there is a more streamlined representation we could use?

src/UI_Components/ScrollView.re
bryphe: Good question... It seems simplified versus the previous version, so that is a win!

One piece of the implementation - detecting the 'settled' state and going into an `Idle` mode - seemed similar to the spring implementation. But doesn't seem straightforward to generalize that piece, at this time.

test/UI/AnimationTest.re
```diff
@@ -1,318 +1,82 @@
...
-          delay: Time.Seconds(0.),
-          toValue: 10.,
-          repeat: true,
-          easing: Easing.linear,
-          direction: `Normal,
-        },
-      );
-    let _playback = Animated.start(myAnimation);
+  test("repeat", ({expect, _}) => {
+    let anim = Animation.(animate(Seconds(1.)) |> repeat |> floatify);
```
bryphe: These test changes are a good showcase of how much simpler the new API is! 🎉

examples/Hello.re
```diff
@@ -4,38 +4,25 @@ open Revery.UI;
 open Revery.UI.Components;
 
 let logo = {
+  let rotationAnimation =
+    Animation.(
+      zip((
+        animate(Seconds(9.)) |> tween(0., 6.28) |> repeat |> delay(Seconds(1.)),
```
bryphe: I like this new API vs the old `Animated.options` API - great idea @glennsl !

src/Core/Tick.re
bryphe: Interesting, I hadn't seen this technique before. It's kind of like using the exception as a typed effect (and reminds me of how 'suspense' in React works)

src/UI/Animation.rei
```diff
@@ -0,0 +1,25 @@
...
+  | Complete(Time.t); // Elapsed time
+
+let const: 'a => t('a);
+let animate: Time.t => t(normalizedTime);
+let delay: (Time.t, t('a)) => t('a);
+let repeat: t('a) => t('a);
+let alternatingRepeat: t('a) => t('a);
+let ease: (Easing.t, t(normalizedTime)) => t(normalizedTime);
+let tween: (float, float, t(normalizedTime)) => t(float);
+let map: ('a => 'b, t('a)) => t('b);
```
bryphe: I really like this new API surface for animations. Great use of composable functions to express arbitrarily complex animations (having the `animate` primitive be just a pure function of Time to normalized time is really cool, and means it is easy to extend with other building blocks - like step functions, etc!).

The `map` is a very useful addition and fills a gap we had with our current API - often, you want to map an animation to a non-float value (like `display(Visible)` if a value is greater than a certain threshold), or to a float value in a different 'space' - easy to express with `map`

test/UI/AnimationTest.re
glennsl: Part of that is because some of the complexity of the old implementation has been moved into the `timer` hook, which isn't tested. But the separation makes that pretty simple too, so I'm not too afraid of bugs.

Also, the animation state isn't tested yet, and there's some complexity in calculating the elapsed time, and it isn't entirely accurate in all cases either. Perhaps it's too much complexity. I think it's more featureful than needed.

src/UI_Components/ScrollView.re
glennsl: This code and comment is out of date, btw. The latest version of this is at the bottom of the PR description. It now uses an `onComplete` callback, so the switch on `bounceAnimationState` has essentially moved into the `animation` hook. Also, the `active` prop is no longer set since the animation automatically stops when it competes, and starts again on reset.

I think the pattern in the current implementation is pretty nice. Perhaps you can take something from that to use in the spring hook,

src/Core/Tick.re
glennsl: In OCaml, raising an exception is how you `break` from a `while` loop. I think that was part of the motivation for making them so light-weight and not trace the stack by default.

Also, if you haven't you should check out how algebraic effects will work when it lands. I think you're going to love that :slightly_smiling_face: (just don't get too excited, it'll still be a while...)

src/UI_Hooks/Revery_UI_Hooks.re
glennsl: Yeah, I might have some ideas :laughing: (For the uninitiated, see #632)

examples/Hello.re
Et7f3: Can we omit some argument ?

src/UI_Hooks/Revery_UI_Hooks.re
bryphe: So much nicer! 😎

src/UI/Animation.rei
```diff
@@ -10,24 +8,176 @@ module NormalizedTime: {
   let fromFloat: float => t;
 };
 
-type t('a);
+/**
```
bryphe: Wow! Very comprehensive and thorough docs. 

Great example to follow for the rest of the framework... Thank you @glennsl !

glennsl:
I believe this is finally done. The last rebase was pretty tricky, but I haven't noticed any actual bugs resulting from it. Just an imaginary one :stuck_out_tongue:

src/Core/Tick.re
bryphe: > In OCaml, raising an exception is how you break from a while loop. I think that was part of the motivation for making them so light-weight and not trace the stack by default.

Oh, didn't know that! I read that they were light-weight, but didn't know the motivation.

> Also, if you haven't you should check out how algebraic effects will work when it lands. I think you're going to love that 🙂 (just don't get too excited, it'll still be a while...)

Definitely! I was thinking of that - would be a very clean way to represent this 😄 Between that and multicore... a lot of exciting stuff on the horizon...

examples/Hello.re
bryphe: Awesome!

test/UI/TestFramework.re
```diff
@@ -1,7 +1,78 @@
...
+open Revery_UI;
+
+open Rely.MatcherTypes;
+
+// Approximate equality for floats
+let (=~.) = (~tolerance=0.0001, a, b) => abs_float(a -. b) < tolerance;
+
+type animationStateExtensions = {toEqual: Animation.state => unit};
+
+let animationStateExtensions = (actual, {createMatcher}) => {
```
bryphe: Cool, neat to have an example of a custom matcher!

bryphe:
Awesome! In addition to the API improvements - I believe this fixes a bug with the 'settled' final state of the animation (which wouldn't always get to `1.0`).

Looks great @glennsl - thanks for this!

bryphe:
BTW, with the approval - do you get the `Squash and Merge` option now? If so - go for it! 😎

glennsl:
I do. Thanks! :smile:

PR #632, opened 2019-11-06
Time for a change (of the representation of time (to use float instead))

glennsl:
This hides and changes the representation of `Time.t` to `float` in order to simplify the implementation and provide more primitives without having to deal with the complexity of the old representation.

It's a breaking change, mostly because the representation was exposed and extensively used, but that's a very small change. I've also taken the opportunity to make a few other breaking changes, however, while I was at it. The list of breaking changes are:

- `Time.Seconds` -> `Time.seconds`
- `Time.MIlliseconds` -> `Time.milliseconds`
- You also can;t pattern match on these constuctors, of course, but I found no instance of that in revery itself.
- `Time.to_float_seconds` and `Time.of_float_seconds` have been removed, since they were deprecated.
- `Time.getTIme` -> `Time.now`
- `Time.increment` has been removed in favor of `Time.(+)`

Some new functions have also been added, most importantly the infix operators `+`, `-`, `*` and `/`, but also `Time.zero` as a shortcut and since 0 doesn't really have a unit. This now enables writing code such as:

```reason
Time.(startTime + delay > seconds(2.));
```

instead of

```reason
Time.ofSeconds(Time.toSeconds(startTime) +. Time.toSeconds(delay) > 2.);
```

Lastly, it's also possible to expose `Time.t` as a `float`, which would allow the convenience of using float values and operators directly. But it's good practice to hide the representation, because that allows changing it if needed (like this does), and to maintain invariants that might later be desired, such as preventing the representation of negative time values for example.


src/Core/Time.re
```diff
@@ -1,43 +1,17 @@
...
-  | Milliseconds(x) => x /. 1000.
-  };
+let ofSeconds = t => t;
+let toSeconds = seconds;
 
-/**
-  [ofSeconds(time)] converts the time [t] to a float value representing the time in seconds.
-*/
-let ofSeconds = of_float_seconds;
+let (+) = (+.);
```
bryphe: Woah, great idea to add these!

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
+type t;
+
+let zero: t;
+let seconds: float => t;
```
bryphe: I like it! This preserves the 'type safety' of knowing we are always using seconds under the hood, and being explicit that we are always getting seconds back when converting to a float - but with much better ergonomics than we had previously. 🎉

glennsl:
I'll just go ahead and merge this then, so I can rebase the animation PR on it.

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
...
+let milliseconds: float => t;
+
+/**
+  [ofSeconds(time)] converts the time [t] to a float value representing the time in seconds.
+*/
+let ofSeconds: float => t;
+
+/**
+  [ofSeconds(seconds)] converts the value seconds to a [t] representing the time in seconds.
```
Et7f3: fix comment.

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
+type t;
+
+let zero: t;
+let seconds: float => t;
+let milliseconds: float => t;
+
+/**
+  [ofSeconds(time)] converts the time [t] to a float value representing the time in seconds.
+*/
+let ofSeconds: float => t;
```
Et7f3: shouldn't be toSeconds ? the output type is the type time.

src/Core/Time.re
```diff
@@ -1,43 +1,17 @@
...
+type t = float;
 
-/**
-@deprecated Replaced by [toSeconds]
-*/
-[@deprecated]
-let of_float_seconds = (v: float) => Seconds(v);
+let zero = 0.;
+let seconds = t => t;
+let milliseconds = t => t /. 1000.;
```
Et7f3: Isn't more precise to have seconds = `milliseconds *. 1000.`

src/Core/Time.rei
glennsl: Done.

src/Core/Time.rei
glennsl: No, the value of `t` is opaque. It represents time abstractly, so "seconds" here refers to the input being in seconds. This function is carried over from the old implementation and is not strictly needed, since it's just an alias of `seconds`.

Actually, I see now that I've aliased it incorrectly in the implementation. This naming confuses me too :laughing: Will fix.

glennsl: Done. They're effectively the same anyway of course, but it's meaning is less confusing this way.

src/Core/Time.re
glennsl: No, I don't that actually matters because float represent `1e-3` as something along the lines of `(1, -3)` I think.

src/Core/Time.rei
Et7f3: fixed in 1e9fa54e38b1db61c1a962f63ceffef498390a7c

src/Core/Time.re
Et7f3: Good point `less chance of unintentional behavioral changes`

src/Core/Time.rei
Et7f3: fixed in 1e9fa54e38b1db61c1a962f63ceffef498390a7c

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
+type t;
```
Et7f3:
```suggestion
type t = pri float;
```
So we can avoid call of toSecond in other file ?

src/UI/AnimationTicker.re
```diff
@@ -6,6 +6,6 @@ open Revery_Core;
 
 let onTick: Event.t(Time.t) = Event.create();
 
-let time = () => Time.getTime();
+let time = () => Time.now();
```
Et7f3:
```suggestion
let time = Time.now;
```
This reduce indirection.

src/Core/Time.rei
glennsl: I briefly brought that up at the end of the PR description:

> it's also possible to expose Time.t as a float, which would allow the convenience of using float values and operators directly. But it's good practice to hide the representation, because that allows changing it if needed (like this does), and to maintain invariants that might later be desired, such as preventing the representation of negative time values for example.

glennsl: It's a trade-off, but we can always add this later without breaking anything. If we add it now and then decide to remove it later, we will break things. So I thought it safest to leave it out for now.

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
...
+  [ofSeconds(seconds)] converts the value seconds to a [t] representing the time in seconds.
+*/
+let toSeconds: t => float;
+
+let (+): (t, t) => t;
+let (-): (t, t) => t;
+let ( * ): (t, float) => t;
+let (/): (t, float) => t;
+
+let show: t => string;
```
Et7f3:
```suggestion
let toString: t => string;
```

src/Core/Time.re
```diff
@@ -1,43 +1,17 @@
...
+let (+) = (+.);
+let (-) = (-.);
+let ( * ) = ( *. );
+let (/) = (/.);
 
-/**
-  [ofSeconds(seconds)] converts the value seconds to a [t] representing the time in seconds.
-*/
-let toSeconds = to_float_seconds;
+let show = t => string_of_float(t) ++ "s";
```
Et7f3:
```suggestion
let toString = t => string_of_float(t) ++ " s";
``` This is misleading. We can think it print.

File: src/UI/AnimationTicker.re (hunk 7 above)
Comment (glennsl) [reply src/UI/AnimationTicker.re]: AnimationTicker is removed in #631, so I don't think this really matters, but nice catch :smile:

File: src/Core/Time.re [hunk 10]
```diff
@@ -1,43 +1,17 @@
...
-    );
-  };
-
-let show = (v: t) => string_of_float(to_float_seconds(v)) ++ "s";
-
-/**
-    [getTime()] returns the current system time [t]
-*/
-let getTime = () => of_float_seconds(Unix.gettimeofday());
+let now = () => Unix.gettimeofday();
```
Et7f3:
```suggestion
let now = Unix.gettimeofday;
```
This reduce indirection. Is this also removed in #631 😂

src/Core/Time.rei
glennsl: This is a convention used by @bryphe throughout the codebase. I get 26 hits with this. So while I would prefer `toString` as well, I don't think we should break the convention solely here. We ought need to apply it across the codebase (and of course get @bryphe's approval first!)

src/Core/Tick.re
```diff
@@ -1,7 +1,7 @@
 module type Clock = {let time: unit => Time.t;};
 
 module DefaultClock = {
-  let time = () => Time.getTime();
+  let time = () => Time.now();
```
Et7f3:
```suggestion
  let time = Time.now;
```
This reduce indirection. Is this also removed in #631 😂

src/Core/Time.re
glennsl: No, this ought to survive :rofl: Let me know when you've finished going through, then I'll batch them up.

src/Core/Time.rei
Et7f3: Ok if you have thought about it. Use it if you need for #631

src/Core/Time.re
glennsl: It's a convention brought over from Haskell, through it's deriving mechanism which has then been imitated by the deriving ppx in OCaml. And Haskell being a pure language wouldn't print of course, so it makes more sense for `show` to mean "convert to string" there.

src/Core/Time.rei
```diff
@@ -0,0 +1,27 @@
+type t;
+
+let zero: t;
+let seconds: float => t;
+let milliseconds: float => t;
+
+/**
+  [ofSeconds(seconds)] converts the value seconds to a [t] representing the time.
```
Et7f3:
```suggestion
  [ofSeconds(seconds)] converts a float value in seconds to a [t] representing the time.
```

examples/Stopwatch.re
```diff
@@ -102,9 +100,7 @@ module Clock = {
                 marginVertical(20),
                 width(200),
               ]
-              text={string_of_float(
-                state.elapsedTime |> Time.to_float_seconds,
-              )}
+              text={string_of_float(state.elapsedTime |> Time.toSeconds)}
```
Et7f3:
```suggestion
              text=Time.show(state.elapsedTime)
```

examples/Stopwatch.re
```diff
@@ -102,9 +100,7 @@ module Clock = {
                 marginVertical(20),
                 width(200),
               ]
-              text={string_of_float(
-                state.elapsedTime |> Time.to_float_seconds,
-              )}
+              text=Time.show(state.elapsedTime)
```
Et7f3:
```suggestion
              text={Time.show(state.elapsedTime)}
```
To make `esy format` happy in the CI.

glennsl:
Checks passed. Thanks for reviewing @Et7f3!

src/Core/Time.rei
Et7f3: @glennsl Also `and to maintain invariants that might later be desired` when prefixed with `pri` the compiler know that float isn't equal to float and the only way to create such value is using functions from Time.re

glennsl: Yeah, that comment was directed towards exposing it completely, so you can manipulate it directly with float operators and such. The downsides of using a private type abbreviation is that you don't have the flexibility of changing the representation without it being a breaking change, and it's also not clear entirely clear what the `float` value represents, whether it's in seconds or milliseconds, or if there's an offset or epoch.

src/Core/Time.re
bryphe: Thanks for the explanation! Good points.

src/Core/Time.rei
bryphe: Thanks @Et7f3 and @glennsl for the discussion around this!

Indeed, I can see where `show` is misleading... For some reason I thought it was 'canonical' OCaml (due to the ppx_deriving_show) - but I can see how it is more ambiguous than `toString` because `show` sounds like it has a side effect.

I'd actually prefer `toString` too (coming from C# / JS worlds!) - so I'd be open to changing and replacing this convention. It sounds like we have 3 votes in favor of switching from `show` to `toString` 😄 Logged an issue for it here: https://github.com/revery-ui/revery/issues/635

PR #633, opened 2019-11-11
Chore: Mark lockfiles as generated to hide them in github PRs

bryphe:


bryphe:
Oh, that's awesome @Et7f3 🎉 Even better - works _everywhere_!

PR #636, opened 2019-11-12
refactor/fix: timing issue with input-cursor

lessp:
I noticed something weird with the input-cursor. As far as I could tell there hasn't been any changes to the logic, but separating it to a custom hook seems to have solved the issue.


src/UI_Components/Input.re
```diff
@@ -5,11 +5,71 @@ open Revery_UI_Primitives;
...
+    | Tick(Time.t);
+
+  let use = (~interval, ~isFocused) => {
+    let%hook (state, dispatch) =
+      Hooks.reducer(
+        ~initialState={time: Time.zero, toggle: Off}, (action, state) => {
+        switch (action) {
+        | Reset => {toggle: On, time: Time.zero}
+        | Tick(increasedTime) =>
+          let newTime = Time.(state.time + increasedTime);
```
glennsl: I still don't understand what the reason for this is, rather than having the tick run at 500ms?

src/UI_Components/Input.re
```diff
@@ -5,11 +5,71 @@ open Revery_UI_Primitives;
...
+
+          /* if newTime is above the interval a `Tick` has passed */
+          newTime >= interval
+            ? {
+              toggle:
+                switch (state.toggle) {
+                | On => Off
+                | Off => On
+                },
```
glennsl: Wouldn't it be simpler to have `state.toggle` just be a `bool`? Then this would be just `toggle: !state.toggle`.

src/UI_Components/Input.re
```diff
@@ -5,11 +5,71 @@ open Revery_UI_Primitives;
...
+          Some(clear);
+        },
+      );
+
+    let cursorOpacity =
+      switch (isFocused, state.toggle) {
+      | (true, Off)
+      | (false, _) => 0.0
+      | (true, On) => 1.0
+      };
```
glennsl: Similarly, this could be
```reason
let cursorOpacity = isFocused && state.toggle ? 1.0 : 0.0;
```

And perhaps `state.toggle` would be better name something like `state.isOn`

src/UI_Components/Input.re
glennsl: Oh, maybe it's because the tick runs constantly whether focused or not, and to get immediate feedback when being focused. In that case, it might be a good idea to implement a pause and reset mechanism similar to what the `timer` hook has.

Even better, implement that in the `tick` hook and use that, so everyone can benefit from it!

lessp: Yes, I remembered I'd been down the route of simply having the tick run at e.g. `500ms`, but hit your second comment! 😄 Can look into implementing it in the `Tick`-hook!

src/UI_Components/Input.re
lessp: Yes, that makes sense with the change in naming, thanks!

src/UI_Components/Input.re
lessp: Fixed in: https://github.com/revery-ui/revery/pull/636/commits/0bfee3a2085b524f6b28904bf376b97ec65c820f

src/UI_Components/Input.re
lessp: Fixed in: https://github.com/revery-ui/revery/pull/636/commits/0bfee3a2085b524f6b28904bf376b97ec65c820f

src/UI_Components/Input.re
lessp: I think for now I'll keep it to fixing the more immediate issue with the `Input`-component. Might be an idea to create a ticket to extend the `tick`-hook as well, though!

glennsl: Yeah, that sounds reasonable. The logic is much better now than it was at least!

lessp: Agreed, thanks a lot for the help and for reviewing!

PR #637, opened 2019-11-12
Time ergonomics improvements

glennsl:
- `Time.seconds` and `Time.milliseconds` converted to `int`
- `Time.ms` added as alias of `Time.milliseconds` (But I stopped before adding `Time.s`, because that would be ridiculous)
- Renamed `Time.ofSeconds` and `Time.toSeconds` to `Time.ofFloatSeconds` and `Time.toFloatSeconds`

There are still times where specifying seconds using a `float` would be nicer. `Time.seconds(1.5)` is nicer than `Time.ms(1500)`, but not enough to suffer through all the `.`s in every other scenario. It's also still possible to specify float values if needed using `Time.ofFloatSeconds`.


bryphe:
Cool! Seems reasonable to me. 

> Time.seconds(1.5) is nicer than Time.ms(1500), but not enough to suffer through all the .s in every other scenario. It's also still possible to specify float values if needed using Time.ofFloatSeconds.

Agreed. Will be interesting to see how this works out! Thanks for thinking through these API ergonomics, @glennsl !

PR #641, opened 2019-11-13
4.7.0 Compatibility

glennsl:
Recent commits have begun using some functions added in 4.8.0. This adds polyfill for those, and pins the ocaml devDependency to 4.7.x to make sure we don't accidentally break 4.7.0 compatibility in the future.


Et7f3:
For reference. Maybe we could do like stdlib-shim https://github.com/ocaml/stdlib-shims/blob/master/src/dune

PR #642, opened 2019-11-13
[WIP] feat: support remote images

### LINKED ISSUE #537: Load image from absolute path/url
Trying to display images from URL I found it's currently not supported. I decided to download images to some tmp directory and use an absolute path as an Image src. And that's also not supported right now: https://github.com/revery-ui/revery/blob/aa2051c7c6e562b0b13acd43ff4724e02067097f/src/Draw/ImageRenderer.re#L25

The best option would be to have:
```reason
type src = Url(string) | Relative(string) | Absolute(string)
```
which would allow loading images from `/assets`, any directory and URL.

I'm willing to help with that feature but I don't know what's the best approach to download image and use it as a texture in ImageRenderer.

lessp:
While experimenting with Revery I'm using a separate fork where I've, kind of naively, added support for remote-images. I'm opening so we can discuss and take the implementation further.

Questions:

- What to do about errors?

Would close #537


bryphe:
Wow, this would be so nice to have 😍 Thanks @lessp for looking at this

lessp:
I added `Revery_Fetch` which currently just wraps `fetch-native-lwt` in https://github.com/lessp/reason-fetch/. 
I don't know if this is something we want to add now, it might be something for a later discussion 🤷‍♂ 

For Windows-support the main-blocker is:

https://github.com/ocsigen/lwt/issues/745

which will solve:

https://github.com/lessp/reason-fetch/pull/18
https://github.com/lessp/reason-fetch/issues/14

And we'd also obviously want to handle the pin to anmonteiro's in order for it not to leak the resolutions.

https://github.com/revery-ui/revery/pull/642/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R56-R58

-----

Also, I'm not sure how to make this work in JS, but feels like it shouldn't be too hard?

lessp:
Closing in favour of #796

Issue #472, opened 2019-04-26
position `Absolute disable Input component


Et7f3:
- Operating System: Windows
- Revery Version: github:revery-ui/revery#a3cab38a79a737a724d25359be5a9ff4a947a468
- OCaml version: version 4.07.1
- Native, Bytecode, or JS build: Native

- repro:
`let init app =
  let win = App.createWindow app "OCalc" in
  Input.createElement ~style:Style.[top (100); left (300); position `Absolute] ~children:[] () |>
  UI.start win |> ignore

let _ = App.start init
`
- Actual Result:
Can't edit value.
- Expected Result:
We can edit value.

- Additional Information:
None

bryphe:
Was able to reproduce this.

It seems the root issue is that, when we apply a `style` to the `<Input />` control, it actually applies it to a _nested_ `<View />`. This means that the internal view as `position(Absolute)`, and is taken out of the flow - so the input has no height. Therefore, we never hit the click handler, because the 0-height bounding box never registers a hit with the mouse.

WhoAteDaCake:
What would be a solution to prevent this @bryphe ? I would love to help

parkerziegler:
👋 Hey folks, @bmathews and I started to take a look at this yesterday. @bryphe I think your suspicions are correct. When `position(Absolute)` is applied, the nested `View` (`aquamarine` border in this GIF) is taken out of the flow, so the wrapping `Clickable` (`red` border in this GIF) has no height and thus can't be clicked. 

![191005_revery_absolute_input](https://user-images.githubusercontent.com/19421190/66260128-351d2780-e76f-11e9-95e1-560e387b49a7.gif)

What's weird to me is that if we move the `Absolute` positioning from the nested `View` to the `Clickable` (just for the `position(Absolute)` case), the `Clickable` is no longer actually clickable. In the GIF below, the `Clickable` has `position(Absolute))`, and the nested `View` does not. However, the `Clickable` (red) can't seem to be clicked 🤔 

![191005_not_clickable](https://user-images.githubusercontent.com/19421190/66260257-14ee6800-e771-11e9-8e09-b0eb531c51e2.gif)

Here's the diff for that second GIF – the positioning seems correct, I'm not sure why the `Clickable` isn't receiving the click event.

```reason
<Clickable
  onFocus={() => {
    dispatch(ResetCursorTimer);
    dispatch(SetFocus(true));
    Sdl2.TextInput.start();
  }}
  onBlur={() => {
[...]

PR #644, opened 2019-11-15
Input component improvements

### LINKED ISSUE #203: UI - Input Component
I noticed there wasn't an issue to track against adding an `Input` component to revery (I started having a look into this), not sure if we have all the pieces we need in place for at least a minimal implementation.

If there are any specific behaviors to keep in mind or anyone has any suggestions it would be nice to collect them here

### Outstanding work

- [x] Controlled Inputs - relates to #305
- [x] Keyboard navigation
- [ ] Cursor/Click editing
- [x] Move focus management to higher level - relates to #290 
- [ ] Text wrapping - relates to #300

glennsl:
This greatly simplifies the internal logic of the Input component, and at the same time exposes `cursorPosition` for control. The only breaking change is that `onChange` now has `cursorPosition` as an additional argument. `value` and `cursorPositon` can otherwise be controlled individually.

This refactoring was motivated by an experienced bug (but unreported I think) where mashing the keyboard caused the cursor to "lag" behind, getting positioned behind the last character and inputting characters in the wrong order. A fast typer on a slow machine might have been able to reproduce this in normal conditions, and given the complexity of the logic I would be surprised if this was the only bug.

I understand the motivation for the current implementation was to imitate React's input component as closely as possible (ref. #406), and while the motivation is understandable and the effort commendable, I don't think it's a good idea to go to such lengths that we reproduce the flawed design of the DOM. Specifically, because the DOM does not allow control over cursor position, React has no choice but to deal with that as best as it can. But since we do have control over it, we can yield control to the consumer when needed instead of trying to predict their intention, while at the same time providing a reasonably simple default implementation that will work in most use cases.

I've continued by adding a few more features (see below), and believe this now fixes #203, although this implements scrolling instead of wrapping.

#### Todo (but can be merged at any time if desired)
- [x] Click to position caret
- [x] Scroll instead of wrap overflowing text
-  ~~Selection~~ (edit: should be a separate PR)

#### Known issues
- [x] Cursor "jumps" when going "out of bounds"
- [x] Placeholder text renders over cursor


src/UI_Components/Input.re
```diff
@@ -266,6 +268,52 @@ let%component make =
...
+          if (width > offset) {
+            let isCurrentNearest = width - offset < offset - last;
+            isCurrentNearest ? i : i - 1;
+          } else {
+            loop(i + 1, width);
+          };
+        };
+
+      loop(1, 0);
+    };
```
glennsl: Perhaps this should be moved into `Text`

src/UI_Components/Input.re
```diff
@@ -266,6 +268,52 @@ let%component make =
...
     Selector.select(style, FontFamily, "Roboto-Regular.ttf");
 
+  let handleClick = (event: NodeEvents.mouseButtonEventParams) => {
+    let rec offsetLeft = node => {
+      let Dimensions.{left, _} = node#measurements();
+      switch (node#getParent()) {
+      | Some(parent) => left + offsetLeft(parent)
+      | None => left
+      };
+    };
```
glennsl: This probably belongs in `Node` in one form or other.

src/UI_Components/Input.re
```diff
@@ -270,33 +270,25 @@ let%component make =
...
+        // out of view to the left, so align with left edge
+        scrollOffset := cursorOffset;
+      } else if (cursorOffset - scrollOffset^ > container.width) {
+        // out of view to the right, so align with right edge
+        scrollOffset := cursorOffset - container.width;
+      }
+
+    | None => ()
+    };
+  };
```
glennsl: This is a bit of an odd pattern, in connection with the hook at the top:

```reason
let%hook (scrollOffset, _setScrollOffset) = Hooks.state(ref(0));
```

We need the computed value immediately, and _also_ need to keep the previous value to compute the next. This could have been done with an ordinary binding for the immediately computed value and a separate `ref` hook to keep it for the next render. But that's pretty awkward since the computed value has to be "set" twice:

```reason
let scrollOffset = cursorOffset - container.width;
setPreviousScrollOffset(scrollOffset);
scrollOffset;
```

I've therefore opted for using a state hook with an ordinary `ref` cell, with a setter that's never used. While also wondering if this is essentially what the `ref` hook ought to be.

glennsl:
I'm not sure how to fix the last remaining issue, the placeholder text being rendered over the cursor. Also not sure it's worth it if the component and styling model is going to change anyway.

From what I can gather, it worked before even with the cursor being positioned "before" the text, because the cursor was both absolutely positioned AND a sibling of the text. Now, on the other hand, while the cursor is still absolutely positioned it is no longer a sibling, because the text needs to be contained in a clipping container and the cursor needs to not be clipped. I think z-ordering would have been the fix for this if it was the DOM, but I don't think we have that yet.

Also, CI failure seems to be a fluke.

src/UI_Components/Input.re
glennsl: Done in https://github.com/revery-ui/revery/pull/644/commits/97732eaa251864b6d2a77efe2163617bb48cdf06

src/UI_Components/Input.re
glennsl: Done in https://github.com/revery-ui/revery/pull/644/commits/97732eaa251864b6d2a77efe2163617bb48cdf06

src/UI_Components/Input.re
```diff
@@ -117,292 +132,279 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
-      character: event.text,
-      altKey: false,
-      ctrlKey: false,
-      shiftKey: false,
-      superKey: false,
+    switch (Option.bind(textRef, r => r#getParent())) {
+    | Some(containerNode) =>
+      let container: Dimensions.t = containerNode#measurements();
+
+      if (cursorOffset < scrollOffset^) {
```
bryphe: Wow! So nice to have this case working now 👍

src/UI_Components/Input.re
```diff
@@ -117,292 +132,279 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
-    switch (valueAsProp) {
-    | Some(v) => v
-    | None => state.internalValue
-    };
+  let value = Option.value(value, ~default=state.value);
+  let showPlaceholder = value == "";
+  let cursorPosition =
+    Option.value(cursorPosition, ~default=state.cursorPosition);
+
+  module Styles = {
```
bryphe: I like this convention for demarcating the style-related pieces

src/UI_Components/Input.re
```diff
@@ -117,292 +132,279 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
-    let createChangeEvent = inputString => {
-      value: inputString,
-      character,
-      keycode: event.keycode,
-      altKey: Key.Keymod.isAltDown(event.keymod),
-      ctrlKey: Key.Keymod.isControlDown(event.keymod),
-      shiftKey: Key.Keymod.isShiftDown(event.keymod),
-      superKey: Key.Keymod.isGuiDown(event.keymod),
-    };
+  // TODO:This ought to be in the reducer, but since reducer calls are deferred
```
bryphe: Thanks for calling this out!

bryphe:
> This refactoring was motivated by an experienced bug (but unreported I think) where mashing the keyboard caused the cursor to "lag" behind, getting positioned behind the last character and inputting characters in the wrong order. A fast typer on a slow machine might have been able to reproduce this in normal conditions, and given the complexity of the logic I would be surprised if this was the only bug.

I really like the way you simplified the logic here (and I'm no longer able to repro the fast-typing issue!)

> I don't think it's a good idea to go to such lengths that we reproduce the flawed design of the DOM. Specifically, because the DOM does not allow control over cursor position, React has no choice but to deal with that as best as it can. But since we do have control over it, we can yield control to the consumer when needed instead of trying to predict their intention, while at the same time providing a reasonably simple default implementation that will work in most use cases.

Good point. We have the opportunity with Revery to get control over these aspects w/o inheriting the quirks or flaws of the DOM, so we should take advantage of that.

> I'm not sure how to fix the last remaining issue, the placeholder text being rendered over the cursor. Also not sure it's worth it if the component and styling model is going to change anyway.

I'm OK with the behavior as it is today. We can revisit the placeholder text in another issue, or once we have more control around layout / style to implement it cleanly. 

Being able to type reliably is the most important aspect for an input control.

glennsl:
Oops. Good catch @Et7f3! Fixed in https://github.com/revery-ui/revery/pull/644/commits/e493a4a21c142546510cc106653c868944dd2686

src/UI_Components/Input.re
```diff
@@ -117,292 +132,280 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
+    onAnyClick=handleClick
     onKeyDown=handleKeyDown
     onTextInput=handleTextInput>
-    <View style=viewStyles>
-      cursor
-      {hasPlaceholder ? placeholderText : inputText}
+    <View style=Styles.box>
+      <View style=Styles.marginContainer>
+        <cursor />
+        <View style=Styles.textContainer> <text /> </View>
```
Et7f3:
```suggestion
        <View style=Styles.textContainer> <text /> </View>
        <cursor />
```

src/UI_Components/Input.re
```diff
@@ -117,292 +132,280 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
     componentRef={autofocus ? Focus.focus : ignore}
+    onAnyClick=handleClick
     onKeyDown=handleKeyDown
     onTextInput=handleTextInput>
-    <View style=viewStyles>
-      cursor
-      {hasPlaceholder ? placeholderText : inputText}
+    <View style=Styles.box>
+      <View style=Styles.marginContainer>
+        <cursor />
```
Et7f3:
```suggestion
```

src/UI_Components/Input.re
Et7f3:
```suggestion
        <View style=Styles.textContainer> <text /> </View>
        <cursor />
```

src/UI_Components/Input.re
Et7f3:
```suggestion
```

src/UI_Components/Input.re
glennsl: Hmm, for some reason I think I've assumed that this would put the cursor after the text. I'm pretty sure it would in HTML when `left: 0` isn't specified.

But it seems to work here so... Yay! Nice one @Et7f3

Et7f3:
Can't `cursorPosition` be in changedEvent ? Most user have `{value; _}` so it is a non-breaking change.

glennsl:
`changedEvent` is no longer used because it assumed the change originates from a key event, which it no longer does. It also didn't seem very useful. If you want to react to key strokes, use `onKeyDown` instead.

I see I've forgotten to remove the type though, which I should.

Et7f3:
Many code have handler of the form `({value; _})`. Can we have a InputEvent that have value and cursorPosition so we can add other field without break again.

src/UI_Components/Input.re
```diff
@@ -117,292 +122,280 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
-    };
+  let value = Option.value(value, ~default=state.value);
+  let showPlaceholder = value == "";
+  let cursorPosition =
+    Option.value(cursorPosition, ~default=state.cursorPosition)
+    |> min(String.length(value));
+
+  module Styles = {
+    open Style;
+    include Styles;
```
Et7f3:
```suggestion
```
include for only one value 👀.

src/UI_Components/Input.re
```diff
@@ -117,292 +122,280 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
+
+    let _all =
+      merge(
+        ~source=[
+          flexDirection(`Row),
+          alignItems(`Center),
+          justifyContent(`FlexStart),
+          overflow(`Hidden),
+          cursor(MouseCursors.text),
+          ...default,
```
Et7f3:
```suggestion
          ...Styles.default,
```

src/UI_Components/Input.re
```diff
@@ -117,292 +122,280 @@ let removeCharacterAfter = (word, cursorPosition) => {
...
-    | Some(v) => v
-    | None => state.internalValue
-    };
+  let value = Option.value(value, ~default=state.value);
+  let showPlaceholder = value == "";
+  let cursorPosition =
+    Option.value(cursorPosition, ~default=state.cursorPosition)
+    |> min(String.length(value));
+
+  module Styles = {
```
Et7f3: Modules declared in function I'm not sure it is a good idea.

glennsl:
> Can we have a InputEvent that have value and cursorPosition so we can add other field without break again.

We could, but it seems unlikely that there should be more state to track. If there is, that seems to imply that the model has changed so much that it probably _ought_ to break existing code. Breaking changes aren't inherently bad, and I usually prefer to err on the side of changes being breaking. That's why we have exhaustiveness checking, warnings on missing record fields in patterns etc. Records also aren't very ergonomic, as you need to bring the fields into scope to access them. There are of course plenty of reasons to use them too, but I don't think "just-in-case" is a good reason.

Et7f3:
#472 is still present. Do you consider to fix in this PR or let this issue to be fixed by another PR.

Et7f3:
The record came in scope when we include the Input component so it isn't a ergonomic problem.
> We could, but it seems unlikely that there should be more state to track
Ok.

src/UI_Components/Input.re
glennsl: There are three values, although only one is actually used, but the _intention_ is to include everything, so that's what it does. What's the problem with doing so, do you think?

Et7f3: Including without making interfaces include all values and rexport them all but it is not a problem for 3 values I have read `Style`, the module with near 100 values. Sorry.

src/UI_Components/Input.re
glennsl: I don't know that modules are heavier then stand-alone functions as long as it's just used as a namespace. Either way, this pattern certainly isn't ideal. The reason it's here is because `_all` depends on `style`, which is a prop, and is also used by other top=level definitions. But on second thought I think it's better to just parameterize it and pass it down, so I think I'll change it to do that. It'll be much less messy too, that way.

Et7f3: Ah almost all value depend on prop style. So it is fine, I think.

glennsl:
> #472 is still present. Do you consider to fix in this PR or let this issue to be fixed by another PR.

I think the issue is orthogonal to this PR, so I don;t see why it should block it.

> The record came in scope when we include the Input component so it isn't a ergonomic problem.

It is if you use named functions, which I think is a good pattern. You could of course then argue that it isn't particularly useful for this particular event since it'll usually just be dispatched right away, but it still sets a precedent.

I think the same rules should be applied here as we do generally. And if you want other rules to be applied I think you need to argue for it more generally. If the argument is, as I understand it is, that we should use a record just in case we need to add more fields later, I think that is not something we should do in general, and therefore I don't think we should do so here either. That's my reasoning at least.

Et7f3:
Revamp are useful to reduce bug. You have fixed some issues when revamp the animation. That why I quoted #472.

glennsl:
Yeah it definitely could have, so good idea to check!

PR #645, opened 2019-11-18
Add a titlebar style option

zbaylin:
This ties into the newly added `Sdl2.Window.setMacTitlebarTransparent`, which should eventually be expanded to other OSs.

I created a new file, `WindowStyles.re` to house all the styling of windows we may have in the future.

### DIFF
--- a/src/Core/WindowCreateOptions.re
+++ b/src/Core/WindowCreateOptions.re
@@ -17,6 +17,11 @@ type t = {
     */
   decorated: bool,
   /**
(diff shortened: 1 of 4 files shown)


src/Core/WindowCreateOptions.re
```diff
@@ -16,6 +16,11 @@ type t = {
     If [decorated] is true, the window will be framed with the OS-specific adorners.
     */
   decorated: bool,
+  /**
```
bryphe: Thanks for adding the doc comments 👍

PR #648, opened 2019-11-19
Fix #647, timer set on mount not being cleared 

### LINKED ISSUE #647: Performance: Example app no longer goes to 'idle'
I was investigating an issue where Onivim 2 seems to continuously re-render (ie, it never goes into an 'idle' state) - this is problematic because when a user looks at the CPU usage, it'll seem high (since we're in the render-loop). 

I checked out the Revery example app, and noticed that it was also failing to go into 'Idle'. It's expected that the Hello/Animation example doesn't 'Idle', because the animations will always trigger a re-render. However, navigating away from the example app _should_ clear the timers, and allow Revery to go into an idle state (we should get that log message saying "Downshifting into Idle" after a few seconds).

I bisected and saw that it seems like we stopped get the 'idle state' with the animation / timing changes in https://github.com/revery-ui/revery/commit/e523fb6e52b21897f13fbced366bfe1584f78a69

I debugged a little bit, and the animations for the `<Logo />` component seem to be the culprit - specifically, there seem to be two tickers created - one via the `OnMount` effect, and one via the `condition` effect. The ticker created on the `OnMount` effect is cleared when the `condition effect` kicks in.

Upon navigating away, though, it seems all of these tickers are not cleared. I see these tickers get created:
```
Ticker - interval: 0
Ticker - interval: 1
Ticker - interval: 2
```

But I only see clears for 0 & 1 upon navigating away:
```
Ticker - clear: 0
Ticker - clear: 1
```

I'd expect that, after navigating away from the animation example - there'd be no active tickers. @glennsl , since you looked at this recently - do you have any ideas on why we are getting a ticker that isn't cleared when navigating away from the Animation example? 

Let me know if you need any more details from me!

glennsl:
Fixes #647 

Ach, foiled by the naming of `Ref.ref` again!

So what happened was that the `mountCleanup` closure was returned as the dispose function on mount, but that closure would always refer to the initial value of `maybeDispose`, `None`, because it's not an actual `ref`.

Also, it still worked if the condition changes because that called `mountCleanup` directly instead of using a stored closure! So it seemed to work when you paused the logo animation, and it ALSO worked on unmount AFTER having paused and resumed, because that stored the `dispose` function directly, instead of a closure referring to a stale `maybeDispose`.

The solution is to put an ordinary `ref` in a `state` (or `ref`) hook, and to use ti as an ordinary `ref` instead of using the hook setter.

### DIFF
--- a/src/UI_Hooks/Revery_UI_Hooks.re
+++ b/src/UI_Hooks/Revery_UI_Hooks.re
@@ -16,7 +16,7 @@ let time = (~tickRate=Time.zero, ()) => {
 // TODO: Workaround for https://github.com/briskml/brisk-reconciler/issues/27
 // Remove when fixed
 let mountIfEffect = (condition, handler) => {
-  let%hook (maybeDispose, _setDispose) = state(Pervasives.ref(None));
+  let%hook (maybeDispose, _setDispose) = state(Stdlib.ref(None));
   let mountCleanup = () =>
     switch (maybeDispose^) {
     | Some(dispose) =>
--- a/src/UI_Hooks/Revery_UI_Hooks.re
+++ b/src/UI_Hooks/Revery_UI_Hooks.re
@@ -16,12 +16,12 @@ let time = (~tickRate=Time.zero, ()) => {
 // TODO: Workaround for https://github.com/briskml/brisk-reconciler/issues/27
 // Remove when fixed
 let mountIfEffect = (condition, handler) => {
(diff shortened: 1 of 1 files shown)


bryphe:
> So what happened was that the mountCleanup closure was returned as the dispose function on mount, but that closure would always refer to the initial value of maybeDispose, None, because it's not an actual ref.

Great catch @glennsl !

PR #651, opened 2019-11-21
Bugfix: Ensure timer stops when the bounce is 'idle'

bryphe:
__Issue:__ This came about from investigating onivim/oni2#946 - Revery would never shift into an 'idle' state, even though nothing was happening with the editor.

I started 'bisecting' the UI - I eventually found if I removed the file explorer, the issue did not repro. I dug further and found that it was only the `<ScrollView />` that caused the issue - if I removed the `<ScrollView />`, the app would downshift into idle.

The `<ScrollView />` chooses an animation strategy based on the animation state (a really neat, functional implementation!):
```
  let%hook (actualScrollTop, _bounceAnimationState, resetBouncingAnimation) =
    switch (bouncingState) {
    | Idle => Hooks.animation(Animation.const(actualScrollTop))

    | Bouncing(force) =>
      Hooks.animation(
        bounceAnimation(~origin=actualScrollTop, ~force), ~onComplete=() =>
        setBouncingState(_ => Idle)
      )
    };
```

I found, though, in the `Idle` state - the timer would continually fire - even though it shouldn't, as the animation is over instantly.

I thought that, perhaps this was a bug with a constant-animation in general - but mysteriously, this doesn't reproduce in the Revery example app (the ScrollView example). I'm still not exactly sure what is different between the `<ScrollView />` used in the file explorer, and the `<ScrollView />` used in the example app that would trigger these differences. Perhaps the way we render the UI triggers a different disposal route

__Fix:__ A simple fix that seems to address this is to add `~active=false` to the ScrollView's animation in the idle state. This fixes onivim/oni2#946 - but not sure if it is the best fix.

### DIFF
--- a/src/UI_Components/ScrollView.re
+++ b/src/UI_Components/ScrollView.re
@@ -59,7 +59,9 @@ let%component make =
 
   let%hook (actualScrollTop, _bounceAnimationState, resetBouncingAnimation) =
     switch (bouncingState) {
-    | Idle => Hooks.animation(Animation.const(actualScrollTop))
+    | Idle =>
+      // TODO: Why isn't Animation.const always sufficient to stop the timer?
+      Hooks.animation(~active=false, Animation.const(actualScrollTop))
 
     | Bouncing(force) =>
       Hooks.animation(


bryphe:
Thanks for the details, @glennsl ! Great summary of the timeline / behavior.

> Unless I'm missing something here, what difference between oni and revery could possibly cause that to happen? One interesting experiment could be to try using a reducer instead, but I have no explanation for believing that would actually do anything.

The only difference I can imagine is a difference in the timeline of how component updates are triggered. In the Revery example app, the _only_ updates come via the hooks in the animation.

In Onivim, however, we have a top-level `update` that happens as well - when the `Store` changes state: https://github.com/onivim/oni2/blob/7902cf064d5aa189e12faea91334dd9f40aa1a67/src/bin_editor/Oni2_editor.re#L63

My suspicion is that, perhaps when these calls are interleaved (an update triggered by a hook, and a top-level update triggered by a state change), it could hit a race condition like you mentioned.

The way we handle / process updates might not be perfect - this is where we 'reconcile' and run side effects:
https://github.com/revery-ui/revery/blob/c9714c0801834a7bd33636d1d3df9f006850d627/src/UI/Container.re#L30

A couple of thoughts:
- Perhaps we should 'drain' this - when a hook changes state, like `setState` immediately, that would be deferred until the next time we render (it'd trigger the `onStale` callback, and we'd know to render - but there might be an `update` that sneaks in in between).

> I'm wondering if it's worth pursuing this further given this fix and the inability to reproduce it in Revery though. If it happens again it would hopefully provide some more information to help pin it down. I think we should also make sure to revisit this fix a bit later to see if the issue has somehow fixed itself so we can remove it.

Definitely! I think it is something we should keep an eye out for - my worry is timers not being disposed / memory leaks w/ race conditions of this class. It'd be nice to know exactly what is going on! I'll bring this in for now - but we can keep an eye out

Issue #444, opened 2019-04-08
UI - Dialog: request feature file chooser


Et7f3:
Hello,

I have noticed that Dialog module is not available on api doc: https://www.outrunlabs.com/revery/api/revery/revery/

but it is include in Revery modules.

https://github.com/revery-ui/revery/pull/276
https://github.com/revery-ui/revery/issues/249

bryphe:
Ah, ya, we should include the `Revery.Native` module in the documentation (it's pretty out-of-date - working on some improvements here: #446 ).

We currently just have the ability to show a native platform `alert` - but it'd be great to expand the functionality in `Revery_Native` to include showing a file picker via the native platform APIs 👍

PR #656, opened 2019-11-26
[Native] Implement the macOS/Cocoa and Gtk File Dialogs

zbaylin:
**EDIT**: This is now fixed, and there is more information below.

This provides bindings to `NSOpenPanel`. This is in response to #444 by @Et7f3.

I'm _so_ close, but for some reason, I keep getting strange memory errors on any operations on the string(s) returned.

For example, the code
```reason
Console.log("WORKS:");
Console.log(x);
Console.log("CRASHES:");
Console.log("" ++ x);
```
results in
error

Hopefully some fresh eyes can help with this.

@Et7f3, any help would be greatly appreciated!

### DIFF
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -2,5 +2,6 @@
 
 [@noalloc]
 external openFiles:
-  (~startDirectory: string=?, ~fileTypes: list(string)=?, string) => unit =
+  (~startDirectory: string=?, ~fileTypes: list(string)=?, string) =>
+  option(array(string)) =
   "revery_alertOpenFiles";
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -1,7 +1,12 @@
 [@noalloc] external alertSupported: unit => bool = "revery_alertSupported";
 
-[@noalloc]
 external openFiles:
-  (~startDirectory: string=?, ~fileTypes: list(string)=?, string) =>
+  (
+    ~startDirectory: string=?,
+    ~fileTypes: array(string)=?,
+    ~allowMultipleFiles: bool=?,
+    ~title: string=?,
+    unit
+  ) =>
   option(array(string)) =
   "revery_alertOpenFiles";
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -4,13 +4,34 @@ external _openFiles:
   (
     ~startDirectory: option(string),
     ~fileTypes: option(array(string)),
-    ~allowMultipleFiles: bool,
+    ~allowMultiple: bool,
+    ~canChooseFiles: bool,
+    ~canChooseDirectories: bool,
+    ~buttonText: option(string),
     ~title: option(string),
     unit
   ) =>
   option(array(string)) =
-  "revery_alertOpenFiles";
+  "revery_alertOpenFiles_bytecode" "revery_alertOpenFiles_native";
 
 let openFiles =
-    (~startDirectory=?, ~fileTypes=?, ~allowMultipleFiles=false, ~title=?, ()) =>
-  _openFiles(~startDirectory, ~fileTypes, ~allowMultipleFiles, ~title, ());
+    (
+      ~startDirectory=?,
+      ~fileTypes=?,
+      ~allowMultiple=false,
+      ~canChooseFiles=true,
+      ~canChooseDirectories=false,
+      ~buttonText=?,
+      ~title=?,
+      (),
+    ) =>
+  _openFiles(
+    ~startDirectory,
+    ~fileTypes,
+    ~allowMultiple,
+    ~canChooseFiles,
+    ~canChooseDirectories,
+    ~buttonText,
+    ~title,
+    (),
+  );
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -1,5 +1,6 @@
-//open Reglfw.Glfw;
-
 [@noalloc] external alertSupported: unit => bool = "revery_alertSupported";
 
-// [@noalloc] external alert: (NativeWindow.t, string) => unit = "revery_alert";
+[@noalloc]
+external openFiles:
+  (~startDirectory: string=?, ~fileTypes: list(string)=?, string) => unit =
+  "revery_alertOpenFiles";
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -1,12 +1,16 @@
 [@noalloc] external alertSupported: unit => bool = "revery_alertSupported";
 
-external openFiles:
+external _openFiles:
   (
-    ~startDirectory: string=?,
-    ~fileTypes: array(string)=?,
-    ~allowMultipleFiles: bool=?,
-    ~title: string=?,
+    ~startDirectory: option(string),
+    ~fileTypes: option(array(string)),
+    ~allowMultipleFiles: bool,
+    ~title: option(string),
     unit
   ) =>
   option(array(string)) =
   "revery_alertOpenFiles";
+
+let openFiles =
+    (~startDirectory=?, ~fileTypes=?, ~allowMultipleFiles=false, ~title=?, ()) =>
+  _openFiles(~startDirectory, ~fileTypes, ~allowMultipleFiles, ~title, ());
--- a/src/Native/Dialog.re
+++ b/src/Native/Dialog.re
@@ -7,6 +7,7 @@ external _openFiles:
     ~allowMultiple: bool,
     ~canChooseFiles: bool,
     ~canChooseDirectories: bool,
+    ~showHidden: bool,
     ~buttonText: option(string),
     ~title: option(string),
     unit
@@ -21,6 +22,7 @@ let openFiles =
       ~allowMultiple=false,
       ~canChooseFiles=true,
       ~canChooseDirectories=false,
+      ~showHidden=false,
       ~buttonText=?,
       ~title=?,
       (),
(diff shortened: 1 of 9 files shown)


Et7f3:
Can you select file extention ? like "*.txt". On windows we can. It may be useful to add function to generate the string if it haven't the same syntax.

zbaylin:
yes you can -- I haven't pushed it yet but I will when I get back home.

src/Native/Dialog.re
```diff
@@ -1,5 +1,7 @@
-//open Reglfw.Glfw;
-
 [@noalloc] external alertSupported: unit => bool = "revery_alertSupported";
 
-// [@noalloc] external alert: (NativeWindow.t, string) => unit = "revery_alert";
+[@noalloc]
+external openFiles:
+  (~startDirectory: string=?, ~fileTypes: list(string)=?, string) =>
```
Et7f3: Could the last arg be `option(string)` because it is optionnal on windows.

src/Native/dialog.cpp
```diff
@@ -42,4 +55,43 @@ CAMLprim value revery_alert(value vWindow, value vMessage) {
...
+  }
+  char** fileList;
+#ifdef __APPLE__
+  fileList = revery_open_files_cocoa(startDirectory, NULL, title);
+#endif
+
+  if (fileList) {
+    CAMLlocal2(camlArr, mlData);
+
+    int len = sizeof(fileList) / sizeof(char*);
```
Et7f3:
```suggestion
    int len = sizeof(fileList) / sizeof(*fileList);
``` less error prone. And change when you change `fileList` type

File: src/Native/dialog.cpp [hunk 3]
```diff
@@ -42,4 +55,43 @@ CAMLprim value revery_alert(value vWindow, value vMessage) {
...
+      int strl = strlen(fileList[i]);
+
+      mlData = caml_alloc_string(strl);
+
+      memcpy(String_val(mlData), fileList[i], strl);
+
+      Store_field(camlArr, i, mlData);
+    }
+
+    CAMLreturn(Val_some(camlArr));
```
Et7f3: Nothing to free ? `fileList` for instance ?

zbaylin: Fixed in latest commit.

src/Native/dialog.cpp
zbaylin: I changed how this worked, so it isn't needed anymore, but thank you!

src/Native/Dialog.re
zbaylin: Oops! It's optional on macOS too. My bad -- fixed in latest commit.

zbaylin:
I fixed the alloc error. It was a result of me mistakenly adding `[@noalloc]` before my external declaration. Now the dialog is fully functional. I will add more functionality to it (changing the "OK" button text, allowing opening directories, disallowing opening files, etc.)

src/Native/dialog.cpp
```diff
@@ -13,33 +14,123 @@
...
+    int canChooseFiles = Bool_val(vCanChooseFiles);
+    int canChooseDirectories = Bool_val(vCanChooseDirectories);
+
+    if (vFileTypes != Val_none) {
+        CAMLlocal1(camlArr);
+        camlArr = Some_val(vFileTypes);
+        fileTypesSize = Wosize_val(camlArr);
+
+        // Allocate space for an array
+        fileTypes = (char **)malloc(sizeof(char *) * fileTypesSize);
```
Et7f3:
```suggestion
        fileTypes = (char **)malloc(sizeof(*fileTypes) * fileTypesSize);
```
less error prone.

src/Native/dialog.cpp
```diff
@@ -13,33 +14,123 @@
...
+
+        // Populate the array with the CAML array;
+        for (int i = 0; i < fileTypesSize; i++) {
```
Et7f3: Reverse loop + no need for local str :eyes: .

src/Native/dialog.cpp
```diff
@@ -13,33 +14,123 @@
...
+    fileList = revery_open_files_cocoa(startDirectory, fileTypes, fileTypesSize,
+                                       allowMultiple, canChooseFiles,
+                                       canChooseDirectories, buttonText, title);
+#endif
+
+    if (fileList) {
+        CAMLlocal1(camlArr);
+
+        int len = -1;
+        while (fileList[++len] != NULL) {
```
Et7f3:
```suggestion
        while (fileList[len]++ != NULL) {
```

src/Native/dialog.cpp
```diff
@@ -13,33 +14,123 @@
...
+#ifdef __APPLE__
+    fileList = revery_open_files_cocoa(startDirectory, fileTypes, fileTypesSize,
+                                       allowMultiple, canChooseFiles,
+                                       canChooseDirectories, buttonText, title);
+#endif
+
+    if (fileList) {
+        CAMLlocal1(camlArr);
+
+        int len = -1;
```
Et7f3:
```suggestion
        int len = 0;
```

src/Native/dialog.cpp
```diff
@@ -13,33 +14,123 @@
...
+        while (fileList[++len] != NULL) {
+        }
+        camlArr = caml_alloc(len, 0);
+
+        for (int i = 0; i < len; i++) {
```
Et7f3: reverse loop ?

src/Native/dialog_cocoa.c
```diff
@@ -14,4 +14,80 @@ void revery_alert_cocoa(void *pWin, const char *szMessage) {
...
+    }
+
+    if (buttonText)
+        [panel setPrompt:[NSString stringWithCString:buttonText
+                                            encoding:NSUTF8StringEncoding]];
+
+    if (nsFileTypes) [panel setAllowedFileTypes:nsFileTypes];
+
+    // Run the actual panel/modal
+    NSInteger result = [panel runModal];
```
Et7f3: panel leak ?

zbaylin: fixed in latest commit

src/Native/dialog.cpp
zbaylin: Resolved in latest commit

zbaylin:
I also am now adding bindings to the Gtk/Linux file dialog.

src/Native/dialog_cocoa.c
```diff
@@ -14,4 +14,82 @@ void revery_alert_cocoa(void *pWin, const char *szMessage) {
...
+    int allowMultiple, int canChooseFiles, int canChooseDirectories,
+    int showHidden, const char *buttonText, const char *title) {
+    /* Creates an empty NSArray of filetypes (NSString's)
+        If [fileTypes] is not null, copy the C-strings to NSString's to the
+       NSArray
+    */
+    NSArray *nsFileTypes = NULL;
+    if (fileTypes) {
+        NSMutableArray *tmpArr =
+            [[NSMutableArray alloc] initWithCapacity:fileTypesSize];
```
Et7f3: Where it is released ? `tmpArr`

zbaylin: Yep!

src/Native/dialog_cocoa.c
```diff
@@ -4,7 +4,7 @@
 #import <Cocoa/Cocoa.h>
 
 void revery_alert_cocoa(void *pWin, const char *szMessage) {
-    NSWindow* pCocoaWin  = (NSWindow *)pWin;
+    NSWindow *pCocoaWin = (NSWindow *)pWin;
 
     NSView *view = [[NSView alloc] init];
     NSAlert *alert = [[NSAlert alloc] init];
```
Et7f3: Should it bee released ?

zbaylin: This is not my code -- but I can modify it if you want.

src/Native/dialog_cocoa.c
```diff
@@ -14,4 +14,82 @@ void revery_alert_cocoa(void *pWin, const char *szMessage) {
...
+    // Run the actual panel/modal
+    NSInteger result = [panel runModal];
+
+    [panel release];
+    [nsFileTypes release];
+
+    // If a file(s) was selected...
+    if (result == NSModalResponseOK) {
+        // ...get the list of URLs
+        NSArray *urls = [panel URLs];
```
Et7f3: Is `urls` automatically released ?

src/Native/dialog_cocoa.c
Et7f3: Yeah I know but if it is a leak. I appreciate if this is fixed.

zbaylin: I don't think it's a leak. If you don't use `CAMLparam` & `CAMLreturn`, the OCaml garbage collector operates on the values passed into the function.

Et7f3: It if fine with OCaml but with the Obj-C part
> The rule is quite simple: you only need to release an object if you obtained it by using new, alloc, retain or copy (NARC for short). That's it. As soon as you're done with a NARC object and you don't need it anymore, you must release it. If you obtained the object though any other means, then you must not release it.

https://stackoverflow.com/a/6937337/7227940
For me I understand alloc/release like malloc/free in C.

src/Native/dialog_cocoa.c
Et7f3: Seem fixed in 838db54.

src/Native/dialog_cocoa.c
zbaylin: Ohh I see what you mean now.

zbaylin: Resolved in latest commit

bryphe:
Thank you @zbaylin and @Et7f3 ! This is awesome!

PR #661, opened 2019-12-01
Respect position in reconciler

glennsl:
I'm surprised revery works as good as it does just ignoring this. But unfortunately I stumbled across a bug, as I do, which is likely caused by this. I've therefore changed `node#addChild` to insert by position, and removed `_revChildren` as it no longer serves much of a purpose other than to add complexity. I also believe the assumption that elements typically should be appended is wrong. But even if it isn't, an ordinary list is probably the wrong data structure to use here anyway. A double-ended linked list or skip list might be better.

We should probably implement `moveNode` properly as well, and `removeNode` might be made slightly more efficient by using position rather than `id`, but that's probably not all that significant.

There seems to be no significant difference in the benchmark despite significant alterations to the reconciliation code. But perhaps the benchmark doesn't really test this.

Before:
```
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  BENCHMARK                                                     |  ITERATIONS |  TIME              |  MINOR GC  |  MAJOR GC  |  MINOR ALLOC  |  PROMOTED  |  MAJOR ALLOC  |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Node: create single node                                      |  10000      |  0.00162792205811  |  4         |  0         |  820302       |  307       |  307          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout single node (force re-layout)                  |  10000      |  0.00284290313721  |  3         |  0         |  540158       |  111       |  111          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout node tree (4 deep, 4 wide) (force re-layout))  |  10000      |  1.94358897209     |  821       |  273       |  179092661    |  22167906  |  22167906     |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout single node                                    |  10000      |  0.00290989875793  |  3         |  0         |  540158       |  111       |  111          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout node tree (4 deep, 4 wide)                     |  10000      |  1.94690203667     |  852       |  273       |  179091066    |  21563312  |  21563312     |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

After:
```
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  BENCHMARK                                                     |  ITERATIONS |  TIME              |  MINOR GC  |  MAJOR GC  |  MINOR ALLOC  |  PROMOTED  |  MAJOR ALLOC  |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Node: create single node                                      |  10000      |  0.00154399871826  |  4         |  0         |  800302       |  299       |  299          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout single node (force re-layout)                  |  10000      |  0.00318002700806  |  3         |  0         |  540158       |  111       |  111          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout node tree (4 deep, 4 wide) (force re-layout))  |  10000      |  1.94854593277     |  841       |  259       |  179089613    |  21305774  |  21305774     |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout single node                                    |  10000      |  0.00287389755249  |  3         |  0         |  540201       |  111       |  111          |
|----------------------------------------------------------------+-------------+--------------------+------------+------------+---------------+------------+---------------|
|  Layout: layout node tree (4 deep, 4 wide)                     |  10000      |  1.98566317558     |  847       |  269       |  179091657    |  21534614  |  21534614     |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

### DIFF
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -119,8 +119,7 @@ class node (()) = {
   pub getStyle = () => _style;
   pub setEvents = events => _events = events;
   pub getEvents = () => _events;
-  pub getChildren = () =>
-    _children;
+  pub getChildren = () => _children;
   pub getWorldTransform = () => {
     let state = _cachedNodeState |> getOrThrow("getWorldTransform");
     state.worldTransform;
@@ -259,34 +258,31 @@ class node (()) = {
     let bboxClipped = _this#getBoundingBoxClipped();
     BoundingBox2d.isPointInside(bboxClipped, p);
   };
-
   pub addChild = (child: node, position: int) => {
     let rec insert = (i, node, before, after) =>
       if (i > 0) {
         switch (after) {
-          | [] => after
-          | [head, ...tail] => insert(i - 1, node, [head, ...before], tail)
-        }
+        | [] => after
+        | [head, ...tail] => insert(i - 1, node, [head, ...before], tail)
+        };
       } else if (i == 0) {
-        insert(i - 1, node, before, [node, ...after])
+        insert(i - 1, node, before, [node, ...after]);
       } else {
         switch (before) {
-          | [] => after
-          | [head, ...tail] => insert(i, node, tail, [head, ...after])
-        }
+        | [] => after
+        | [head, ...tail] => insert(i, node, tail, [head, ...after])
+        };
       };
     _children = insert(position, child, [], _children);
     child#_setParent(Some((_this :> node)));
     _this#markLayoutDirty();
   };
-
   pub removeChild = (n: node) => {
     _children =
       List.filter(c => c#getInternalId() != n#getInternalId(), _children);
     n#_setParent(None);
     _this#markLayoutDirty();
   };
-
   pub firstChild = () => List.hd(_this#getChildren());
   pub getParent = () => _parent;
   pub getMeasureFunction = () => None;
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -262,7 +262,7 @@ class node (()) = {
     let rec insert = (i, node, before, after) =>
       if (i > 0) {
         switch (after) {
-        | [] => after
+        | [] => insert(0, node, before, after)
         | [head, ...tail] => insert(i - 1, node, [head, ...before], tail)
         };
       } else if (i == 0) {
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -28,15 +28,6 @@ type cachedNodeState = {
 };
 class node (()) = {
   as _this;
-  /* We use revChildren for appending, as appending to the _head_ of a list
-   * is much cheaper than appending to the _tail_ of a list.
-   * However, we often want the child in the 'correct' order - so we'll track
-   * when we've made changes to `_revChildren`, and update `_children` on request
-   * if `_revChildren` is dirty. `_childrenInvalid` tracks if `_children` needs
-   * to be updated.
-   */
-  val mutable _revChildren: list(node) = [];
-  val mutable _childrenInvalid = false;
   val mutable _children: list(node) = [];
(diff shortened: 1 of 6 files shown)


glennsl:
Can confirm that this fixes the issue with virtualized tree rendering in Oni2.

cc @wokalski :tada:

src/UI/UiEvents.re
```diff
@@ -101,7 +101,7 @@ let getTopMostNode = (node: node, pos) => {
 
       let ignored = mode == Ignore;
 
-      let revChildren = node#getRevChildren();
+      let revChildren = List.rev(node#getChildren());
       let ret =
         switch (revChildren) {
         | [] => ignored ? None : Some(node)
```
glennsl: I'm not entirely sure what this does. Is it just a depth-first search? Does it need to be done on a reversed list, or is that just because it's assumed to be the cheapest way to get the list of children?

src/UI/Node.re
```diff
@@ -277,16 +258,28 @@ class node (()) = {
...
+        | [head, ...tail] => insert(i - 1, node, [head, ...before], tail)
+        };
+      } else if (i == 0) {
+        insert(i - 1, node, before, [node, ...after]);
+      } else {
+        switch (before) {
+        | [] => after
+        | [head, ...tail] => insert(i, node, tail, [head, ...after])
+        };
+      };
```
glennsl: This is a generic `List` function. If there was a `Utility` module or something similar we should move it there.

bryphe: Good point... I don't think we have a `Utility` module here. Might make sense to have one like we have for Onivim.

src/UI/Reconciler.re
```diff
@@ -15,8 +15,8 @@ type node = reveryNode;
 
 let onStale: Event.t(unit) = Event.create();
 
-let insertNode = (~parent: node, ~child: node, ~position as _) => {
-  parent#addChild(child);
+let insertNode = (~parent: node, ~child: node, ~position) => {
+  parent#addChild(child, position);
```
bryphe: 🎉 🎉

src/UI/UiEvents.re
bryphe: Good question... this was an optimization introduced here https://github.com/revery-ui/revery/pull/536 to fix https://github.com/onivim/oni2/issues/665

We had a bug where we would look for the 'deepest' node instead of the 'topmost' node. To find the topmost node we'd look to find the _last_ child that matches the bounding box, recursively - so we'd end up reversing the nodes one very level.

At the time - this caused a very noticeable perf decrease - in the Quickmenu, for example, you'd visibly see a lag in the hover highlight when moving the mouse. However, I brought in this PR and didn't see that same perf issue - so if that's the case, removing the complexity is reasonable!

If we do see an issue again, in the future, with the performance of `getTopMostNode`... we could add a benchmark to cover that scenario which would also serve to document the performance requirements for that method.

Thanks for looking at simplifying this!

src/UI/Node.re
glennsl: Moved it to a local ListEx module for now at least.

PR #672, opened 2019-12-19
[Native] Bind to NSUserNotification(s) on macOS

zbaylin:
This creates the following API:

```
Revery_Native
  ├ type t
  ├ fun create
  ├ fun schedule
  └ fun dispatch
```

As of right now I need to implement do not disturb awareness and other times when NOT to show the notification


Et7f3:
Can you merge master + convert cpp to c (cpp might not be needed) after that is still -fPIC for windows is still relevant.

zbaylin:
> Can you merge master + convert cpp to c (cpp might not be needed) after that is still -fPIC for windows is still relevant.

I did this in commit #863ea00

src/Native/config/discover.ml
```diff
@@ -2,19 +2,19 @@ module C = Configurator.V1
 
 type os = Windows | Mac | Linux | Unknown
 
-type config = {libs: string list; cflags: string list; flags: string list}
+type config = {libs: string list; cflags: string list; cxxflags: string list; flags: string list}
```
Et7f3: are they still needed ?

zbaylin: No, I will fix that

zbaylin: Fixed in latest commit

zbaylin:
Alright looks like everything is done and is about to pass CI! 🎉 

Just in time for Christmas (although for me Hannukah has been happening since Sunday). Merry Christmas/Happy holidays everyone, and hopefully we can get this merged after the celebration(s) are over.

zbaylin:
@bryphe I can't request review since I'm not a member of the revery org, but let me know if you want to see any changes/improvements to the API

src/Core/App.re
```diff
@@ -184,6 +186,8 @@ let start = (~onIdle=noop, initFunc: appInitFunc) => {
     };
   };
 
+  Cocoa.setAppDelegate();
```
Et7f3: Is is mandatory in the reason side ? Can't you call [this](https://github.com/revery-ui/revery/pull/672/files#diff-d5f74c4ebde7224424f2f4c066f284cbR12) in notification function [revery_dispatchNotification](https://github.com/revery-ui/revery/pull/672/files#diff-7fefeeef1b753386656f889f57e1cb88R23) or [revery_scheduleNotificationFromNow](https://github.com/revery-ui/revery/pull/672/files#diff-30d9ca20ca5359755d388d3d2b0dceeaR33). With a static variable like `char g_delegate_set = 0;` (or `bool g_delegate_set = false;` if you `#include <stdbool.h>`).

zbaylin: I don't think that's a good idea. The "Application Delegate" does exactly what it sounds like: delegates the actions of the application. This is far from Notification specific, as you can tell in `ReveryApplicationDelegate.c`, which has "callbacks" like "on app start", etc. It would be a bad idea to have an indeterminate Application Delegate at any point in the application, because that would mean you wouldn't be able to guarantee the behavior of the application. Right now, as soon as the app starts, we are switching to Revery's implementation of the AppDelegate (which replaces the essentially empty SDL AppDelegate).

src/Native/Cocoa.re
```diff
@@ -0,0 +1,19 @@
+/* Cocoa.re
+     All macOS specific code, primarily setting the custom AppDelegate
+   */
+%import
```
bryphe: Looks like there was a failure building `esy '@bench' run` with this - does that need a PPX as well?

src/Core/App.re
bryphe: Looking at the docs for the delegate: https://developer.apple.com/documentation/appkit/nsapplication - I think it does make sense to register it at the top-level (if we start using it for other aspects besides notifications - it might be confusing that we only have it in the context of notifications.

src/Native/Cocoa.re
zbaylin: @bryphe I am not sure why it didn't compile -- it doesn't make much sense to me. `esy @bench run` compiles on both macOS and Windows, so I don't know why it wouldn't on Linux. I might have to boot up a Linux vm to test it. The strange thing is, in the previous CI build it ran and compiled fine with the `%import` expression. Was there any change to the cache that might have caused this?

bryphe:
> @bryphe I can't request review since I'm not a member of the revery org, but let me know if you want to see any changes/improvements to the API

Just sent you an invite! Thanks for all you've done on the native side 🎉 

Changes overall look good to me - thanks for all the working diving into the APIs here. Will help as we add more Cocoa-integrated functionality! (And thanks @Et7f3 for all the help reviewing).

The CI failure is the last thing - I wonder why that started? Might be worth bisecting to see when it failed, or if we need to include the optcomp ppx in the bench project.

zbaylin:
@bryphe 
> The CI failure is the last thing - I wonder why that started? Might be worth bisecting to see when it failed, or if we need to include the optcomp ppx in the bench project.

It is very strange, considering it ran fine in both the Windows and macOS branches. I'm wondering if it has anything to do with a caching issue?

src/Native/Cocoa.re
bryphe: > The strange thing is, in the previous CI build it ran and compiled fine with the %import expression. Was there any change to the cache that might have caused this?

I'm not very familiar with `ppx_optcomp`, but I doubt it. This should work whether its cached or not.

Given the build failure, I wonder if we could simplify and do this without `ppx_optcomp`?

Something like:
```
external cocoaSetAppDelegate: unit => unit = "revery_cocoaSetAppDelegate";
let _noop = () => ();

let setAppDelegate = switch (Environment.os) {
| Darwin => cocoaSetAppDelegate
| _ => noop;
};
```

We'd have to move the `Environment` module or `Environment.os` from Core -> Native (since Core depends on Native, but Native needs the Environment module).

zbaylin: The reason that doesn't work is because the linking mechanism expects for there to be a C function named `revery_cocoaSetAppDelegate`. Before I added ppx_optcomp, I added a bunch of `ifdef` directives in the cocoa files, but this only works if the function you're returning is `void`/`Val_unit`.

src/Core/App.re
Et7f3: Ok then can you add a init function so all native dependant code initialisation is stored at one place. Thus we don't need ppx_optcomp in reason world.

Et7f3: And write in the doc we use a custom responder so user have to override our.

src/Native/Cocoa.re
bryphe: Got it. Does a strategy like what we did  for `revery_alert` work?
```
CAMLprim value revery_alert(value vWindow, value vMessage) {
    CAMLparam2(vWindow, vMessage);
    const char *szMessage = String_val(vMessage);
    void *pWin = (void *)vWindow;

#ifdef WIN32
    revery_alert_win32(pWin, szMessage);
#elif __APPLE__
    revery_alert_cocoa(pWin, szMessage);
#elif __linux__
    revery_alert_gtk(pWin, szMessage);
#else
    printf("WARNING - Not implemented: alert");
#endif
    CAMLreturn(Val_unit);
}
```

src/Core/App.re
Et7f3: Other advantage of init function is all code is store in one place.

src/Native/Cocoa.re
zbaylin: That would work, but again only if you're returning a "`void`" data type. Let's say I was returning a primitive like an int or even a C data type with empty type information in OCaml. Since there is no "null" in OCaml, your only real option is to do something like casting a void pointer to a that data type.

src/Core/App.re
zbaylin: Sure. Where would that be kept? Directly in `Revery_Native`?

i.e.
```reason
ReveryNative.init()
```

Et7f3: Yeah :)
(also add a clean function registered with at_exit)

zbaylin: Will do 👍

bryphe: Sounds good! This could be where we register other native-init stuff, like creating a GTK application

zbaylin:
> Why warning ? It is ok if nothing has to be done for init.

This is true. I can remove this.

> What is the aim of one module for only one function ? It might be confusing that we can also call Initialization.init.

You can't call Initialization.init except for within Revery_Native, because I included it rather than saying `module Initialization = Initialization`

> Why warning ? It is ok if nothing has to be done for init.

src/Native/Revery_Native.c
```diff
@@ -0,0 +1,28 @@
...
+#else
+#include "ReveryGtk.h"
+#endif
+
+#define Val_none Val_int(0)
+
+#define Some_val(v) Field(v, 0)
+
+CAMLprim value revery_initialize() {
+#ifdef __APPLE__
```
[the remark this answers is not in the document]
zbaylin: Before @Et7f3 said he preferred an init function because it localizes everything in one place, and bryphe agreed.

src/Native/Revery_Native.c
```diff
@@ -0,0 +1,28 @@
...
+
+#include <caml/alloc.h>
+#include <caml/callback.h>
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
+#include <string.h>
+
+#define UNUSED(x) (void)(x)
+
+#ifdef WIN32
```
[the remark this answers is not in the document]
zbaylin: Yeah part of me wants to make a `caml_tools.h` that just has everything in it. I'll do that now.

src/Native/Revery_Native.c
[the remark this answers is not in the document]
Et7f3: + this is like our main so you can inline `revery_cocoaSetAppDelegate` (it should not be called from other place)

src/Native/Revery_Native.c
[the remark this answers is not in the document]
Et7f3: Or in utilities.h (maybe this is so useful that they can be upstreamed)

zbaylin:
Alright I think we have all our bases covered now. @bryphe are we ready to merge this in? As soon as we do I'll start working on the Linux and Windows implementations.

src/Native/caml_tools.h
```diff
@@ -0,0 +1,18 @@
+#include <caml/alloc.h>
```
Et7f3: #ifndef guard are missing.
+ I don't think put all caml header in one file is good(some c file need only one file) If it is good ocaml itself should have provide one file.

zbaylin: What `#ifndef` guard are you referring to?

I can see why you would be against it. I can put it back to the way it was.

Et7f3: guard is a archaic system when you don't have import https://fr.wikipedia.org/wiki/Include_guard

src/Native/dialog.c
```diff
@@ -1,6 +1,21 @@
...
-#include "caml_tools.h"
+#include <caml/alloc.h>
+#include <caml/callback.h>
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
+
+#include "caml_values.h"
+
+static inline value Val_some(long v) {
+  (void)Val_some;
```
Et7f3: 👀 Why ?

zbaylin: My bad. Fixed.

src/Native/caml_tools.h
zbaylin: Fixed.

PR #673, opened 2019-12-22
Et7f3/fix warnings c

Et7f3:
This PR transform C++ into C (we don't need the overhead of C++ right now) + add `-Wall -Wextra -Werror` in c_flags so we can catch bug earlier. If one day we have to use C++ we should add `cxx_flags` (or equivalent) in dune files.


Et7f3:
The last status was clean for OS X and Linux now it should be all green (because I tested on windows).

src/Native/dune
```diff
@@ -4,9 +4,8 @@
     (preprocess (pps lwt_ppx))
     (library_flags (:include flags.sexp))
     (js_of_ocaml (javascript_files dialog.js))
-    (c_names dialog_cocoa dialog_win32 dialog_gtk)
-    (cxx_names dialog)
-    (c_flags (:include c_flags.sexp))
+    (c_names dialog dialog_cocoa dialog_win32 dialog_gtk)
+    (c_flags :standard -Wall -Wextra -Werror (:include c_flags.sexp))
```
bryphe: Thanks for adding this / fixing the warnings, @Et7f3 ! 🎉

bryphe:
Good call. I think using `-Wall` and `-Werror` + hooking up address sanitizer for our tests makes sense to keep our C layer as solid as we can.

PR #664, opened 2019-12-07
Skia: Add reason-skia as esy dependency

bryphe:
Just a baby-step to parcel out some of the changes in the Skia branch happening in #567  - we can start bringing in `reason-skia` as a build dependency to exercise the CI path.

This will increase first-time build times; but we will be bringing in this dependency anyway!

__TODO:__
- [x] Fix OSX build issues
- [x] Fix Windows build issues
- [x] Fix Linux build issues

Current Windows build isssue:
```
##[command]"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "d:\a\_temp\82cd0153-5503-48e3-84a3-485600024e59.cmd""
info install 0.5.8 (using package.json)
info fetching: done
error command failed: "tar" "xf" "/cygdrive/d/a/1/s/_esy/default/dist/da80c02f0818cee1e16ea5d93347a718" "-C" "/cygdrive/c/Users/VSSADM~1/AppData/Local/Temp/esy-40015b"
      stderr:
               tar: revery-ui-esy-skia-b7091c8/tools/gyp: Cannot create symlink to '../third_party/externals/gyp/': No such file or directory
               tar: Exiting with failure status due to previous errors
               
      stdout:
               
error: error running command
  unpacking esy-skia@github:revery-ui/esy-skia#b7091c8
  installing esy-skia@github:revery-ui/esy-skia#b7091c8
esy: exiting due to errors above
```

Current Linux build issue:
```
      ocamlc src/Native/dialog_gtk.o (exit 2)
[...]

PR #676, opened 2019-12-24
CI: Fix publish build cache step

bryphe:


Et7f3:
To see availables packages `pkg-config —list-all` this will help troubleshoot gtk problem in #664

PR #681, opened 2019-12-25
Test: Log build-env w/o ctypes to check PKG_CONFIG_PATH

bryphe:
Looks like package gets found here: `Will find package 'gtk+-3.0' in file '/usr/lib/x86_64-linux-gnu/pkgconfig/gtk+-3.0.pc`


bryphe:
Looks like the issue is addressed with: https://github.com/esy-ocaml/yarn-pkg-config

PR #683, opened 2019-12-26
Skia: Switch from reason-fontkit to reason-harfbuzz

bryphe:
Skia handles font rendering and measurements for us - so we don't need to worry about bringing in `freetype2` anymore. 

However, we still need `harfbuzz` to handle shaping.

`reason-fontkit` used both harfbuzz+freetype2 - now we can break that dependency and just use harfbuzz.


bryphe:
This is integrated now in #567

PR #685, opened 2019-12-27
Overhaul text-wrapping using revery-text-wrap library

### LINKED ISSUE #617: Text wrapping is borked
There are several issues with the current text wrapping implementation:

- If it can't fit the first word in the max width, it wraps after the first character of the first word
- The character at the wrap point is simply removed, which means in-word characters disappear if the wrap point is in a word.
- Tabs, and probably other white-space characters, seem to not be considered white-space.
- Consecutive and compound white-space is handled poorly

---

*Edit: original title and post:*

Title: Text wrapping drops character where it wraps when forced to break a word

The character is probably thrown away because it's assumed to be white-space, since it's designed to break between words, not _in_ words.

It also doesn't calculate the optimal location to break the word, but instead just breaks after the first character.

2019-10-24-213616_152x36_scrot

OhadRau:
This commit replaces the text wrapping code in TextWrapping.re with
my new implementation from [revery-text-wrap](https://github.com/OhadRau/revery-text-wrap).
This new code should be much faster, performing wrapping in O(n) time.
However, there are currently some limitations: this doesn't dynamically
check the width of the line while wrapping, which means that certain
glyphs will be the wrong size, since unicode-combined characters and ligatures aren't
understood by the library (yet). However, the actual wrapping looks much
better, with most of the known bugs fixed from the old implementation.

text-rendering

Resolves #617


OhadRau:
Thanks for reviewing!

>Why does it use a functor? 

That's actually a question I was asking myself at the time when I was writing the library. There's 2 (small) reasons behind that, but I'm not sure either is a great answer.

1) It made the code a little simpler since I could avoid passing the `width_of_char` function to the `width_of_token` function (and before refactoring the code there were some other dependencies). It may also be useful for tokenization to have access to the "configuration" in one way or another. However, this can also be solved by passing around some of the state either directly or as part of a configuration record.

2) The other reason is that I imagine the use-case of this as having a few modules around for different environments. For example, if you had a rich text editor, it would be nice to have a `CodeBlockWrap` module and a `TextWrap` module that you could switch between easily. These would already differ in terms of the font used & the wrapping style (whether to hyphenate, whether to overflow, whether to remove extra whitespace, etc.), but in the future we could also pass in custom rules for when to break & such (`let decide = (~token, ~lastToken, ~position, ...) => [`Break | `Hyphenate(position) | `AutoHyphenate | `Ignore | `Continue]`). Either way, the functor does make it a little easier to juggle the different states in that case.

With both of those reasons in mind, I'm still not sure whether functors are the right decision here so just wanted to bring the reasoning into the conversation so I can hear some more thoughts on this. Shouldn't be a difficult change to make it use records or something else, so happy to change it if that seems important.

>Why depend on Batteries?

Right, this is also a good question. There were some places where performance seemed better (originally I was using Batteries' substring module which is a lot faster at slicing) but I do want to get rid of the dependency either way. I started refactoring some of the code to depend less on batteries, just didn't get to removing `String.slice` yet. Should fix this in the next commit hopefully.

>AppendList looks like it's basically just a Queue, or a Deque if you eventually also need to prepend

Lol I'm so used to avoiding the stdlib's data structures in OCaml that I just assumed the Queue implementation would not be performant enough. Looking at the code it should be fine, so I'll refactor to use that.

OhadRau:
Went ahead and changed all 3 of those things. Cut down a ton on code and one less dependency :)

src/Core/TextWrapping.re
```diff
@@ -4,17 +4,22 @@ type wrapType =
...
+  let rec list_of_queue = q =>
+    if (Queue.is_empty(q)) {
+      []
+    } else {
+      /* We need to split into two lines because the RHS of ::
+         gets evaluated first, leading to recursion before the
+         item is removed from the queue. */
+      let first = Queue.take(q);
+      [first, ...list_of_queue(q)]
+    };
```
glennsl: Wouldn't it be better to have `wrap` return a list and make this and the queue an implementation detail? Are there cases where you might want to not do this?

OhadRau: That might be the way to go, but we could actually increase performance just a little bit if we use the Queue internally since we don't need to copy everything into a list (so no list allocation & no extra iteration to make it a list). Thoughts?

glennsl:
> if you had a rich text editor, it would be nice to have a CodeBlockWrap module and a TextWrap module that you could switch between easily. These would already differ in terms of the font used & the wrapping style (whether to hyphenate, whether to overflow, whether to remove extra whitespace, etc.), but in the future we could also pass in custom rules for when to break & such

You can also do this with plain function arguments using partial application. But it might definitely be more natural to use a functor once you add configurable rules, since you'll probably want to configure that only once project-wide.

OhadRau:
Yeah, a big part of the question here is how general should I really make this library, since Revery's needs are probably a little bit simpler I could just make the library fit those directly. But if I implement things like custom rules the library becomes much larger in scope & it becomes less specific to Revery. For now I went with just parameters for those options but will probably reevaluate the decision based on where the library goes in the near future.

OhadRau:
So this actually causes a regression in the calculator example. Looking into that real quick

image

src/Core/TextWrapping.re
glennsl: The resulting list would typically be very small though, 1 or 2 elements. I don't think the difference will be significant other than for pathological cases.

OhadRau:
Fixed that bug & improved the text example a little bit. It doesn't refresh when switching hyphenation mode (because that doesn't trigger a `measure`) but other than that looks pretty good. Hyphenation is still a bit wonky so I'll look into that, but it seems like it has to do with the measured character widths not lining up with the actual character width (due to kerning/ligature weirdness). One solution to this might be making the text wrapping a 2-pass process where the first wrap measures based on the assumed *width of each character* and then the second pass goes back and adjusts based on the measured *width of the entire line*. Regardless, that seems a little out of scope right now since hyphenation is disabled by default.

src/Core/TextWrapping.re
OhadRau: Fixed!

glennsl:
To fix the CI failure you need to update and commit the lockfiles for the sub-projects. Run `esy @test i && esy @bench i && esy @doc i`, then commit and push.

Et7f3:
> To fix the CI failure you need to update and commit the lockfiles for the sub-projects. Run `esy @test i && esy @bench i && esy @doc i`, then commit and push.

@glennsl or use this shell script https://github.com/revery-ui/revery/blob/master/update-lockfiles.sh 🙃 
Maybe I can add what I proposed for brisk-reconciler https://github.com/briskml/brisk-reconciler/pull/70 (i.e. it auto submit new lockdirs)

OhadRau:
Hmm.. doesn't look like the lock-files fixed that... any idea why?

glennsl:
Oh, you need to run `esy @js i` too (or just run `./update-lockfiles.sh`). Sorry I missed that.

You also might have to run `esy format` if you haven't.

PR #689, opened 2019-12-28
[Native] Revery_Native.Icon Badge API

zbaylin:
This implements a very simple Badge API in `Revery_Native`

API:
```
Revery_Native
  |
  +-> Badge
        |
        +-> func setNumber: int => int
        |
        +-> func hide: unit => unit
```


zbaylin:
Alright I modified the API a bit
```
Revery_Native
  |
  +-> Badge
        |
        +-> func set: string => unit
        |
        +-> func hide: unit => unit
```

In Electron, the badge API only works on macOS. [This repo](https://github.com/viktor-shmigol/electron-windows-badge) basically modifies the app icon to include a badge on Windows. We could essentially do the same with something like ImageMagick, but I'm not sure if it's worth it. 

@bryphe -- thoughts on whether or not to have this API on Windows/Linux?

OhadRau:
Can't we use this on Windows? https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/badges

zbaylin:
@OhadRau That's a UWP only API that isn't accessible for WIndows 32 apps (unlike Toast Notifications)

zbaylin:
In #665 I switched to housing the progress bar API under `Revery_Native.Icon`, and I think it would be appropriate to move this under there too, especially because on Windows the type represented as `Icon.t` (`ITaskbarList3`) is also used to set the overlay icon.

As stated before, Win32 has no API for badges. Interestingly, inspecting the Discord `asar` shows that they have independent `ico` files for each number. @Et7f3 noted that we could use SDL to dynamically generate the bitmap/icon and display it using `SetOverlayIcon`

zbaylin:
🎉 I got a dynamic badge working on Win32! It uses GDI to import a template image `badge.bmp` and draw text on it. If the text is longer than 3 characters, it shows the first two and ellipsizes the rest.

This GIF is a pretty good demonstration (sorry for the low quality, Github only allows 10MB):

screencast 2020-01-12 22-26-45

I'm pretty excited about this, because not even Electron supports this natively on Windows.

zbaylin:
I removed a dependency on a `badge.bmp` (which was a simple bitmap I created in Photoshop) as @Et7f3 pointed out that packages not in the `Revery` namespace wouldn't have access to that file. Now the badge is generated as a MetroUI red rectangle. Thanks @Et7f3!

The next thing I want to do before merging is integrate this: https://docs.microsoft.com/en-us/windows/win32/gdi/string-widths-and-heights. Right now there is (some) overflow, and it doesn't make sense to have 1 character strings be the same font size as 3 character strings.

src/Native/icon_win32.c
```diff
@@ -49,29 +49,27 @@ void revery_setIconBadge_win32(void *win, void *ih, char *badgeStr) {
...
-        displayStr = (char *)malloc((6) * sizeof(char));
-        displayStr[0] = badgeStr[0];
-        displayStr[1] = badgeStr[1];
-        // ...an ellipsis...
-        displayStr[2] = displayStr[3] = displayStr[4] = '.';
-        // ...and a null termination.
-        displayStr[5] = '\0';
+    int fontSize;
+
+    switch (strlen(badgeStr)) {
```
Et7f3: in OCaml string carry their length in first byte. So maybe not needed.

src/Native/icon_win32.c
```diff
@@ -83,13 +81,13 @@ void revery_setIconBadge_win32(void *win, void *ih, char *badgeStr) {
...
     HFONT hFont = CreateFont(
-                      20, 0, 0, 0, FW_BOLD, 0, 0, 0, DEFAULT_CHARSET, OUT_DEFAULT_PRECIS,
+                      fontSize, 0, 0, 0, FW_BOLD, 0, 0, 0, DEFAULT_CHARSET, OUT_DEFAULT_PRECIS,
                       0, DEFAULT_QUALITY, DEFAULT_PITCH | FF_DONTCARE, "Arial");
     SelectObject(memHDC, hFont);
     // Make sure the text background is transparent
     SetBkMode(memHDC, TRANSPARENT);
     SetTextColor(memHDC, RGB(255, 255, 255));
-    DrawText(memHDC, displayStr, strlen(displayStr), &textRect, DT_CENTER | DT_WORDBREAK);
+    DrawText(memHDC, badgeStr, strlen(badgeStr), &textRect, DT_CENTER | DT_SINGLELINE | DT_VCENTER | DT_END_ELLIPSIS);
```
Et7f3: :heart: DT_END_ELLIPSIS

src/Native/icon_win32.c
```diff
@@ -31,4 +34,85 @@ void revery_hideIconProgress_win32(void *win, void *ih) {
...
+    default:
+        fontSize = 20;
+        break;
+    }
+
+
+
+    // Create a rect to paint the text into on top of the bitmap
+    RECT textRect;
+    textRect.left = 0;
```
Et7f3:
```suggestion
    textRect.left = (32 - fontSize) / 2;
```
So it aligned vertically I don't see where we specify line height.

zbaylin: It already is aligned vertically. Check out `DrawText`, i.e. `DT_CENTER` and `DT_VCENTER`.

Line height is also specified as the first argument to `DrawText`

src/Native/icon_win32.c
zbaylin: I thought of this too, but `String_val` converts it to a standard C-string (`char*`) so `strlen` is necessary

src/Native/icon_win32.c
Et7f3: `memHDC` ?

zbaylin: My mistake 😳, it is the first argument to `CreateFont`:
https://github.com/revery-ui/revery/blob/3c1cf15f02243e8ea41fd75fbdb7e709f9b395c4/src/Native/icon_win32.c#L83-L85

zbaylin:
Bumping this PR because I think it's done and just requires approval. @bryphe or @Et7f3 (or really anyone with a Windows box), if you could retest this on a Windows machine, that would be really helpful. I've just been testing on a VM.

PR #691, opened 2019-12-28
Native/dialog.c: Rename "revery_alertOpenFiles_bytcode" to "revery_alertOpenFiles_bytecode"

OhadRau:
Previously this function was called "revery_alertOpenFiles_bytcode" which would result in a build failure using the bytecode backend


bryphe:
Thanks for fixing this, @OhadRau !

PR #693, opened 2019-12-29
Enable bytecode builds

OhadRau:
This PR is a really simple fix to get bytecode builds working (updates `examples/dune` to provide the proper ocamlc flags & allow building bytecode) and then 2 simple `esy` scripts to get build bytecode/native executables work.

### DIFF
--- a/examples/Hello.re
+++ b/examples/Hello.re
@@ -104,12 +104,15 @@ module AnimatedText = {
   };
 };
 
-let render = () =>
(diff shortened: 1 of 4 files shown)


Et7f3:
merge #691 two times 😂

OhadRau:
Ahh yeah this is going into the feat/hot-reload branch so that's why we need it twice

PR #696, opened 2019-12-30
package.json: Use text-wrap from revery-ui org; remove batteries dependency

OhadRau:
Quick fix because of the CI failures


bryphe:
Thanks for the quick fix @OhadRau ! Looks great 👍 I'll bring it now to unblock CI-

PR #698, opened 2019-12-30
API: App - add '~askNicely' argument to App.quit

bryphe:
From a conversation with @glennsl today - we should allow a `force` quit (which ignores any `canQuit` callbacks registered with the window).

We assume that, if a developer is calling `App.quit`, then they probably mean to force quit - so `force` defaults to true. However, if `force` is `false`, we'll still run the `canQuit` callbacks.


bryphe:
Thanks @glennsl for the feedback. I updated it to `~askNicely` 👍

PR #701, opened 2020-01-01
A few improvements to the spring hook.

glennsl:
Adds the ability to set the position immediately and to configure the threshold below which it should be considered at rest. Also cleaned up and simplified the hook quite a bit.

### DIFF
--- a/examples/SpringExample.re
+++ b/examples/SpringExample.re
@@ -9,6 +9,13 @@ module SpringyLogo = {
     let%hook (imageWidth, _setImmediately) =
       Hooks.spring(
         ~target=targetPosition,
+        ~initialState=
+          Spring.{
+            value: 0.,
+            velocity: 10.,
+            acceleration: 0.,
+            time: Time.now(),
+          },
         Spring.Options.create(~damping, ~stiffness, ()),
       );
 
--- a/examples/SpringExample.re
+++ b/examples/SpringExample.re
@@ -6,7 +6,7 @@ module SpringyLogo = {
   let%component make = (~stiffness, ~damping, ()) => {
     let%hook (targetPosition, setTargetPosition) = Hooks.state(1.0);
 
-    let%hook imageWidth =
+    let%hook (imageWidth, _setImmediately) =
       Hooks.spring(
         ~target=targetPosition,
         Spring.Options.create(~damping, ~stiffness, ()),
--- a/examples/SpringExample.re
+++ b/examples/SpringExample.re
@@ -8,7 +8,7 @@ module SpringyLogo = {
 
     let%hook imageWidth =
       Hooks.spring(
-        targetPosition,
+        ~target=targetPosition,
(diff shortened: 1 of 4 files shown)


src/UI_Hooks/Spring.re
```diff
@@ -1,68 +1,21 @@
...
-      ~initialState=initialState(v, opts, Time.now()),
-      springReducer,
-    );
+let spring = (~target, ~restThreshold=0.1, options) => {
+  let%hook (previousState, setPreviousState) =
+    Ref.ref(Spring.create(target, Time.now()));
 
-  let%hook (time, _) =
-    Timer.timer(~active=curr.isActive, ~tickRate=Time.zero, ());
+  let isActive =
```
bryphe: Wow, nice simplification here!

examples/SpringExample.re
```diff
@@ -6,10 +6,10 @@ module SpringyLogo = {
   let%component make = (~stiffness, ~damping, ()) => {
     let%hook (targetPosition, setTargetPosition) = Hooks.state(1.0);
 
-    let%hook imageWidth =
+    let%hook (imageWidth, _setImmediately) =
       Hooks.spring(
-        targetPosition,
-        Spring.Options.create(~damping, ~stiffness, ~initialValue=2.0, ()),
+        ~target=targetPosition,
+        Spring.Options.create(~damping, ~stiffness, ()),
```
bryphe: There is a subtle behavioral difference here - in the previous iteration, setting `initialValue` of 2.0 and `targetPosition` of 1.0 meant that the spring would be 'stretched out' initially and bounce in.

Is it possible to do that with this new model? I tried this:
```
    let%hook (imageWidth, _setImmediately) =
      Hooks.spring(
        ~target=targetPosition,
        Spring.Options.create(~damping, ~stiffness, ()),
      );

    let%hook () = Hooks.effect(OnMount, () => {
      _setImmediately(2.0); 
      None;
    });
```
but didn't quite give the results I expected.

glennsl: Ah, I missed that possibility. I considered allowing the initial state to be passed in, which would offer much more flexibility and be more intuitive IMO. But I think I dropped or forgot about it because I considered it along with merging the options and state data structures, which didn't seem like such a great idea after all.

I can't think of any good reason for not adding an optional `initialState` parameter though, so if you don't have any objections either I can do that.

bryphe: `initialState` does seem better - applying an initial velocity / acceleration could be interesting too!

I'm OK either way - we can add it later if we see a need. Just wanted to validate my understanding 👍 Thanks for the details!

glennsl: Added in https://github.com/revery-ui/revery/pull/701/commits/8ab5088885a7c08926e6383234a8c025ac0c9b70.

We can add convenience functions later to create a state from an an initial and target value, to avoid having to specify acceleration and velocity manually.

Issue #703, opened 2020-01-02
OSX - Catalina: Key repeat broken


bryphe:
- Operating System: OSX Catalina
- Revery Version: `master` (commit `99441c7`)

The key repeat is broken on OSX Catalina, somewhere between `d2fd73b` and `99441c7`.

__REPRO:__
- Run revery example app (`esy run`)
- Go to input example
- Press and hold `j`

Expected: `j` should repeat and you should end up with `jjjj...` in the input box
Actual: Only one `j` is entered

Interestingly, this does not repro on OSX 10.14.

This is blocking upgrade for Onivim 2: onivim/oni2#1149

CrossR:
Another interesting thing is that it looks like backspace and enter (probably more) do work still, but letters don't.

Ie in that same example, you can hold backspace to delete all of the letters, or in Oni2 holding enter does move the cursor as much as expected.

CrossR:
I first see this behaviour once this PR is merged: https://github.com/revery-ui/revery/pull/672

I expect its something unrelated to actual input, and more something with some macOS specific option (`ApplePressAndHoldEnabled` or something) and hooking up the notifications has tripped that up.

That is further solidified by the fact that if you hold something like `i` you get the accent pop-up menu (in the corner of the window) and that it doesn't exhibit for arrows or backspace/enter.

<img width="838" alt="image" src="https://user-images.githubusercontent.com/10038688/71701657-c4fc5d00-2dc2-11ea-8b5c-e8e232982f27.png">

glennsl:
Apparently this is still an issue. Reported on oni2: https://github.com/onivim/oni2/issues/1222

CrossR:
Should be fixed with #716.

bryphe:
Tested on 10.14/Catalina - looks good to me. Thanks @CrossR and @zbaylin !

PR #704, opened 2020-01-03
[WIP] Fix key-repeat.

CrossR:
Fix https://github.com/revery-ui/revery/issues/703. (Well will once actually moved to a sensible place).

Where does it make sense to add this? I'd guess a new function that we let users call to disable it. The one potential complication is that it only needs setting one time. Once its set, you get it set permanently. So we'd probably need a disable option too.

(For anyone testing this, you can follow up with a `defaults delete com.revery.examples` in bash to remove the key it adds for now, to test before and after.)

I've confirmed the fix works over in Oni2 as well.

### DIFF
--- a/src/Native/cocoa/ReveryAppDelegate.c
+++ b/src/Native/cocoa/ReveryAppDelegate.c
@@ -17,6 +17,11 @@
     if (self) {
         _notificationActions = [NSMutableDictionary new];
     }
+
+    NSString *valueToSave = @"true";
+    [[NSUserDefaults standardUserDefaults] setObject:valueToSave forKey:@"ApplePressAndHoldEnabled"];
+    [[NSUserDefaults standardUserDefaults] synchronize];
+
     return self;
 }
 


CrossR:
@zbaylin , not sure if you have any input on this, since you've been working in the area?

Et7f3:
> Values returned from NSUserDefaults are immutable, even if you set a mutable object as the value. For example, if you set a mutable string as the value for “MyStringDefault”, the string you later retrieve using the stringForKey: method will be immutable. If you set a mutable string as a default value and later mutate the string, the default value won’t reflect the mutated string value unless you call setObject:forKey: again.

zbaylin:
Unfortunately I'm still on Mojave (10.14) because there is some 32-bit software I still need to use, so I'm afraid I won't be a ton of help on this.

bryphe:
I'll bring this in so that we can sync Onivim 2 up to the latest Revery (this bug was blocking). Thanks again @CrossR !

CrossR:
I think in the future, we may want to allow this to be configurable. Perhaps by moving it to its own function so its opt-in, or inversely have it be opt-out. For now though, its a good compromise such that the expected behaviour is given in the Examples etc (as well as being consistent across platforms, as no other platform has a pop-up menu for accents etc).

zbaylin:
@CrossR the annoying thing about making it opt in would be that you would have to add some optional compilation magic to `Revery_Native.init`, because it wouldn't make sense to pass that parameter in on platforms that don't require it. 

In the PR that I initially added `init` in, I initially added it as something like `Revery_Native.Cocoa.setAppDelegate()`. However, since adding stuff like [`ppx_optcomp`](https://github.com/janestreet/ppx_optcomp) proved to be too complex, we moved it to a universal `init` function. Hopefully sometime in the future we can think of a more elegant solution.

Issue #709, opened 2020-01-10
Clickables cannot be nested


glennsl:
`Clickable` is currently implemented by capturing the mouse on `mousedown` and then executing `onClick` on `mouseup` if the pointer is still in the elements bounding box at that point. Not a bad approach, in theory, but unfortunately `mousedown` bubbles and the capture API is imperative and exclusive, which means the shallowest clickable is always going to hijack the capture and register the clicks as happening on itself.

There are a few ways I can think of to address this:

1. Fixing the bubbling so it's conditional somehow. There are a few approaches I can think of for doing this as well, all of which are breaking:
  a. Require event handlers to specify in advance whether events should be propagated. There are various ways to do this too, of course, but they basically boil down to the same concept.
  b. Pass a `next` callback to the event handler, along with the event, so events can be bubbled selective, CPS-style.
  c. Require event handlers return a value to indicate whether to bubble.
2. Implement `Clickable` using a local flag that's set on `mousedown` and reset on `mouseout`/`mouseup` instead of using `capture`. These event will still bubble however, generating click events from all clickables in the hierarchy, so I think we'll need a global mutable as well to make it exclusive and prevent that. Pretty nasty, but non-breaking if it works.

I'll try to get option 2 working, but this isn't a long-term solution and we'll need to fix bubbling at some point anyway. so if anyone has ideas, concerns or other comments about the above, now might be a good time to bring them to light :slightly_smiling_face:

Et7f3:
`Addresses` is not a close keyword but 710 is merged so I close it.

PR #710, opened 2020-01-11
Fix nested clickables

glennsl:
Addresses #709

This allows clickables to be nested without ancestors hijacking the clicks of its descendants. It does so by implementing a local capture mechanism instead of using the global one that can be hijacked by anything anywhere in the codebase. By combining a module-local and component-local flag we can at least ensure that nothing outside the module can interfere, and we prevent different instances of the component from stepping on each others toes. The first instance to capture, which will be the deepest node, keeps the "lock" until it releases it, on `mouseleave`, `mouseup` or unmount. Hopefully that covers all the bases, but if not it's at least easier to reason about the mutation when it's local and encapsulated.

This also simplifies and fixes the `BubbleEvent` implementation so it should now work if it's to be used for anything, although currently it's not. And it cleans up various bits around the place. This isn't strictly necessary but, you know, I was in the area and trying to grok these things...

### DIFF
--- a/examples/NestedClickable.re
+++ b/examples/NestedClickable.re
@@ -1,0 +1,29 @@
+open Revery;
+open Revery.UI;
+open Revery.UI.Components;
+
+module Styles = {
+  open Style;
+
+  let outer = [
+    position(`Absolute),
+    top(0),
+    bottom(0),
+    left(0),
+    right(0),
+    justifyContent(`Center),
+    alignItems(`Center),
+    backgroundColor(Colors.yellow),
+  ];
+};
+
+let%component clickies = () => {
+  let%hook (text, setText) = Hooks.state("Click something");
+
+  <Clickable
+    style=Styles.outer onClick={() => setText(_ => "Clicked outside")}>
+    <Button title=text onClick={() => setText(_ => "Clicked inside")} />
+  </Clickable>;
+};
+
+let render = () => <clickies />;
--- a/src/UI/Mouse.re
+++ b/src/UI/Mouse.re
@@ -450,10 +450,8 @@ let dispatch =
 
         switch (deepestNode) {
         | None =>
-          /*
-           * if no node found, call bubbled MouseOut on deepestStoredNode if there's some stored nodes
-           * And recursively send mouseLeave events to storedNodes if they exist
-           */
+          // if no node found, call bubbled MouseOut on deepestStoredNode if there's some stored nodes
+          // And recursively send mouseLeave events to storedNodes if they exist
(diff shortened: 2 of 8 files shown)


src/UI/Mouse.re
```diff
@@ -428,80 +428,70 @@ let rec handleMouseEnterDiff = (deepestNode, evtParams, ~newNodes=[], ()) => {
...
-      if (mouseDown) {
-        switch (getFirstFocusable(node, pos)) {
-        | Some(node) => Focus.dispatch(node)
-        | None => Focus.loseFocus()
-        };
-      } else {
-        ();
+    (cursor: Cursor.t, evt: Events.internalMouseEvents, node: Node.node) =>
+  if (node#hasRendered()) {
+    let pos = getPositionFromMouseEvent(cursor, evt);
```
bryphe: Thanks for cleaning up the logic here!

src/UI/Mouse.re
```diff
@@ -428,80 +428,70 @@ let rec handleMouseEnterDiff = (deepestNode, evtParams, ~newNodes=[], ()) => {
...
+          | [node, ..._] => bubble(node, MouseOut(mouseMoveEventParams))
+          };
+
+          sendMouseLeaveEvents(storedNodesUnderCursor^, mouseMoveEventParams);
+
+        | Some(deepNode) =>
+          switch (storedNodesUnderCursor^) {
+          | [] =>
+            /*
+             * If some deepNode is found and there aer no storedNodes
```
bryphe: Just a minor type ("aer" should be "are")

src/UI/UiEvents.re
```diff
@@ -1,60 +1,29 @@
...
-  };
-
-  let generateId = () => {
-    let id = ref(1);
-    () => {
-      id := id^ + 1;
-      id^;
+module BubbleEvent: {
+  type t =
+    pri {
```
bryphe: Neat refactoring here, and thanks for removing `id` 👍

examples/NestedClickable.re
```diff
@@ -0,0 +1,29 @@
...
+    justifyContent(`Center),
+    alignItems(`Center),
+    backgroundColor(Colors.yellow),
+  ];
+};
+
+let%component clickies = () => {
+  let%hook (text, setText) = Hooks.state("Click something");
+
+  <Clickable
```
bryphe: Thanks for adding an example to exercise this case!

src/UI/Mouse.re
glennsl: Fixed :)

PR #716, opened 2020-01-15
[macOS] Fix #713

### LINKED ISSUE #713: Bug - OSX: Application window doesn't get focus on start
Looks like this is a regression from https://github.com/revery-ui/revery/pull/672

_REPRO:__
- On OSX, run the example app (`esy run`)
- Start typing

__Expected:__ Application should be focused, and taking the keyboard input
__Actual:__ The console is focused, and taking keyboard input

This is a blocker for upgrading Onivim 2 to latest Revery, so I'm going to do a partial revert so we can upgrade

zbaylin:
This fixes #713 by adding an `sdlDelegate` attribute to `ReveryAppDelegate` and calling it's `applicationDidLaunch` method (among others) when `ReveryAppDelegate`'s is called.

@bryphe @CrossR this works on my machine, but please verify it on yours! I don't want the same mistake to happen again.

My apologies that this is necessary in the first place.

### DIFF
--- a/src/Native/Revery_Native.c
+++ b/src/Native/Revery_Native.c
@@ -1,8 +1,5 @@
 #include <stdio.h>
 
-#include <caml/alloc.h>
-#include <caml/callback.h>
-#include <caml/memory.h>
 #include <caml/mlvalues.h>
 #include <string.h>
 
@@ -20,9 +17,10 @@
 
 CAMLprim value revery_initialize() {
 #ifdef __APPLE__
+    SDLAppDelegate *sdlDelegate = [NSApp delegate];
 // TODO: Bring back when #713 is fixed
-//    ReveryAppDelegate *delegate = [ReveryAppDelegate new];
-//    [NSApp setDelegate:delegate];
+    ReveryAppDelegate *delegate = [ReveryAppDelegate newWithSDLDelegate:sdlDelegate];
+    [NSApp setDelegate:delegate];
 #elif WIN32
     HRESULT hr = CoInitialize(NULL);
     if (hr != S_OK) {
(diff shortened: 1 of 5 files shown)


src/Native/Revery_Native.c
```diff
@@ -20,9 +17,10 @@
 
 CAMLprim value revery_initialize() {
 #ifdef __APPLE__
+    SDLAppDelegate *sdlDelegate = [NSApp delegate];
 // TODO: Bring back when #713 is fixed
-//    ReveryAppDelegate *delegate = [ReveryAppDelegate new];
-//    [NSApp setDelegate:delegate];
+    ReveryAppDelegate *delegate = [ReveryAppDelegate newWithSDLDelegate:sdlDelegate];
```
bryphe: Nice find @zbaylin ! Thanks for the investigation & fix 👍

src/Native/notification.c
```diff
@@ -29,8 +29,7 @@ CAMLprim value revery_dispatchNotification(value vNotificationT) {
     value onClickCaml = Field(vNotificationT, 2);
     mute = Int_val(Field(vNotificationT, 3));
 #ifdef __APPLE__
-    // TODO: Bring back when focus bug is fixed!
-    // revery_dispatchNotification_cocoa(title, body, onClickCaml, mute);
+    revery_dispatchNotification_cocoa(title, body, onClickCaml, mute);
```
bryphe: Cool, thanks for bringing this back!

src/Native/notification.c
```diff
@@ -29,8 +29,7 @@ CAMLprim value revery_dispatchNotification(value vNotificationT) {
...
     UNUSED(title);
```
bryphe: I guess we can remove these `UNUSED`'s as well - since they are used now in the `revery_dispatchNotification_cocoa` call

PR #717, opened 2020-01-16
Add width slider to SpringExample for manual control

glennsl:
Really minor change, but good to have for testing the `setImmediately` function to make sure it works as it should, and to confirm that I've screwed up somewhere else.


bryphe:
Awesome! Thanks for adding this @glennsl

bryphe:
Looks like latest failures are due to `nasm.us` being down ☹️

PR #718, opened 2020-01-16
Add debug logging for event dispatching, and mouse events in particular

glennsl:
This adds a few useful debug log messages related to events, and mouse events in particular. It adds `ppx_deriving` as a dependency to pretty-print events.

I've also aligned the namespaces with the convention used for Oni2.

### DIFF
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -6,6 +6,8 @@ module RenderPass = Revery_Draw.RenderPass;
 
 open Revery_Math;
 
+module Log = (val Revery_Core.Log.withNamespace("Revery.UI.Node"));
+
 module ListEx = {
   let insert = (i, node, list) => {
     let rec loop = (i, before, after) =>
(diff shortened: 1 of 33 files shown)


src/UI/Node.re
```diff
@@ -294,46 +296,52 @@ class node (()) = {
...
-      | (TextEdit(_), _)
-      | (KeyDown(_), _)
-      | (KeyUp(_), _) => ()
+    Log.debugf(m =>
+      m(
+        "Received event on node %i: %s",
+        _internalId,
+        NodeEvents.show_event(evt),
+      )
+    );
```
glennsl: This one can get a bit intense, as it logs a message for every node that is visited during bubbling. It can be very useful to see which nodes actually get the events though. Ideally we'd have another log level, like "Verbose" perhaps, for this and the "render frame" messages, but in lieu of that, perhaps it would be better to disable it.

Et7f3:
windows have failed because it had a SIGSEGV https://dev.azure.com/revery-ui/revery/_build/results?buildId=3892&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&s=d654deb9-056d-50a2-1717-90c08683d50a&t=531b9b18-2697-5845-25e5-46f788fceb1b&l=20
I have relaunched it (via the GUI) but the lock are up to date.
I have pushed the new lockdir.

src/UI/Node.re
bryphe: Ah, ya, this one would be intense!

I like the idea of another log level. But I think disabling it for now makes sense.

glennsl:
This is currently blocked by #721

PR #719, opened 2020-01-17
API - Style: Bring back 'boxShadow' style property

bryphe:
This brings back the `boxShadow` property to help unblock shadows for this PR: https://github.com/onivim/oni2/pull/1211

The problem with `<BoxShadow />`, in its current form, is that it impacts both _layout_ and _events_. This is a more general blocker to implementing the 'composition components' design here: https://github.com/revery-ui/revery/issues/489

A component like `<BoxShadow />` should purely be about drawing, and not impact layout / or event handling - it's very confusing if it interferes with that!

In order to implement this properly - we'll likely have to revisit our underlying component model, and create lower level primitives representing:
- __Rendering__
- __Layout__
- __Event handling / gestures__

The main problem is that in our current `node` object hierarchy, each `node` is responsible for all three of those - and thus, any intermediate node impacts all of those systems.


examples/Boxshadow.re
```diff
@@ -10,52 +10,40 @@ let parentStyles =
...
   ];
 
-let firstBoxStyle =
+let firstShadow =
   Style.[
     backgroundColor(Colors.blue),
     position(`Relative),
     width(100),
     height(100),
+    boxShadow(
```
bryphe: FYI @glennsl - I reverted the example so we could show the `boxShadow` API in the context of styles

PR #720, opened 2020-01-18
Update .gitattributes

imbsky:


Et7f3:
Which rei file is not marked as reason interface ? and we have some .re file that are marked as c++ so I don't think it is the right fix.

imbsky:
Linguist related changes take some time. Also, basically I should have changed re and rei at the same time in the previous PR, but I forgot about that so I updated it this time. (Even if it's correct at the moment, it's not a bad idea to force it in advance for the future.)

Et7f3:
It was merged 4 days ago :eyes: (it shouldn't be that long). I will wait .re file are marked as reason before merge.

imbsky:
Just out of curiosity, which code is marked as C++?

Et7f3:
image
https://github.com/revery-ui/revery/search?l=C%2B%2B

CrossR:
I can at least confirm that when running `linguist` locally, the previous PR did fix the issue.

Before the previous PR:

image

After:

image

It must just be taking a while?

All the files that GitHub says are C++ are classed as Reason locally now.

imbsky:
This was relatively quick for a small project, but it can take a while for a project of this size.

Et7f3:
@CrossR Is there some files with wrong language or we can merge.

CrossR:
I didn't see any Reason files that were being picked up as C++, or anything other than Reason from my look over.

PR #723, opened 2020-01-18
Refactoring - Logging: Switch to Console.log for printing

bryphe:
This is a prepatory change for: https://github.com/revery-ui/reason-sdl2/pull/36

That PR addresses an issue that's been problematic on Windows for a while - https://github.com/onivim/oni2/issues/523 - the fact that a console window opens up briefly when running Revery (or Onivim 2). 

This isn't noticeable when running from the console, but when running it independently from explorer (ie, after installing it) reproduces it. It's a very jarring and poor user experience.

The fix for that is pretty simple - we just need to tell the compiler that its a GUI app and not a console application.

However, when we do that, it breaks the default `stdout`/`stdin`/`stderr` file descriptors - any call to `print_endline` or related APIs will crash with `Bad file descriptor`. Described in more detail here: https://github.com/ocaml/ocaml/issues/9252

To work around this, I've adjusted the logging strategy to use native `printfs` vs going through the OCaml-exposed file descriptors.

Since this would be a common thing that any Revery app would need to do along with printing - I've wrapped this into the `Revery.Log.enablePrinting` method, which creates a formatter for this and calls `Timber.App.enablePrinting`. Let me know if you have better ideas for this, though, @glennsl !

Once https://github.com/revery-ui/reason-sdl2/pull/36 is in - on Windows, we'll also want to try running `AttachConsole` when logging is enabled, and if that fails, we'll `AllocConsole` so that there is a console window available somewhere (whether an existing session, or worst case, we'll spawn a new one)

__TODO:__
- [ ] Only use `Console.log` formatter on Windows

### DIFF
--- a/src/Core/Log.re
+++ b/src/Core/Log.re
@@ -3,12 +3,16 @@ include Timber;
 let enablePrinting = () => {
   // Additional utilities for logging in Revery
   let buffer = Buffer.create(0);
-  let formatter = Format.make_formatter(Buffer.add_substring(buffer), () => {
-  // We use `Console.log` instead of `print_endline` / default formatter to work around:
-  // https://github.com/ocaml/ocaml/issues/9252
-    Console.log(Buffer.contents(buffer));
-    Buffer.clear(buffer);
-  });
+  let formatter =
+    Format.make_formatter(
+      Buffer.add_substring(buffer),
+      () => {
+        // We use `Console.log` instead of `print_endline` / default formatter to work around:
+        // https://github.com/ocaml/ocaml/issues/9252
+        Console.log(Buffer.contents(buffer));
+        Buffer.clear(buffer);
+      },
+    );
 
   Logs.format_reporter(~app=formatter, ~dst=formatter, ())
   |> Logs.set_reporter;
@@ -17,4 +21,3 @@ let enablePrinting = () => {
 };
 
 let enableDebugLogging = Timber.App.enableDebugLogging;
-
--- a/src/Core/Log.re
+++ b/src/Core/Log.re
@@ -1,1 +1,20 @@
 include Timber;
+
+let enablePrinting = () => {
+  // Additional utilities for logging in Revery
+  let buffer = Buffer.create(0);
+  let formatter = Format.make_formatter(Buffer.add_substring(buffer), () => {
+  // We use `Console.log` instead of `print_endline` / default formatter to work around:
+  // https://github.com/ocaml/ocaml/issues/9252
+    Console.log(Buffer.contents(buffer));
+    Buffer.clear(buffer);
+  });
+
+  Logs.format_reporter(~app=formatter, ~dst=formatter, ())
+  |> Logs.set_reporter;
+
+  Timber.App.enablePrinting();
+};
+
+let enableDebugLogging = Timber.App.enableDebugLogging;
+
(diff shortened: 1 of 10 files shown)


src/Core/Log.re
```diff
@@ -1 +1,23 @@
...
+let enablePrinting = () => {
+  // Additional utilities for logging in Revery
+  let buffer = Buffer.create(0);
+  let formatter =
+    Format.make_formatter(
+      Buffer.add_substring(buffer),
+      () => {
+        // We use `Console.log` instead of `print_endline` / default formatter to work around:
+        // https://github.com/ocaml/ocaml/issues/9252
+        Console.log(Buffer.contents(buffer));
```
glennsl: Why `Console.log` instead of `Printf.printf`?  Not that it's a big dependency, but it is an unnecessary one, isn't it?

glennsl: Also, wouldn't this kill color output?

glennsl: Oh, sorry I misunderstood the ocaml issue. It's _native_ `printf` that works, not `Printf.printf` 🤦‍♂

bryphe: Ah exactly, sorry that wasn't clear from the description!

Ya, unfortunately `Printf.printf` exhibits the same crash as `print_endline` (anything that uses the OCaml-runtime `stdout`will hit the crash - `Printf.printf` uses it here: https://github.com/ocaml/ocaml/blob/82d7971b4f3a18dbc6dcf5d228d2f4c2d3a66fb2/stdlib/printf.ml#L29)

But `Console.log` uses the native-c `printf`, which works OK after allocing/attaching a new console.

glennsl:
Wouldn't this kill color output? ANd if so, should we perhaps then just do this for Win32?

src/Core/Log.re
bryphe: It would be nice to stick with the OCaml `stdout` though! I worry about getting tanked by a dependency that uses it, and not realizing it. Hopefully we'll get some ideas here: https://github.com/ocaml/ocaml/issues/9252

bryphe:
> Wouldn't this kill color output? ANd if so, should we perhaps then just do this for Win32?

I actually still get colored output. As far as I know, it should work - the formatting / color codes still get passed into `printf` AFAIK (The stub is pretty simple: https://github.com/facebookexperimental/reason-native/blob/9d206353299385156cd790dc427314e8d8d2389e/src/console/nativeChannels.c#L21)

But if you wouldn't mind testing on your terminal just to make sure - that would be great! I still see the colors on iTerm2 on OSX, at least

glennsl:
Alright, if it works on OSX I trust it'll work on Linux as well.

I don't fully understand this code yet, but the call used to set up colors references `stdout` and `stderr`: 

https://github.com/dbuenzli/fmt/blob/master/src/fmt_tty.ml#L60-L61

These seem to be used as a key to store the `style_renderer`, so I would think using a different output channel would also result in the `style_renderer` not being used, but I guess there's some hackery and magic involved that I probably won't understand without significant effort. If it works, it works!

bryphe:
> Alright, if it works on OSX I trust it'll work on Linux as well.

I checked on my CentOS 7 machine and still had the colors there, too!

> I don't fully understand this code yet, but the call used to set up colors references stdout and stderr:

Oh good point, you're right. Not sure why it works, but maybe it would be better to only use that formatter on Windows then, just to be safe.

glennsl:
I've tried adding this to timber, but that results in no colors and a blank line in between each message. Am I doing anything wrong here?

https://github.com/glennsl/timber/compare/bugfix/logging/console-log

glennsl:
Fixed in https://github.com/glennsl/timber/pull/2

bryphe:
Better fix in #724 - thanks @glennsl !

PR #729, opened 2020-01-22
Fix box shadow not taking color alpha value into account

glennsl:


bryphe:
I'll pull this in now, since I'm upgrading Revery anyway-

PR #733, opened 2020-01-25
API - Window: Add additional methods; refactor to interface

bryphe:
This adds some additional API methods to properly implement https://github.com/onivim/oni2/pull/988

In addition, I've refactored the `Window` module to an interface, and cleaned up some of the events (expose them via methods instead of `Event.t`s off the record)


src/Core/Window.re
```diff
@@ -4,6 +4,15 @@ type windowRenderCallback = unit => unit;
...
 type windowCanQuitCallback = unit => bool;
 
+type callback = unit => unit;
+type mouseMoveCallback = mouseMoveEvent => unit;
+type keyEventCallback = Key.KeyEvent.t => unit;
+type mouseButtonCallback = mouseButtonEvent => unit;
+type mouseWheelCallback = mouseWheelEvent => unit;
+type textEditCallback = textEditEvent => unit;
+type textInputCallback = textInputEvent => unit;
+type unsubscribe = unit => unit;
```
glennsl: I would really prefer to not have these, as they just obscure the actual shape of the function. `mouseMoveEvent => unit` is an obvious callback for the mouse move event, while `mouseMoveCallback` is an opaque blob that I have to go through hoops to look up in order to find the actual argument type (assuming it's a single argument to `unit` function in the first place, which it might not be).

What's the benefit of having these?

src/Core/Window.re
```diff
@@ -508,7 +552,7 @@ let center = (w: t) => {
   Sdl2.Window.center(w.sdlWindow);
 };
 
-let toString = w => {
+let show = w => {
   Sdl2.Window.show(w.sdlWindow);
 };
```
glennsl: Are we going back to `show` again? It's not long since we converted to `toString`.

src/Core/Window.re
bryphe: That makes sense - you're right, I think it's cleaner to have these just directly as arguments.

The only one I left is `unsubscribe` to communicate some semantics about how that `unit => unit` function is intended to be used.

src/Core/Window.re
bryphe: This method isn't actually a `toString` (`t => string`) method - it's actually an effectful (`unit => unit` method) that 'shows' the window, aka, the inverse of `hide`. It's a [thin wrapper](https://github.com/revery-ui/reason-sdl2/blob/951388b82254c63319fa0bea430e5e8feb6920e5/src/sdl2.re#L137) around the [SDL_ShowWindow](https://wiki.libsdl.org/SDL_ShowWindow) method.

bryphe: It looks like it was converted from `show` -> `toString` in https://github.com/revery-ui/revery/pull/639 (along with all the other `show`s!). Having the type signature available would've made this more obvious, because it wasn't obvious just looking at `Window.re` (The `reason-sdl2` module has confusion around `show` as well...)

bryphe: But I think this is a good example of a case where `toString` is semantically clearer

glennsl: Oh, yeah I see now that I've put on my glasses. Nice catch!

PR #738, opened 2020-01-29
add UI.element and UI.renderFunction type aliases

glennsl:
`element` is useful as a short-hand when you add interfaces to modules that export components.

`renderFunction` is what `Ui.start` returns, but the type wasn't exposed, so couldn't actually be referenced..


bryphe:
I'll bring this in prior to merging Skia in - so we can pick up this commit in Onivim 👍

PR #741, opened 2020-01-30
Dependency: Upgrade reason-harfbuzz

bryphe:
This prunes the link flags brought in by `reason-harfbuzz`, and fixes a bug where images wouldn't load on linux builds.


zbaylin:
Oops! I didn't see this before I made #742. Hopefully this is sufficient to close that!

bryphe:
@zbaylin - me too! 🤞 Would you mind testing this out and see if it fixes it?

Unfortunately I can't reproduce, so this is really just a guess.

zbaylin:
Unfortunately this didn't fix it. I still get the linking error. I can post a more detailed one, but I don't want to spam Github issues.

bryphe:
Fixed in #758

PR #743, opened 2020-01-30
Perf: Skip matrix allocation + multiply (preConcat) if no transforms

bryphe:
A hot path for Revery performance, besides the core rendering, is the `recalculation` step:
https://github.com/revery-ui/revery/blob/395ed2ed531ddc9c8632cb299067e6ce8b87cf43/src/UI/Node.re#L241

This is where we compute various aspects of the `node`, in preparation for rendering and hit testing:
- `transform`
- `worldTransform`
- `boundingBox`
- `boundingBoxClipped`

These tend to be expensive (because matrix multiplication is expensive), and allocations are also expensive when they happen a lot - reducing allocations is keep to keeping garbage collection at bay!

One inefficiency that was occurring in this phase is that, when we were computing the animation transforms - the `style=Style.[transforms(....)]` property - we were always allocating a matrix and doing a multiply, but most of the time this would be the identity matrix, and would be a no-op.

In the common case, where there are no transforms specified, we should simply do nothing and avoid the allocation + matrix multiplication.

### DIFF
--- a/src/UI/Node.re
+++ b/src/UI/Node.re
@@ -215,7 +215,7 @@ class node (()) = {
     let bbox = BoundingBox2d.transform(b, worldTransform);
     bbox;
   };
-  pri _recalculateBoundingBoxClipped = (bbox) => {
+  pri _recalculateBoundingBoxClipped = bbox => {
     switch (_this#getParent()) {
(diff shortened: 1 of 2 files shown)


bryphe:
Testing in Onivim 2, this shaves about 1ms / 20% off the `recalculate` time (from ~6ms -> ~5ms)

PR #760, opened 2020-02-06
Fix #759: Use Skia.Gr.Gl.Interface.makeNative by default

### LINKED ISSUE #759: Crash when setting up render context on Arch/Xorg
With help from @CrossR we've pinpointed it to having been caused by #751, which fixed the same kind of crash on Wayland. Unfortunately this fix now breaks Arch/Xorg. Reverting to `let context = Skia.Gr.Context.makeGl(None);` resolves the crash.

**Supplementary information:**

From running `Oni2_editor` with `valigrind -v` I get this stack trace:

```
==19956== 2 errors in context 1 of 1:
==19956== Invalid read of size 1
==19956==    at 0xA29153: eat_space_sep_strings (GrGLExtensions.cpp:54)
==19956==    by 0xA29153: GrGLExtensions::init(GrGLStandard, GrGLFunction<unsigned char const* (unsigned int)>, GrGLFunction<unsigned char const* (unsigned int, unsigned int)>, GrGLFunction<void (unsigned int, int*)>, GrGLFunction<char const* (void*, int)>, void*) (GrGLExtensions.cpp:111)
==19956==    by 0xA14015: GrGLMakeAssembledGLInterface(void*, void (*(*)(void*, char const*))()) (GrGLAssembleInterface.cpp:76)
==19956==    by 0x898241: gr_glinterface_assemble_gl_interface (gr_context.cpp:92)
==19956==    by 0x892FDA: skia_wrapped_stub_82_reason_skia_make_sdl2_interface (skia_generated_stubs.c:717)
==19956==    by 0x6E0E76: camlSkiaWrapped__Skia_generated_stubs__fun_2721 (in /home/glennsl/dev/remote/glennsl/oni2/_esy/default/store/b/oni2-f405637c/default/src/bin_editor/Oni2_editor.exe)
==19956==    by 0x753F9CF: ???
==19956==    by 0x68C8E2: camlRevery_Draw__CanvasContext__create_219 (in /home/glennsl/dev/remote/glennsl/oni2/_esy/default/store/b/oni2-f405637c/default/src/bin_editor/Oni2_editor.exe)
==19956==    by 0x752DD9F: ???
==19956==    by 0x75340EF: ???
==19956==    by 0x7465737: ???
==19956==    by 0x73F3167: ???
==19956==    by 0x702B0C7: ???
==19956==  Address 0xfffffffffffffd78 is not stack'd, malloc'd or (recently) free'd
```

bryphe:
...and fallback to `makeSdl2` only if that fails

Picks up https://github.com/revery-ui/reason-skia/pull/51

Hopefully fixes #759

### DIFF
--- a/src/Draw/CanvasContext.re
+++ b/src/Draw/CanvasContext.re
@@ -30,10 +30,10 @@ let create = (window: Revery_Core.Window.t) => {
   let interface =
     switch (Skia.Gr.Gl.Interface.makeNative()) {
     | None =>
-      Log.info("Unable to create native interface. Falling back to SDL2...");
+      Log.trace("Unable to create native interface. Falling back to SDL2...");
       Skia.Gr.Gl.Interface.makeSdl2();
     | Some(_) as nativeInterface =>
-      Log.info("Native interface created successfully.");
+      Log.trace("Native interface created successfully.");
       nativeInterface;
     };
   Log.info("Creating Skia context...");
@@ -43,7 +43,7 @@ let create = (window: Revery_Core.Window.t) => {
     Log.error("Unable to create skia context");
     None;
   | Some(glContext) =>
-    Log.info("Skia context created successfully.");
+    Log.trace("Skia context created successfully.");
     let framebufferInfo =
       Gr.Gl.FramebufferInfo.make(
         Unsigned.UInt.of_int(0),
--- a/src/Draw/CanvasContext.re
+++ b/src/Draw/CanvasContext.re
@@ -23,13 +23,27 @@ let createFromSurface = (surface: Skia.Surface.t) => {
 };
 
 let create = (window: Revery_Core.Window.t) => {
-  let sdlGlInterface = Skia.Gr.Gl.Interface.makeSdl2();
-  let context = Skia.Gr.Context.makeGl(Some(sdlGlInterface));
+  // Issue #759 - first, let's try to create a native context, since that is the most reliable...
+  // We'll fall back to an SDL2 context if not available (ie, in Wayland)
+  // TODO: There is still something busted with the way GL is being setup, for the SDL strategy not to work!
+  // Likely a fix or change is required here: https://github.com/revery-ui/reason-sdl2/blob/94dcd9094534c693998984fd684c642b0f658a43/src/sdl2_wrapper.cpp#L1065
+  let interface =
+    switch (Skia.Gr.Gl.Interface.makeNative()) {
+    | None =>
+      Log.info("Unable to create native interface. Falling back to SDL2...");
+      Skia.Gr.Gl.Interface.makeSdl2();
+    | Some(_) as nativeInterface =>
+      Log.info("Native interface created successfully.");
+      nativeInterface;
+    };
+  Log.info("Creating Skia context...");
+  let context = Skia.Gr.Context.makeGl(interface);
   switch (context) {
   | None =>
     Log.error("Unable to create skia context");
     None;
   | Some(glContext) =>
+    Log.info("Skia context created successfully.");
     let framebufferInfo =
       Gr.Gl.FramebufferInfo.make(
         Unsigned.UInt.of_int(0),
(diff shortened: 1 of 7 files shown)


src/Draw/CanvasContext.re
```diff
@@ -23,13 +23,27 @@ let createFromSurface = (surface: Skia.Surface.t) => {
...
+    switch (Skia.Gr.Gl.Interface.makeNative()) {
+    | None =>
+      Log.info("Unable to create native interface. Falling back to SDL2...");
+      Skia.Gr.Gl.Interface.makeSdl2();
+    | Some(_) as nativeInterface =>
+      Log.info("Native interface created successfully.");
+      nativeInterface;
+    };
+  Log.info("Creating Skia context...");
+  let context = Skia.Gr.Context.makeGl(interface);
```
glennsl: This would be much nicer with the new binding operators in 4.08 I think. There's no pre-defined operators coming out of the box though, so we'd have to make our own.

src/Draw/CanvasContext.re
```diff
@@ -23,13 +23,27 @@ let createFromSurface = (surface: Skia.Surface.t) => {
...
   switch (context) {
   | None =>
     Log.error("Unable to create skia context");
     None;
   | Some(glContext) =>
+    Log.info("Skia context created successfully.");
```
glennsl: These log messages are more debug-like. This isn't actionable information for normal users as they typically don't have any understanding of Skia, native and SDL2 context, and wouldn't be able to do anything with it even if they did.

bryphe: Moved them to `Log.trace`

src/Draw/CanvasContext.re
bryphe: Totally, I miss our `Option.bind` and other helper functions (like `OptionEx.or_`, which would be useful for implementing the fallback to sdl2 here...). And I imagine it could be even nicer with the binding operators.

I'll hold off on this for now so we can get this in. A few things we'll have to consider:
- Should we create a common place to put bind operators and utilities that we use across Onivim / Revery? Or duplicate them?
- Are all the sandboxed projects set up to use OCaml 4.08? I should update the README to specify that the minimum supported version of OCaml 4.08 once we incorporate the binding operators.

src/Draw/CanvasContext.re
glennsl: These will only be logged once, won't they? So I'd think `debug` would suffice. No big deal though.

src/Draw/CanvasContext.re
glennsl: > Should we create a common place to put bind operators and utilities that we use across Onivim / Revery? Or duplicate them?

Or consider using something that already exists for most of these things, like base or ocaml-containers?

> Are all the sandboxed projects set up to use OCaml 4.08?

I've lost track. but I've at least patched quite a few of them to support 4.08, and presumably used 4.08 as a dev dependency to make sure they worked. The only one I "demoted" to 4.07 was Revery, because we kept using 4.08 features for it that broke Oni.

bryphe: Using a pre-existing lib would be ideal, for sure! Do you have a preference regarding `base` or `ocaml-containers`?

glennsl: No, I haven't looked into any of them in particular detail. `ocaml-containers` is probably less opinionated, and possibly lighter. But `base` is needed for a few ppxes we might want, and so is a "free" dev-time dependency in that case. But Jane Street has opinions. Lots of opinions. But also really good engineers and business-need for high performance code.

bryphe: Good points. Reading through some of the `base` documentation:
https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/Hash/index.html

I like the emphasis on performance (and we already have it is a dep) - and of course their business-need is really geared towards perf! So I'm leaning towards exploring that.

glennsl: It's very invasive though. More of a substitute than an extension, and if used as recommended with `open Base` everywhere, will break a lot of existing code. I think containers is easier to ease into, and out of, but if Base is already there we might as well try it out. Is it a dependency of revery too, or just oni?

The Jane Street ecosystem also uses Async instead of Lwt. I don't think using Base with Lwt will conflict, but it may have synergies with Async.

PR #766, opened 2020-02-09
[WIP/POC] Libscroll Integration

szbergeron:
Creating to track feedback on integration approaches

Might be a little more complicated than I thought, since each Revery scrollview needs it's own corresponding tracking object from libscroll to avoid bugs akin to [this](https://bugs.freedesktop.org/show_bug.cgi?id=38909).

Usage follows 2 phases (pipelined):
1. Open for events
2. Locked for render

Most sane way I can think to do this is to add a Revery event that marks end of frame (signals everything to move from phase 1 to phase 2), which each scrollview itself subscribes to. Pan events are not directly used to transform the viewport, but are simply passed to libscroll scrollview instances. When an end-of-frame event comes in, the scrollview is told to mark end-of-frame and then the resulting transform directive is used much as how scroll events are handled now.

I'm a bit shaky on how some areas of Revery fit together so if any part of this sounds dumb just let me know :)


szbergeron:
This would also want wheelEvents to be a little bit smarter than they are now, or separate events between precise pan and normal wheel events (would probably make incremental integration a bit easier too)

src/Core/dune
```diff
@@ -5,4 +5,4 @@
     (c_names file)
     (c_flags :standard -Wall -Wextra -Werror)
     (preprocess (pps ppx_deriving.show))
-    (libraries threads console.lib str lwt sdl2 skia flex Rench re Revery_Native revery-text-wrap timber))
```
glennsl: You're taking out `skia` and putting `fontkit` back?

szbergeron: @glennsl whoops, not intentionally. I think that file got left behind as I was jumping between branches

glennsl:
You also need to run `./update-lockfiles.sh` to fix the CI.

Is there anything else you want reviewed here?

glennsl:
I guess the plan, but that sounds like @bryphe territory to me :)

szbergeron:
Just wanted to make sure the plan for actually integrating it with scrollviews isn't blatantly stupid

glennsl:
>  to avoid bugs akin to this.

That bug report was resolved with the conclusion that it only applies to "old-style" events, not "new-style" events. Is this integration using old-style events? If so, why? That model does seem pretty broken.

szbergeron:
> > to avoid bugs akin to this.
> 
> That bug report was resolved with the conclusion that it only applies to "old-style" events, not "new-style" events. Is this integration using old-style events? If so, why? That model does seem pretty broken.

That bug was caused by the driver emitting scroll events for "fling", basically where if you start a fling then move the cursor over some other scrollview the first one will stop and the new one will spontaneously start moving. My original plan was to simply add libscroll at the top level and integrate with the render loop, but if it ignores what scrollview the cursor is currently over then a similar bug would occur

glennsl:
Oh, I see, so that bug is just in reference to needing separate tracking objects? So why is the two-phase approach necessary?

szbergeron:
The two phase approach allows interpolating/extrapolating events properly (avoids jitter from misaligned or conflicting input/output refresh rates)

It would theoretically be possible to do simply with timers within libscroll, but that seemed a little janky/unecessary (makes usage a lot less predictable)

It basically just tells libscroll "OK, I'll be asking for the translation required for the next frame", then allows calling the functions to get that translation idempotently

glennsl:
Could it not also use an external timing source, with the current time being passed in on each call?

szbergeron:
> Could it not also use an external timing source, with the current time being passed in on each call?

It definitely could, that seemed like it would expose more of the inner workings than felt convenient for consumers to use. Would that also include the prediction/overshoot period or should that be tacked on at the end of the timestamp given?

glennsl:
> It definitely could, that seemed like it would expose more of the inner workings than felt convenient for consumers to use.

The two-phased approach seemed more like exposing implementation details to me, but then I'm still not sure how that would work. Time on the other hand is an inherent part of the model.

> Would that also include the prediction/overshoot period or should that be tacked on at the end of the timestamp given?

I don't know what any of that means :) Why is prediction needed?

I'm sure you can guess by now that I haven't followed the earlier discussion all that closely. Apologies for sounding like I have no clue what you're talking about, but I mostly don't. So please don't feel like you have to explain everything in excruciating detail if you don't want to. I'm sure @bryphe will be able to give just as good feedback with much less effort needed on your end.

szbergeron:
> > It definitely could, that seemed like it would expose more of the inner workings than felt convenient for consumers to use.
> 
> The two-phased approach seemed more like exposing implementation details to me, but then I'm still not sure how that would work. Time on the other hand is an inherent part of the model.
> 
> > Would that also include the prediction/overshoot period or should that be tacked on at the end of the timestamp given?
> 
> I don't know what any of that means :) Why is prediction needed?
> 
> I'm sure you can guess by now that I haven't followed the earlier discussion all that closely. Apologies for sounding like I have no clue what you're talking about, but I mostly don't. So please don't feel like you have to explain everything in excruciating detail if you don't want to. I'm sure @bryphe will be able to give just as good feedback with much less effort needed on your end.

No worries lol, it's a surprisingly deep subject.

Doing prediction is primarily for touchscreens, but should help make trackpads feel ever so slightly snappier. This basically tries to estimate where a touch would actually be mid way through the next frame's display period, to try to keep the delta from the onscreen content to the actual touch point as low as possible (theoretically zero if scroll velocity is constant.) A nice example of this in action (as far as I can tell) is microsoft surface tablets. From what I can see, they do some of this prediction to keep them as aligned as possible with the given touch point. Android doesn't, which leads to spacial lag (even if the actual temporal lag is imperceptible)

The two phase thing is ideally to allow users to do a minimum of their own math, just say how long frames usually last, how long until the next one will be displayed on screen, and then call mark_frame() every time a render loop happens so libscroll can do its magic. I could expose a way of asking to sample(timestamp) but that would force the consumer (in this case Revery) to do all the prediction math manually and understand the mechanics behind trying to do spacial prediction

glennsl:
Ah, I see. Thanks for explaining! I think the iPad also does a great job with the Apple Pencil.

So the prediction period then is basically just the time between user action and display response? And the only variable there is render time/frame rate?

If so I still don't understand why phases are needed. If for every frame the client asks "give position at this time", can the server not assume this call happens exactly once per frame and estimate the frame rate based on that? And in the rare case that's not true, or the client can provide a better estimate, it can do so using an optional argument?

szbergeron:
> Ah, I see. Thanks for explaining! I think the iPad also does a great job with the Apple Pencil.
> 
> So the prediction period then is basically just the time between user action and display response? And the only variable there is render time/frame rate?
> 
> If so I still don't understand why phases are needed. If for every frame the client asks "give position at this time", can the server not assume this call happens exactly once per frame and estimate the frame rate based on that? And in the rare case that's not true, or the client can provide a better estimate, it can do so using an optional argument?

Good point, thinking about it I've come around to the idea of having a requested timestamp be provided by the consumer (especially around VRR cases and the like, so I can step animations by a precise delta instead of trying to speculate) for how to sample. It also probably allows for way easier testing/mocking on my part to have that be the case.

I can try getting that in place a bit more tonight probably.

Question: what would this look like inside of Revery? IE do elements in revery know when they're about to be rendered such that they could know when to sample?

glennsl:
Yeah, I think the rendering would be driven by the element itself in this case, much like how we do animation:

https://github.com/revery-ui/revery/blob/4ca3483c1f1d2d955ac46846bffc8aa54acb7c7c/src/UI_Hooks/Revery_UI_Hooks.re#L17-L19

This will continuously trigger a re-render as long as `~active` is true, and returns the time to be used in the rest of the render function. The animations in Revery are completely pure functions.

szbergeron:
> Yeah, I think the rendering would be driven by the element itself in this case, much like how we do animation:
> 
> https://github.com/revery-ui/revery/blob/4ca3483c1f1d2d955ac46846bffc8aa54acb7c7c/src/UI_Hooks/Revery_UI_Hooks.re#L17-L19
> 
> This will continuously trigger a re-render as long as `~active` is true, and returns the time to be used in the rest of the render function. The animations in Revery are completely pure functions.

Oh very cool, I'll try to look into it a bit more tonight I guess, since that example almost perfectly fits how the library itself models things (save for events/correction and such, but those are silent here)

glennsl:
> Those hooks only get called after all input events have been serviced right? Or are they called sometimes in the middle?

I'm not entirely sure, but the application loop certainly looks like it does:

https://github.com/revery-ui/revery/blob/4ca3483c1f1d2d955ac46846bffc8aa54acb7c7c/src/Core/App.re#L210-L250

src/UI_Components/ScrollView.re
```diff
@@ -180,10 +180,41 @@ let%component make =
...
             Log.info("Scrollview existed");
+
+            Libscroll.set_source(scrollview, wheelEvent.source);
+
+            switch(wheelEvent.deltaX) {
+            | Some(delta) => {
+                Libscroll.push_pan(scrollview, Libscroll.Axis.Horizontal, delta, wheelEvent.timestamp);
+              }
+            | None => ()
+            };
```
glennsl: Why is this an `option`? Wouldn't a `delta` of `0` be "neutral" and effectively do nothing anyway?

src/UI_Components/ScrollView.re
```diff
@@ -180,10 +180,41 @@ let%component make =
...
+            | Some(delta) => {
+                Libscroll.push_pan(scrollview, Libscroll.Axis.Vertical, delta, wheelEvent.timestamp);
+              }
+            | None => ()
+            };
+
+            switch(wheelEvent.isFling) {
+            | true => Libscroll.push_fling(scrollview, wheelEvent.timestamp);
+            | false => ()
+            };
```
glennsl: Don't use pattern matching for boolean conditionals. That's what `if` expressions are for. They communicate intent better, are less noisy, and also shorter in cases like this where you can skip the `else` branch:

```
            if (wheelEvent.isFling) {
              Libscroll.push_fling(scrollview, wheelEvent.timestamp);
            };
```

src/UI_Components/ScrollView.re
szbergeron: If it wasn't doing interpolation/prediction then yes, but with both enabled it would mean that acceleration goes to infinity and velocity drops to zero, which would cause weird stuttering if event ordering isn't consistent (and also make it so that fling sometimes wouldn't work)

src/UI_Components/ScrollView.re
szbergeron: Ok, will change

src/UI_Components/ScrollView.re
glennsl: Can you not just weed out the `0`s from the interpolation? Or does the `None` have some other significance?

Like, if it was not an option, would this not be equivalent?

```
            if (wheelEvent.deltaX != 0) {
                Libscroll.push_pan(scrollview, Libscroll.Axis.Horizontal, wheelEvent.deltaX, wheelEvent.timestamp);
              }
```

szbergeron: Probably, at least on Wayland I never seem to get a zero delta. I still need to figure out how to drag prediction toward zero if velocity drops but no zero event or fling event comes in.

It seems semantically odd to special case zero delta to me but it might be fine

glennsl: I find it odd that it needs to be special-cased too, but I would  consider both using an `option` and weeding out `0`s special-casing. A delta of `0` seems completely normal to me, so I'm wondering if the problem is somewhere else in the model. Unless you work with imperfect sensor data, which of course happens sometimes.

glennsl: But ideally that would still be handled by the library, not by the consumer.

szbergeron: One option would be to make them always include a delta but only for one axis at a time, this is what Wayland/libinput end up doing, rather than omitting one or the other. This also works for platforms that emit them two at a time, just emit two of those events.

glennsl: Oh, is that the problem? That you might get separate events for the x- and y-axes for the same point in time?

szbergeron: Yeah, that's what this attempts to deal with.

In retrospect, I probably should have copied the single axis model for pan events in my sdl2 change since trying to paper over it ended up just making the PanEvent api uglier

glennsl: If the timestamp is the same, you could also just pop the latest event, check if the timestamp is the same, and if it is add back the one that isn't `0`.

szbergeron: The timestamp can technically be different, the wayland spec allows the events to diverge by any amount of time so long as they stay grouped between pointer frame events. This may never really happen in practice though, not sure.

I could probably amend the sdl2 change to do single axis events, not sure if it's worth it to do that or if I should do as you suggest and try to combine events, or if I should just catch the events at the reason-sdl2|revery boundary and emit *up to* 2 events based on the containsX/containsY members

glennsl: Bummer :disappointed:

IMO whatever removes this kind of leaky abstraction that requires decision making on the consumer-side would be worth it.

szbergeron: Libscroll, sdl, or both? I can probably get the required changes for sdl done tonight, and nothing relies on the pan events yet so that wouldn't *actually* be a breaking change

glennsl: I mean from the POV of the ScrollView in this case. To be able to just have

```reason
Libscroll.push_pan(scrollview, Libscroll.Axis.Horizontal, wheelEvent.deltaX, wheelEvent.timestamp);
Libscroll.push_pan(scrollview, Libscroll.Axis.Vertical, wheelEvent.deltaY, wheelEvent.timestamp);
```

Whether the required changes are done in SDL or Libscroll is less important from this POV at least. Bu whatever is the least leaky in general would be my preference.

szbergeron: I think if sdl is changed correctly, then the event can be directly pushed (though some Sdl2.Axis will have to be mapped to Libscroll.Axis) with 

`Libscroll.push_pan(scrollview, map_axis(wheelEvent.axis), wheelEvent.delta, wheelEvent.timestamp);`

glennsl: If these were polymorphic variants, you wouldn't even need to do the mapping. It's a good use case for it.

szbergeron: Hm, how would that work? Am not super familiar with them

glennsl: Polymorphic variants are the structurally typed equivalents of ordinary variants. Their identity is derived from the type itself, not from any definition. In this case specifically that means SDL and libscroll can share a polymorphic variant type without either library knowing about the other.

By example, it looks like this:

```reason
let functionTakingPolyVariant = (poly: [`Number(int) | `Text(string)]) =>
  switch (poly) {
  | `Number(num) => "Number: " ++ string_of_int(num)
  | `Text(text) => "Text: " ++ text
  };

let str = functionTakingPolyVariant(`Number(42));
```

szbergeron: Something I'm also remembering here, is that wayland sends events for when the source axis changes for a pan, and that contains an undefined delta. I probably need a way of excluding the delta as well. Might make sense to have a more general axis enum indicating horizontal, vertical, or undefined (undefined meaning the delta is meaningless, and shouldn't be interpreted)

szbergeron: Looking at how this interacts with scroll stop events makes me want to bring back the guard, for single accesses.

Effectively what I want to encode is the variant/enum

Info(Source)
Fling(Axis, Source) |
Interrupt(Axis, Source) |
Pan(Axis, Source, delta)

in a way that is fairly ideomatic in c and translates well across to reasonml/rust code. What are your opinions on doing that nicely?

I might be overfitting for the linux way of handling input, but afaik it's pretty similar on OSX and can be fully described for the Windows model (just compose multiple events)

PR #771, opened 2020-02-17
Bugfix - Perf - FontCache: Create upper bound on memory usage

bryphe:
...and fix bug with hashing.

In searching around for the memory leak reported by @CrossR and @glennsl - saw there was an issue in our font cache here:

1) We were using `String.length` for our string hash function, which is not a good hash function...

and

2) There was no upper bound on our shaping caching, so it would just continually grow.

The first was a simple fix - just to switch to use `Hashtbl.hash` instead. For the second, I brought in the `lru` module to have a least-recently used cache, so that we could evict shape results, font metrics, and cached fonts when they are no longer used.

There is a bit of tuning required - the `lru` module pre-allocates ahead of time, so you want a value that is big enough to not 'thrash' cache usage (where keys are continually evicted and added), but you want it small enough to not pre-allocate to much memory. I tuned it a bit testing with Onivim 2 - for the shape cache, a value of `1024` caused thrashing in the scenarios I tested, but `4 * 1024` did not - I bumped it up by another factor to help in large screens.

Unfortunately, I don't think this addresses the leak that @CrossR and @glennsl found - the way we handle shaping didn't change much before and after Skia (aside from that busted hash function)... I suspect there is another underlying cause.


src/Font/FontCache.re
```diff
@@ -68,14 +68,8 @@ let load: string => result(t, string) =
...
-      let metricsCache = MetricsLruHash.create(32);
-
-      // Allow ~32KB of cached shape strings in our LRU cache.
-      // Shaping is expensive, so we want to cache it aggressively...
-      // but we also don't want to blow up memory!
-
-      // The ideal value for this is _just_ large enough that we never miss.
-      let shapeCache = ShapeResultLruHash.create(32 * 1024);
+      let metricsCache = MetricsLruHash.create(~initialSize=8, 64);
+      let shapeCache = ShapeResultLruHash.create(~initialSize=1024, 128 * 1024);
```
glennsl: There are a bunch of magic numbers here. What do they all mean? The arguments ought to be labelled and the values given names by assigning them to bindings in `Constants` for example, at least if they have different meanings than the named arguments would imply.

PR #775, opened 2020-02-19
Bugfix: Ensure timer is always disposed

bryphe:
While experimenting with a cursor animation in `revery-terminal` (https://github.com/revery-ui/revery-terminal/pull/1) - I saw there was a pretty bad bug... Sometimes the spring hook would never settle into a rest state, and continually redraw - even when the animation was completed. This was pretty easy to repro - all it would take is running a `git status` or type commands quickly, but it was intermittent.

I did some logging in the spring hook here: https://github.com/revery-ui/revery/blob/a525e7f07d1417791c6ac2b574ed7fb9e68ceaeb/src/UI_Hooks/Spring.re#L14

And saw that `isActive` was being set to `false` for both springs (the x & y springs), and the `Timer.timer` hook was getting `isActive` of `false`... but one of the timers was still ticking. So it seems like the root issue is in the `Timer.timer` hook - the spring hook is behaving correctly.

Note that this reproduced even before bringing in https://github.com/revery-ui/revery/pull/773 - I brought that in hoping it would fix it 😄 I believe that this bug has occurred in Onivim 2 as well - but less frequently. I have observed cases where we would start re-rendering continuously, and the spring hook could be the culprit.

I was able to fix this by ensuring that we always dispose a previous hook. There may  be a bug somewhere in the way cleanup is occurring in the effect hook: https://github.com/briskml/brisk-reconciler/blob/c22c693099c16a39ad115569c599ddba75cb0736/lib/Hooks.re#L237

It could even be an interplay of the timing when we flush pending effects vs when state changes.

In any case, this 'brute force' fix is to check a previous dispose and run it before we set a new cleanup handler. This seems to fix the issue, but could potentially have re-entrancy issues (although it is safe to call `dispose` an interval multiple times).

### DIFF
--- a/src/UI_Hooks/Timer.re
+++ b/src/UI_Hooks/Timer.re
@@ -12,6 +12,7 @@ let time = (~tickRate=Time.zero, ()) => {
 let timer = (~tickRate=Time.zero, ~active=true, ()) => {
   let%hook (time, setTime) = reducer(~initialState=Time.now(), t => t);
   let%hook startTime = Ref.ref(time);
+  let%hook lastDispose = Ref.ref(None);
 
   let onTick = _dt => setTime(_t => Time.now());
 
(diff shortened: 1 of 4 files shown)


src/UI_Hooks/Timer.re
```diff
@@ -12,16 +12,31 @@ let time = (~tickRate=Time.zero, ()) => {
...
     effect(OnMountAndIf((!=), active), () =>
       if (active) {
         startTime := Time.(now() - (time - startTime^));
-        let dispose = Revery_Core.Tick.interval(onTick, tickRate);
-        Some(dispose);
+
+        switch (lastDispose^) {
+        | None => ()
+        | Some(f) => f()
+        };
```
glennsl: Now that we're on 4.08 we can use `Option.iter` here as well.

bryphe:
Good point. Added detailed comment + test

PR #778, opened 2020-02-22
Add Github Actions for lint, publish doc, and build.

Et7f3:
This PR enable, GitHub Action for build, lint, auto format code, and push docs. A admin user need to enable GitHub pages.


Et7f3:
All users with write access can now write gocs and see live. For others users, they should use their fork.

PR #779, opened 2020-02-25
API - Font Measurements: Make smoothing type explicitly required

bryphe:
This is the revery-side fix for: https://github.com/onivim/oni2/issues/1364

The measurement / metrics of character width are impacted by the font smoothing settings (specifically, whether subpixel text is allowed). 

Because this is an important consideration for measurements, we should make our measurement APIs take a required argument for this - to ensure the caller has considered the smoothing type.

### DIFF
--- a/src/Font/Smoothing.re
+++ b/src/Font/Smoothing.re
@@ -1,0 +1,29 @@
+open Revery_Core;
+
+type t =
+  | None
+  | Antialiased
+  | SubpixelAntialiased;
+
+let default =
+  switch (Environment.os) {
+  | Windows => SubpixelAntialiased
+  // On OSX, default to non-subpixel for Retina displays.
+  // On Linux, subpixel does not reliably set subpixel rendering,
+  // so default to false and make it opt-in
+  | _ => Antialiased
+  };
+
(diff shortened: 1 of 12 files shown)


Et7f3:
@bryphe Do you know the the GHA also format code. So Formatting commit was useless.

bryphe:
@Et7f3 - oh, neat, didn't realize! That's very cool

src/Font/Smoothing.re
```diff
@@ -0,0 +1,29 @@
...
+let default =
+  switch (Environment.os) {
+  | Windows => SubpixelAntialiased
+  // On OSX, default to non-subpixel for Retina displays.
+  // On Linux, subpixel does not reliably set subpixel rendering,
+  // so default to false and make it opt-in
+  | _ => Antialiased
+  };
+
+let setPaint = (~smoothing: t, paint: Skia.Paint.t) => {
```
glennsl: This is backwards. `setPaint` implies that you're setting the paint property of the smoothing object, while in fact it's doing the reverse. Something like `apply` would be a better name I think.

PR #781, opened 2020-02-26
Dependency: timber -> 1.0.0

glennsl:
Finally got timber published. Tried to publish it to opam first, since I think that's where it belongs and since there's already an npm package of the same name, but I'm unable to authenticate with GitHub using 2FA for some unknown reason, so npm it is! It's now published as `@glennsl/timber`.

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "c26d11c7a764289c4d28729dbe44203d",
+  "checksum": "764431ed8b041c356f0f40af3961fa67",
   "root": "revery@link-dev:./package.json",
   "node": {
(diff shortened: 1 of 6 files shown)


glennsl:
Seems like all dependents have to be updated synchronously, because esy/dune/whatever doesn't recognize the library as being the same after adding the namespace, despite the "internal" library name (used in `dune` files) still being the same.

bryphe:
Hmm, @Et7f3 , do you happen to know why the 'Push docs' GH actions failed?

bryphe:
I'll pull this in manually to unblock some of the other downstream changes (https://github.com/revery-ui/revery-terminal/pull/12, https://github.com/revery-ui/revery-terminal/pull/10)

Et7f3:
It has failed on checkout 🤔

Et7f3:
Ah a3fd3dab9e8310464c19f64e9c8551eec30ce187 is not in revery-ui/revery now... That is annoying 🤔   but the gha "push docs" is also triggered on commit on master branch. So not issue @bryphe

PR #782, opened 2020-02-26
use action checkout 2

Et7f3:
Here is to test issue we encounter in #781


bryphe:
Thanks for your help looking into this, @Et7f3 !

Strange - doesn't look like it reproduced here

Et7f3:
Have we the same access ? glennsl and I.

PR #795, opened 2020-03-20
API - App: Add onBeforeQuit handler

bryphe:
This adds an `App.onBeforeQuit` API, to allow hooking a callback when we are for-sure quitting the application. There were some code paths where we weren't able to hook-in when quitting - this ensures that we can always run some final cleanup code prior to the application exiting.

Needed for https://github.com/onivim/oni2/issues/1485


src/Core/App.re
```diff
@@ -50,9 +46,17 @@ let quit = (~askNicely=false, ~code=0, app: t) => {
...
-
   if (Hashtbl.length(app.windows) == 0 || !askNicely) {
+    Revery_Native.uninit();
+
+    // Verify [quit] wasn't called recursively from a beforeQuit handler
+    if (!app.isQuitting) {
+      Log.info("onBeforeQuit");
+      app.isQuitting = true;
+      let _: unit = Event.dispatch(app.onBeforeQuit, ());
+      app.isQuitting = false;
```
glennsl: `let _: unit =`. is superfluous.

But also, why is `isQuitting` set back to `false` here?  Are we no longer quitting? If the purpose is just to guard against  nested calls to `onBeforeQuit`, should it perhaps be more accurately named so it isn't accidentally used in ways it shouldn't?

src/Core/App.re
```diff
@@ -248,7 +257,8 @@ let start = (~onIdle=noop, initFunc: appInitFunc) => {
 
       if (appInstance.idleCount === framesToIdle) {
         Log.debug("Downshifting into idle state...");
-        appInstance.onIdle();
+        let _: unit = Event.dispatch(appInstance.onIdle, ());
+        ();
```
glennsl: Here's this weird pattern again, caused by the unnecessary `let _: unit =`.

Issue #791, opened 2020-03-06
Logging prevents the Revery_Lwt eventLoop from working


lessp:
- Operating System: 
macOS
- Revery Version: 
0.30.0
- OCaml version: 
~4.8
- Native, Bytecode, or JS build: 
Native

`Revery_Lwt.startEventLoop()` does not work when logging is enabled. Removing:

```
Revery.App.initConsole();

Timber.App.enable();
Timber.App.setLevel(Timber.Level.perf);
```

Fixes the issue.

glennsl:
Curious. Does all three lines really need to be removed? Or would removing any of these solve it?

lessp:
> Curious. Does all three lines really need to be removed? Or would removing any of these solve it?

Yes, I'm curious as well!

This was late last night, so I might update the issue with further findings today.

lessp:
I went back and checked and it seems as though the issue only arises when enabling Timber. 🤔

PR #796, opened 2020-03-22
!feat(components): add support for remote-images

### LINKED ISSUE #537: Load image from absolute path/url

lessp:
A new PR to reflect changes from Skia.

revery-remote-image

Closes #537 

BREAKING CHANGE: 
When using the `<Image ... />`-primitive, you now have to pass a variant of either (\`Url or \`File) to `src`. 

In other words, where you previously passed `<Image src="example.png" />` you'll now have to pass 
```reason
<Image src=`File("example.png") />
``` 
or 
```reason
<Image src=`Url("https://example.com/example.png") />
```


Et7f3:
Oups GHA removed lockdir because esy can't solve dependencies.

lessp:
There's now a working example pushed, for now I created a `getContextRemote` for along with `getContext` for performance-reasons.

I'm guessing we probably want a different API, so that will have to be handled either now or in the future. E.g.
```reason
<Image src=Url("..", options) />
```

lessp:
Can you see understand why this fails @Et7f3?

```
info install 0.6.2 (using package.json)
info checking https://github.com/ocaml/opam-repository for updates...
info checking https://github.com/esy-ocaml/esy-opam-override for updates...
error: File exists
  reading package metadata from github:esy-packages/esy-openssl:package.json#b7eab4ff97f6e61ec4523e771176e3aadc25cdd8
  resolving metadata @opam/conf-libssl@github:esy-packages/esy-openssl:package.json#b7eab4ff97f6e61ec4523e771176e3aadc25cdd8
  resolving @reason-native-web/ssl@0.5.9-dev.2
  resolving @reason-native-web/piaf@1.0.1000
  resolving fetch-native-lwt@github:lessp/fetch:fetch-native-lwt.json#0774bc989493f78f87e1689731f41f2a83ea0b78
esy: exiting due to errors above

##[error]Bash exited with code '1'.
```

lessp:
Ok, so there's something weird happening:

```
info install 0.6.2 (using package.json)
info checking https://github.com/ocaml/opam-repository for updates...
info checking https://github.com/esy-ocaml/esy-opam-override for updates...
error: File exists
  reading package metadata from github:esy-packages/esy-openssl:package.json#b7eab4ff97f6e61ec4523e771176e3aadc25cdd8
  resolving metadata @opam/conf-libssl@github:esy-packages/esy-openssl:package.json#b7eab4ff97f6e61ec4523e771176e3aadc25cdd8
  resolving @reason-native-web/ssl@0.5.9-dev.2
  resolving @reason-native-web/lwt_ssl@1.1.4
  resolving @reason-native-web/piaf@1.0.1000
  resolving fetch-native-lwt@github:lessp/fetch:fetch-native-lwt.json#0774bc9
esy: exiting due to errors above
##[error]Process completed with exit code 1.
```

Did you try to run it @Et7f3? (on your Windows-machine, if yes) 🙂

Et7f3:
```
 rd /S/Q esy.lock _esy & esy i
info install 0.6.2 (using package.json)
info checking https://github.com/ocaml/opam-repository for updates...
info checking https://github.com/esy-ocaml/esy-opam-override for updates...
info resolving esy packages: done
info solving esy constraints: done
info resolving npm packages: done
.... fetching @opam/conf-libssl@archive:https://github.com/openssl/openssl/archive/OpenSSL_1_1_1d.tar.gz#sha1:df0ee4811c87c209ebadb4e6                                                                                                                                      info fetching: done
.... installing @opam/conf-libssl@archive:https://github.com/openssl/openssl/archive/OpenSSL_1_1_1d.tar.gz#sha1:df0ee4811c87c209ebadb4                                                                                                                                      info installing: done
```

Work ok on my computer it manage to solve. I try to build.

lessp:
Seems like other people are hitting this issue as well, will follow:

https://discordapp.com/channels/436568060288172042/552911588038934531/694566761294725190

.github/workflows/validate-pr.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         run: npm install -g esy
       - run: brew install coreutils
       - name: Solve doc dependencies
-        run: esy @doc install
+        run: esy @doc installebug
```
Et7f3, line 81:
```suggestion
        run: esy @doc installe
``` ?

File: examples/Examples.re [hunk 2]
```diff
@@ -247,8 +247,8 @@ module ExampleHost = {
 let init = app => {
   Revery.App.initConsole();
 
-  Timber.App.enable();
-  Timber.App.setLevel(Timber.Level.perf);
+  /* Timber.App.enable(); */
+  /* Timber.App.setLevel(Timber.Level.perf); */
```
Et7f3, line 251: What is the issue with timber ?

package.json
```diff
@@ -27,19 +27,24 @@
...
+    ],
     "install": [
       "esy-installer Revery.install",
       "bash -c \"#{os == 'windows' ? 'cp /usr/x86_64-w64-mingw32/sys-root/mingw/bin/*.dll \\'$cur__bin\\'': ':'}\""
     ]
   },
   "dependencies": {
     "@opam/dune-configurator": "*",
-    "@esy-ocaml/reason": "^3.4.0",
+    "@esy-ocaml/reason": "3.6.0",
```
Et7f3, line 40: use ^ or > same fo r bos, fetch-native-lwt

examples/Examples.re
lessp: Not sure, but I've logged an issue: #791

lessp:
Do you know why we're hitting this @Et7f3? 

https://dev.azure.com/revery-ui/revery/_build/results?buildId=5777&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=038b32d5-001e-59e3-852f-ba0d70a2f5a1&l=26

Et7f3:
Caused by #844: short term fix remove buildDev in package.json or specify to build RenchTest package. Long term fix https://github.com/esy/esy/pull/1100 which will allow to have a buildDev In both sandbox.

lessp:
> Caused by #844: short term fix remove buildDev in package.json or specify to build RenchTest package. Long term fix [esy/esy#1100](https://github.com/esy/esy/pull/1100) which will allow to have a buildDev In both sandbox.

Ah, I see. Thanks for the quick help!

lessp:
As for the API I'm thinking `Image.Url`, `Image.Asset` and `Image.Path` or `Image src=`Asset/`Path/`Uri`.

With auto-complete using nested modules might be easier to find and allows for path-specific props, so might be better. Any thoughts?

lessp:
Ok, apart from not working when the logs are enabled (any ideas on that are appreciated too), I think this actually works. 🙂

Interested to hear your thoughts!

Et7f3:
Why such huge diff ? Your change get lost in all the diff.

lessp:
> Why such huge diff ? Your change get lost in all the diff.

This was opened a while ago, and I've used `merge` to keep up with changes from master now and then. This should help:

https://github.com/revery-ui/revery/pull/796/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.re&file-filters%5B%5D=.rei&file-filters%5B%5D=.sh&file-filters%5B%5D=No+extension&hide-deleted-files=true&file-filters%5B%5D=.install&file-filters%5B%5D=.patch&file-filters%5B%5D=.seq&file-filters%5B%5D=dotfile

src/Draw/ImageRenderer.re
```diff
@@ -1,30 +1,169 @@
...
+      | "image/jpg" => ".jpg"
+      | "image/png" => ".png"
+      | "image/svg+xml" => ".svg"
+      | "image/tiff" => ".tif"
+      | "image/webp" => ".webp"
+      | _ => ".jpg";
+
+    module Text = {
+      let toAlphaNumeric = text =>
+        Str.(global_replace(regexp("[^a-zA-Z0-9_]"), "", text));
```
Et7f3, line 60: Str has partial support on jsoo. Have you checked if it is supported ?

src/UI_Primitives/Image.rei
```diff
@@ -0,0 +1,57 @@
+/**
+ * Image.Asset
+ *
+ * Used for asset images.
+ *
+ * Examples:
+ *   <Image.Asset src="example.png" opacity=1.0 width=64 height=64 />
+ */
+module Asset: {
```
Et7f3, line 9: Maybe File (in another PR this change was suggested because asset might bundled with the exec or in memory)

src/UI_Primitives/Image.re
```diff
@@ -17,65 +17,132 @@ let getStyles: (option(int), option(int), Style.t) => Style.t =
...
-          ~onMouseWheel?,
-          (),
-        );
-      let node = PrimitiveNodeFactory.get().createImageNode(src);
-      node#setOpacity(opacity);
-      node#setEvents(events);
-      node#setStyle(styles);
-      node#setResizeMode(resizeMode);
-      Obj.magic(node);
+module Asset = {
```
Et7f3, line 20: Lot of code duplication can you take src to differ fromFile/fromNetwork ?

lessp: You mean as in:

```reason
<Image src=`Url("...") />
<Image src=`File("...") />
```

or try parsing the src as a url and then deciding? 
```reason
isUrl ? `Url(src) : `File(src)
```

src/UI_Primitives/Image.rei
lessp: Yes, that would probably be a better name. I was a bit confused, I thought we were going to need three variations, one for dune-assets, one for file-paths and one for network-requests, but it seems as though we only need 2?

src/Draw/ImageRenderer.re
lessp: Looks like it's supported: 

<https://ocsigen.org/js_of_ocaml/3.6.0/manual/overview>
<https://github.com/ocsigen/js_of_ocaml/search?q=global_replace&unscoped_q=global_replace>

src/Draw/ImageRenderer.re
```diff
@@ -1,30 +1,169 @@
...
+          |> List.find_opt(((k, _v)) =>
+               String.lowercase_ascii(k) == "content-type"
+             )
+          |> Option.map(((_k, v)) => v)
+          |> Option.value(~default="");
+
+        Lwt.return(Ok({data, mediaType}));
+      }
+    );
+  };
```
glennsl, line 41: This is also the wrong place to do I/O I think. A renderer should render, not do do all kinds of other things as well. I can see how the earlier decision to have `fromAssetPath` here lead to this, but adding network I/O and asynchrony makes the spaghetti bowl significantly deeper. I think this needs a proper refactoring to separate file I/O, asset handling, network access and rendering.

lessp: Yes, that probably makes sense! Any pointers on where you'd start? `drawImage` takes a `Skia.Image.t` and that we handle IO in the component? Would that mean that we need to create a wrapper-`%component` for the Image-primitive perhaps?

lessp: ... or would ImageNode work?

lessp: Maybe to begin with, I could start to put all the IO-specific things in a separate module and let `drawImage` take:

```
let drawImage = (~x, ~y, ~width, ~height, ~paint=?, data: Skia.Image.t, v: t) => {
  Canvas.drawImageRect(
    v.canvas,
    data,
    None,
    Rect.makeLtrb(x, y, x +. width, y +. height),
    paint,
  );
};
```

glennsl: Yeah, that seems like a  good start. Not sure what you mean with the wrapped component, but I don;t think the component API needs or ought to change.

lessp: I've pushed some progress, let me know if you think this is the right direction and what else we could do to tidy things up. Really appreciate the feedback @glennsl! 👌

package.json
```diff
@@ -27,41 +27,53 @@
...
+    "rench": "^1.9.1",
+    "reperf": "^1.4.0",
+    "revery-text-wrap": "*"
   },
   "resolutions": {
+    "revery-text-wrap": "link:../revery-text-wrap/package.json",
     "@esy-ocaml/libffi": "esy-ocaml/libffi#c61127d",
     "esy-cmake": "prometheansacrifice/esy-cmake#2a47392def755",
-    "@opam/cmdliner": "1.0.2"
+    "@opam/cmdliner": "1.0.2",
```
Et7f3: This PR is very old :eyes: We haven't that now.

src/Draw/ImageRenderer.re
glennsl: Looks great! I think this already addresses the bulk of my concerns. It would be nice to have the asset handling and networking separated out and to have proper interface files for the new modules. I'm also not sure about the idea of `Image` living in `IO`, but that's more of a cognitive dissonance issue than an actual problem I think.

Good job and thanks for doing this!

lessp: Cool, I'll give another round soon just to tie things up. Thanks!

lessp: I added some interface-files and left a note that the `Image`-module residing in `IO` could be moved and/or have another API-surface. Agree with you on that! 

After your feedback, I think this at least leaves it in a little better state than previously while still having room for improvements!

As for this PR to not have to compete against time and getting stale, I'm thinking we can remove the use of it in the examples (as well as the call to `startEventLoop`), in order to enable `Timber` again, but still get the functionality in. What do you think?

src/IO/Image.rei
```diff
@@ -0,0 +1,20 @@
+/**
+ * fromUrl
+ *
+ * Given a network file-path this returns a promise,
+ * holding either a Ok(option(Skia.Image.t)) or an Error(string).
+ *
+ * Examples:
+ * let result = Image.fromUrl("https://example.com/hello.png");
+ */
+let fromUrl: string => Lwt.t(result(option(Skia.Image.t), string));
```
glennsl, line 10: Why an `option` inside a `result`? What does `Ok(None)` mean?

src/Draw/ImageRenderer.re
glennsl: Yeah, I think that's a good idea if this can't be solved now. If possible, it might be a good idea to try to run `Lwt` the same way we run `Luv` in Oni2, polling on every tick so we're still in full control of the event loop:

https://github.com/onivim/oni2/blob/f5094e83c944108bd8fea9d0a7744ff3e5a33912/src/bin_editor/Oni2_editor.re#L208-L228

> I added some interface-files and left a note that the Image-module residing in IO could be moved and/or have another API-surface. Agree with you on that!

Good call. Thanks for adding the interfaces and comments! I find them very helpful.

src/IO/Image.rei
lessp: Thanks, changed it in: 07edf87

lessp:
I saw that you merged it @glennsl, but I think it's fine, the conclusions I came to are above 🙂

glennsl:
Ah, sorry, that was terrible timing. 

> Images downloaded from online would probably have an option for folder to cache in, delete/keep files after download, caching etc while these would not apply for images loaded from disk since they're already there. I don't think settings like these make a lot of sense as props, but better put in a record.

This seems to me like an application-wide setting, not one that would need to be different for each component instance. Or to have the flexibility of having some instances with different settings, you could provide a functor interface for example.

> Does images over network and file-images take the same props, I think they could, right?

For any optional props that aren't outright contradictory for file-images, I think that's absolutely fine.

lessp:
No worries, just hit me while discussing a bit with Zach!

> This seems to me like an application-wide setting, not one that would need to be different for each component instance. Or to have the flexibility of having some instances with different settings, you could provide a functor interface for example.

Yep, that's true!

PR #804, opened 2020-04-01
API - Hooks - Transition: More declarative API

bryphe:
This removes the 'explicit' imperative setter for transition, and replaces it with an 'implicit' set - if the target value changes that is passed in, the transition will be triggered.

This made switching from the `Hooks.spring` -> `Hooks.transition` more ergonomic in this PR: https://github.com/onivim/oni2/pull/1518

### DIFF
--- a/src/UI_Hooks/Revery_UI_Hooks.re
+++ b/src/UI_Hooks/Revery_UI_Hooks.re
@@ -33,15 +33,14 @@ let animation = (~active=true, ~onComplete=() => (), animation) => {
   (value, animationState, reset);
 };
 
-
 module Internal = {
   type durationCalculation =
-  | Constant(Time.t)
-  | Function({
-    initial: Time.t,
-    calculated: (~current: float, ~target: float) => Time.t,
-    });
-  
+    | Constant(Time.t)
+    | Function({
+        initial: Time.t,
+        calculated: (~current: float, ~target: float) => Time.t,
+      });
+
   let transition =
       (
         ~active=true,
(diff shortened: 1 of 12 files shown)


src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -36,27 +36,52 @@ let animation = (~active=true, ~onComplete=() => (), animation) => {
...
     ) => {
-  let%hook ((startValue, targetValue), setTargetValue) =
-    state((startValue, startValue));
+  let initialValue =
+    switch (initialValue) {
+    | Some(v) => v
+    | None => targetValue
+    };
+
+  let specifiedTargetValue = targetValue;
```
glennsl, line 52: Why not just rename the `targetValue` function argument to specifiedTargetValue`?

src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -36,27 +36,52 @@ let animation = (~active=true, ~onComplete=() => (), animation) => {
 let transition =
     (
       ~duration=Time.seconds(1),
+      ~durationFunc=?,
```
glennsl, line 40: This API is pretty awkward. These should be mutually exclusive, or ideally perhaps would only accept the function but with some convenient way of specifying a constant function.

You could get the best of both worlds by removing the optional function argument from this, and adding another hook that takes the function instead. They can both call a common implementation with a more cumbersome API.

bryphe: Sounds good! Thanks for the suggestion, @glennsl

src/UI_Hooks/Revery_UI_Hooks.re
```diff
@@ -33,30 +33,109 @@ let animation = (~active=true, ~onComplete=() => (), animation) => {
   (value, animationState, reset);
 };
 
+module Internal = {
+  type durationCalculation =
+    | Constant(Time.t)
+    | Function({
+        initial: Time.t,
```
glennsl, line 40: There shouldn't be any need for an initial duration, as that can just be calculated using the function as well.

PR #805, opened 2020-04-01
Fix lockdir update, and reporter

Et7f3:
Ok so now lockdir update should be more resilient to package.json non-valid.

the issue before we removed all the lockdir to ensure we have a clean lockdir and we could detect newer dependencies that will break our build. So if a commit without a lockdir is pushed the CI was out. I have opted for a soft `esy install` I will bring the old behaviour back with the new @revery-bot after.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -26,6 +26,7 @@ jobs:
           esy install -P bench
           esy install -P doc
           esy install -P js
+          git diff --exit-code --cached || (git add *lock; git commit -m "Update lockdir, ")
           #  git add *lock
       - name: Print esy cache
         id: print_esy_cache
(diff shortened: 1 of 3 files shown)


Et7f3:
Should be good now. Maybe I can try to erase the `, ` at the end but useful for concatenation of step message. And we need @revery-bot to retrigger the CI because push back with the GITHUB_TOKEN disable CI : so we have to see previous status.

.github/workflows/validate-pr.yml
```diff
@@ -13,15 +13,23 @@ jobs:
...
-          esy install -P bench || true
-          esy install -P doc || true
-          esy install -P js || true
-          git add *lock
+          esy install -P test
+          esy install -P bench
+          esy install -P doc
+          esy install -P js
+          git diff --exit-code --cached || (git add *lock; git commit -m "Update lockdir, ")
+          #  git add *lock
```
glennsl, line 32: There seems to be a lot of unrelated changes here, but I gather this is the relevant part of the change? Do I understand correctly that if any of the `esy install` steps fail, this task should now be a no-op?

Why have the first and last lines here been commented out instead of removed?

Et7f3: The `|| true` were added because if the doc sandbox fail to solve the main sandbox could still be used to use autoformat functionality. No if one esy install fail the whole check fail. I think it was a bad behaviour because we could have a part solved lockdir.

`esy @test install` might need some escape like ' or " depending on the shell with `esy install -P test` no escape are required.

glennsl: Ah, cool, thanks for explaining :)

PR #812, opened 2020-04-09
API: CanvasContext.drawOval and drawCircle

giltho:
Alignes the CanvasContext API with https://github.com/revery-ui/reason-skia/pull/64
Also adds drawOval that was missing from the API

Here's a screenshot of the example app with a circle and an oval successfully drawn
image

I'm not sure I did things as expected for the locks. I just updated them one by one with `esy @sandbox install`

PS: revery is awesome :)


Et7f3:
> I'm not sure I did things as expected for the locks. I just updated them one by one with `esy @sandbox install`

In therory we have GitHub Action that handle that for us so updating sandbox is not mandatory.
Here the task fail with:
> Resource not accessible by integration
I think we use a too old version I will try to update.
Here the issue is formatting. `esy format`. Without that failure the ci should have pushed change.

giltho:
Ah yes, I formatted before writing the example, but forgot to do it after sorry

PR #814, opened 2020-04-14
Dependency: Use latest brisk-reconciler to allow for let-operators

lessp:
In order to be able to use let-operators within Revery the ppx-lib-dependency within brisk-reconciler needed to be updated. This has been done in: https://github.com/briskml/brisk-reconciler/issues/73


glennsl:
Which let-operators is that?

lessp:
> Which let-operators is that?

E.g.
```re
let (let.map) = (promise, fn) => Lwt.map(fn, promise);

let.map result = ...;
```

Et7f3:
Ah it is repared I don't know how this can be broken. @lessp now CI must be green.

PR #819, opened 2020-04-22
Implement #817 - Add onDoubleClick to Clickable

### LINKED ISSUE #817: Add onDoubleClick event to Node
There is an issue in onivim repository ([#1631](https://github.com/onivim/oni2/issues/1631)) that requires to handle TitleBar's double click so we can maximize/minimize window.

I think it would be good to have onDoubleClick event implementation here.

SirFurness:
Fix #817 by adding an `onDoubleClick` callback and `doubleClickDelay` value to `Clickable`.

Hopefully, this is a reasonable way of doing this.

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -12,10 +12,12 @@ open Revery_UI_Primitives;
 module Hooks = Revery_UI_Hooks;
 module Log = (val Log.withNamespace("Revery.Components.Clickable"));
 
-let isMouseCaptured = ref(false);
+module Constants = {
+  let initialMouseDownTimes = (Time.ms(1000), Time.zero);
+  let doubleClickSpeed = Time.ms(500);
+};
 
-let initialMouseDownTimes = (Time.zero, Time.ofFloatSeconds(-1.));
-let doubleClickSpeed: Time.t = Time.ms(500);
+let isMouseCaptured = ref(false);
 
 let%component make =
               (
@@ -37,12 +39,13 @@ let%component make =
               ) => {
   let%hook isMouseCapturedHere = Hooks.ref(false);
 
-  let%hook mouseDownTimes = Hooks.ref(initialMouseDownTimes);
+  let%hook mouseDownTimes = Hooks.ref(Constants.initialMouseDownTimes);
 
   let isDoubleClick = () =>
-    Time.(fst(mouseDownTimes^) - snd(mouseDownTimes^)) <= doubleClickSpeed;
+    Time.(fst(mouseDownTimes^) - snd(mouseDownTimes^))
+    <= Constants.doubleClickSpeed;
   let resetMouseDownTimes = () =>
-    mouseDownTimes := initialMouseDownTimes
+    mouseDownTimes := Constants.initialMouseDownTimes;
 
   let capture = () =>
     if (! isMouseCaptured^) {
@@ -67,14 +70,13 @@ let%component make =
       releaseCapture();
 
       switch (mouseEvt.button) {
-      | MouseButton.BUTTON_LEFT => {
-        if(isDoubleClick()) {
+      | MouseButton.BUTTON_LEFT =>
+        if (isDoubleClick()) {
           resetMouseDownTimes();
-          onDoubleClick()
+          onDoubleClick();
         } else {
-          onClick()
+          onClick();
         }
-      }
       | MouseButton.BUTTON_RIGHT => onRightClick()
       | _ => ()
       };
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -19,6 +19,8 @@ let%component make =
                 ~style=[],
                 ~onClick=() => (),
                 ~onRightClick=() => (),
+                ~onDoubleClick=() => (),
+                ~doubleClickDelay=0.5,
                 ~onAnyClick=_event => (),
                 ~componentRef=?,
                 ~onBlur=?,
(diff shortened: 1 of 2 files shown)


src/UI_Components/Clickable.re
```diff
@@ -19,6 +19,8 @@ let%component make =
                 ~style=[],
                 ~onClick=() => (),
                 ~onRightClick=() => (),
+                ~onDoubleClick=() => (),
+                ~doubleClickDelay=0.5,
```
glennsl, line 23: I don't think there's need to make this configurable. If anything it should use the appropriate delay for the system.

It should also be a `Time.t` value, to make the unit explicit.

src/UI_Components/Clickable.re
```diff
@@ -33,11 +35,32 @@ let%component make =
               ) => {
   let%hook isMouseCapturedHere = Hooks.ref(false);
 
+  let%hook lastCaptureTime = Hooks.ref(None);
+  let%hook currentCaptureTime = Hooks.ref(0.);
```
glennsl, line 39: I don't understand why `currentCaptureTime` and the capturing in general is needed. Can't we just check `(Time.now() - lastClick) <= doubleClickDelay`?

SirFurness: From double-clicking stuff on my computer, it seems that Time.now() should be calculated on mouse down and not on mouse up.

Are you using `lastClick` as just a different name to `lastCaptureTime` or do you mean storing it differently?

src/UI_Components/Clickable.re
SirFurness: The only problem I have is that when I use `Time.t`,  I get this error
```
46 |         let deltaTime: Time.t = currentCaptureTime^ - lastTime;
                                     ^^^^^^^^^^^^^^^^^^^
Error: This expression has type Revery_Core.Time.t
       but an expression was expected of type int
```

Using `-.` also doesn't help because then it wants a float.
How do I do subtraction with `Time.t`?

glennsl: There's arithmetic operators defined in the `Time` module. You can use them by opening the module locally:
```reason
let deltaTime = Time.(currentCaptureTime^ - lastTime);
```

src/UI_Components/Clickable.re
glennsl: Right, "capture" isn't the same as "mouse down" though, so the name is a bit misleading I think. But more importantly the logic is very convoluted and sensitive to order of operations, which make it prone to break by accident when the next person comes in here to change things around..

I would instead suggest using a two-element "stack" to record mouse down times, so you'll always have the last two times available. So instead of this you'd have:

```reason
  let%hook mouseDownTimes = Hooks.ref((Time.zero, Time.zero));
```

then to set it:

```reason
let onMouseDown = _event => {
  capture();
  mouseDownTimes := (Time.now(), fst(mouseDownTimes^));
};
```

and then to use it:

```reason
  let isDoubleClick = () =>
    snd(mouseDownTimes^) -. lastTime <= doubleClickDelay;
```

Does that make sense?

SirFurness:
The `Time.ofFloatSeconds(-1.)` seems a little weird, but I wanted the initial state to return false for `isDoubleClick()`.

Et7f3:
GitHub action had issue I have re-run but it seem to be out of service https://github.com/actions/checkout/issues/23#issuecomment-572688577 from this comment the new version of cache should help.

src/UI_Components/Clickable.re
```diff
@@ -14,11 +14,15 @@ module Log = (val Log.withNamespace("Revery.Components.Clickable"));
 
 let isMouseCaptured = ref(false);
 
+let initialMouseDownTimes = (Time.zero, Time.ofFloatSeconds(-1.));
```
glennsl, line 17: This might break at some point in the future if we decide that negative times are invalid. But this should work too, right?

```suggestion
let initialMouseDownTimes = (Time.ms(1000), Time.zero);
```

src/UI_Components/Clickable.re
```diff
@@ -14,11 +14,15 @@ module Log = (val Log.withNamespace("Revery.Components.Clickable"));
...
+let doubleClickSpeed: Time.t = Time.ms(500);
```
glennsl, line 18: Could you put this in a `Constants` module, before `isMouseCaptured` above?

src/UI_Components/Clickable.re
```diff
@@ -46,14 +57,24 @@ let%component make =
...
   let onMouseLeave = _event => releaseCapture();
   let onMouseUp = (mouseEvt: NodeEvents.mouseButtonEventParams) =>
     if (isMouseCapturedHere^) {
       releaseCapture();
 
       switch (mouseEvt.button) {
-      | MouseButton.BUTTON_LEFT => onClick()
+      | MouseButton.BUTTON_LEFT => {
+        if(isDoubleClick()) {
+          resetMouseDownTimes();
```
glennsl, line 72: Oh, I didn't think of this. That makes the solution much less elegant. Still acceptable, but I wonder if it might be better to model the states explicitly. That would make it easier to extend for triple click as well.

Something like:

```
let clicks : ref([`None | `One(Time.t) | `Two(Time.t) | `Three(Time.t)]) = ref(`None);
```

And for every click you'll need to start a timer to execute the right click and reset the state after the delay.

Just an idea. I don't think it's necessary to do this now.

glennsl:
CI also complains about formatting. You should run `esy format`, commit and push the changes.

src/UI_Components/Clickable.re
SirFurness: So long as Time.now() is not within 1000ms, but that seems pretty unlikely to happen anyway.

src/UI_Components/Clickable.re
SirFurness: Can I just use a normal type instead of the polymorphic variant?

glennsl: You mean an ordinary variant? Sure, it would just be less contained and slightly less convenient as you have to avoid potential conflicts with other types. You shouldn't use `None` for example, as that would shadow the `None` defined by the `option` type, which is used all over the place.

src/UI_Components/Clickable.re
glennsl: That would be the same as with `Time.ofFloatSeconds(-1.)` though. You could of course also increase it to some absurd amount, as long as you don't approach the number of seconds since 1970 ;)

src/UI_Components/Clickable.re
SirFurness: Is there a way to cancel or restart a timeout? I want to restart the timer after accepting another click.

SirFurness: I worked on this for a little, but it's somewhat trickier. If I implement it eventually I think I'll just submit another pull request so that this bit here can work for now.

glennsl: Cool. No problem!

PR #823, opened 2020-04-30
Dependency: reason-skia -> 6b459c7

glennsl:
Pick up clang 10 build fixes


glennsl:
The github actions unfortunately seem completely broken... :(

bryphe:
> The github actions unfortunately seem completely broken... :(

Weird - looks like the Azure CI is OK, so we can override. Thanks for investigating and fixing this!

PR #824, opened 2020-04-30
add getPosition and onMoved to Window

glennsl:


Et7f3:
Old ci was taken (so it fail on re-run). But it was successful. I merge it.

PR #826, opened 2020-05-01
add position to WindowCreateOptions, and Window.isMaximized

glennsl:
This adds `x` and `y` fields to `WindowCreateOption` to be able to specify the initial window position. It also adds a `Window.isMaximized` function.


bryphe:
> Unfortunately it's not possible to actually specify the initial position when creating the SDL window, as that hasn't been exposed, which causes the window to jump initially.

Hmm, can we expose this in our SDL API? It would be nice to have this for window size too, as we have the same issue - it'd be ideal to pass the window size in directly to window creation, too.

> I think ideally we'd model this closer to SDL's approach, with the x and y position given separately as either UNDEFINED, CENTERED or an absolute value, as that would offer more flexibility than we have today. But that would also be a breaking change. 

I don't think we expose position at all today in Window creation, so at least from Revery's perspective, it wouldn't be a breaking change. This seems like the ideal model!

So these set of improvements would be my ideal:
- In SDL2, Pass in window size to window creation
- In SDL2, Expose a variant for window position that can be `Undefined`, `Centered` or `Absolute(int)`, and modify window creation and size setter to handle this
- Update the Revery API to support those

src/Core/WindowCreateOptions.re
```diff
@@ -21,6 +21,12 @@ type t = {
    independent, but as of right now, Transparent only works on macOS.
     */
   titlebarStyle: WindowStyles.titlebar,
+  /**
+    [position] is the initial position of the [Window], either [`Centered] or
+    [`Positioned(x, y)], where [x] and [y] is repsectively the horizontal and
```
bryphe, line 26: Minor type: repsectively -> respectively

bryphe, line 26: I wonder if there are scenarios where the user would want to horizontal center the window, but set an absolute vertical position? IE, specify `xPosition: Centered, yPosition: Positioned(0)`

glennsl: Yeah, that's why and what I suggested in the OP. But it requires changes to reason-sdl2.

glennsl:
SDL2 (not reason-sdl2) has `x` and `y` arguments for `CreateWindow`. But `reason-sdl2` doesn't expose those. Instead it's hard-coded to always center the window. So I think the most straightforward way is to just expose those as optional arguments. But it makes little sense to default to centering if we're going to support SDL2's notion of undefined positioning as well. Therefore it would be a minor breaking change to `reason-sdl2`.

Then when we get to revery we can decide to not offer undefined positioning, default to centering and therefore not make a breaking change to the revery API.

bryphe:
> SDL2 (not reason-sdl2) has x and y arguments for CreateWindow. But reason-sdl2 doesn't expose those. Instead it's hard-coded to always center the window. 

Ah, I see.

> So I think the most straightforward way is to just expose those as optional arguments. 

Makes sense to me.

> Therefore it would be a minor breaking change to reason-sdl2.

I'm fine with breaking changes to `reason-sdl2` - AFAIK Revery is the only consumer.

> Then when we get to revery we can decide to not offer undefined positioning, default to centering and therefore not make a breaking change to the revery API.

Sounds like a great plan!

glennsl:
> I'm fine with breaking changes to reason-sdl2 - AFAIK Revery is the only consumer.

Cool, in that case we might as well make them mandatory to align it even closer with the SDL2 API.

Now to figure out how to deal with polymorphic variants in the FFI...

bryphe:
> Now to figure out how to deal with polymorphic variants in the FFI...

Ooh... I haven't encountered that before... A couple options:
- It might be worth coercing them to a more FFI-friendly value (ie, an `int` or `byte`)
- It does seem like we can access the underlying value: https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#ss:c-polyvar via the `hash_variant` method, so it might not be too bad. Seems similar to working with normal variants, expect we need to get the ID via the `hash_variant` method, and there is some difference in how variants taking multiple arguments are represented.

glennsl:
Yeah, I think the second option would also yield much more readable stub code. Although also somewhat more prone to error by typo. And it's of course less performant, but that isn't really significant here.

package.json
```diff
@@ -53,6 +53,9 @@
     "revery-text-wrap": "revery-ui/revery-text-wrap#966383e",
     "@glennsl/timber": "1.0.0"
   },
+  "resolutions": {
+    "reason-sdl2": "revery-ui/reason-sdl2#d7da293"
+  },
```
glennsl, line 58: This needs a release of `reason-sdl2` to remove this, then I think it's good to go!

bryphe: Cool! Just published `reason-sdl@2.10.3026`

glennsl:
GHAs are failing on `pkg-config`:

```
  build log:
    # esy-build-package: building: @opam/conf-pkg-config@opam:1.2
    # esy-build-package: pwd: /Users/runner/.esy/3/b/opam__s__conf_pkg_config-opam__c__1.2-29e943ab
    # esy-build-package: running: 'pkg-config' '--help'
    esy-build-package: unable to resolve command: pkg-config
    
  building @opam/conf-pkg-config@opam:1.2
```

Don't know what that's about.

glennsl:
Is the problem just that we don't have an override like [this](https://github.com/esy-ocaml/esy-opam-override/tree/6/packages/conf-pkg-config.1.1) for pkg-config 1.2?

package.json
glennsl: Hmm, because this version of `reason-sdl2` had breaking changes and you just did a path bump, oni2 master is now broken. Will be fixed as soon as https://github.com/onivim/oni2/pull/1703 is merged of course, but still unfortunate. Not sure how this could have been avoided with the versioning scheme used here, but definitely something to keep in mind.

PR #827, opened 2020-05-02
add password type inputs

despairblue:
This adds a boolean option to the input component to obscure the entered value on render:

Screenshot from 2020-05-02 17-13-42

The name `typePassword` is based on the [input type of html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password). I can change it if something like `obscureValue` makes more sense.

### DIFF
--- a/src/UI_Components/Input.re
+++ b/src/UI_Components/Input.re
@@ -106,10 +106,6 @@ let addCharacter = (word, char, index) => {
   (startStr ++ char ++ endStr, String.length(startStr) + 1);
 };
 
-let renderAsPassword = value => {
-  StringLabels.map(value, ~f=_ => '*');
-};
-
 let reducer = (action, state) =>
   switch (action) {
   | Focus => {...state, isFocused: true}
@@ -210,7 +206,7 @@ let%component make =
                 ~onChange=(_, _) => (),
                 ~value=?,
                 ~cursorPosition=?,
-                ~typePassword=false,
+                ~isPassword=false,
                 (),
               ) => {
   let%hook (state, dispatch) =
@@ -376,7 +372,7 @@ let%component make =
       ref={node => textRef := Some(node)}
       text={
         showPlaceholder
-          ? placeholder : typePassword ? renderAsPassword(value) : value
+          ? placeholder : isPassword ? String.map(_ => '*', value) : value
       }
       smoothing
       style={Styles.text(
--- a/src/UI_Components/Input.re
+++ b/src/UI_Components/Input.re
@@ -367,13 +367,19 @@ let%component make =
     </View>;
   };
 
+  let text =
+    if (showPlaceholder) {
+      placeholder;
+    } else if (isPassword) {
+      String.map(_ => '*', value);
+    } else {
+      value;
+    };
+
   let text = () =>
     <Text
       ref={node => textRef := Some(node)}
-      text={
-        showPlaceholder
-          ? placeholder : isPassword ? String.map(_ => '*', value) : value
-      }
+      text
       smoothing
       style={Styles.text(
         ~showPlaceholder,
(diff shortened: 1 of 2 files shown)


Et7f3:
For security reason I think you shouldn't display * because they give indication on how many char you have in your password. Git in terminal doesn't display anything. I think a better approch is only highlight when the password field is selected (so we inform user we are ready to enter password but don't display it.)

lessp:
I would saying mirroring how the web, most OS's and apps handle password-input is the expected behaviour for most users. I guess if one would really like to hide the password you could simply use a text-field with a text-color matching the background.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password

src/UI_Components/Input.re
```diff
@@ -106,6 +106,10 @@ let addCharacter = (word, char, index) => {
   (startStr ++ char ++ endStr, String.length(startStr) + 1);
 };
 
+let renderAsPassword = value => {
+  StringLabels.map(value, ~f=_ => '*');
+};
+
```
glennsl, line 112: There's doesn't seem to be any benefit to using `StringLabels` here. And as it isn't used anywhere else in Revery either, as far as I'm aware, it would be nice to keep being consistent about it.

```suggestion
let renderAsPassword = String.map(_ => '*');
```

Although when a function is so small that its name exceeds its definition, it might be better to just inline it.

src/UI_Components/Input.re
```diff
@@ -206,6 +210,7 @@ let%component make =
                 ~onChange=(_, _) => (),
                 ~value=?,
                 ~cursorPosition=?,
+                ~typePassword=false,
```
glennsl, line 213: I'd say just `~isPassword` is fine. Alternatively we could have a `~type_` property taking a polymorphic variant is we want to extend this with other input types. I think having separate components for those would be fine though.

src/UI_Components/Input.re
```diff
@@ -369,7 +374,10 @@ let%component make =
   let text = () =>
     <Text
       ref={node => textRef := Some(node)}
-      text={showPlaceholder ? placeholder : value}
+      text={
+        showPlaceholder
+          ? placeholder : typePassword ? renderAsPassword(value) : value
+      }
```
glennsl, line 380: With `renderAsPassword` inlined:

```suggestion
      text={
        showPlaceholder
          ? placeholder : typePassword ? String.map(_ => "*", value) : value
      }
```

src/UI_Components/Input.re
despairblue: Yeah, I agree that having different components for date inputs etc makes more sense than blindly copying the HTML input element.

I'll rename it to `isPassword`

despairblue:
@glennsl I addressed your comments. Thanks for the review :clap:

src/UI_Components/Input.re
```diff
@@ -369,7 +370,10 @@ let%component make =
   let text = () =>
     <Text
       ref={node => textRef := Some(node)}
-      text={showPlaceholder ? placeholder : value}
+      text={
+        showPlaceholder
+          ? placeholder : isPassword ? String.map(_ => '*', value) : value
+      }
```
glennsl, line 376:
```suggestion
      text={
        if (showPlaceholder) {
          placeholder;
        } else if (isPassword) {
          String.map(_ => '*', value);
        } else {
          value
        }
      }
```

despairblue:
> Thanks! One more minor issue is that nested ternaries tend not to be very readable, so I suggest using an `if` expression instead, despite all the annoying curlies. Might be an idea to pull it out of the JSX though.

Yeah, thought the same, but wasn't sure if I should pull it out. I did now :)

glennsl:
Awesome, thanks! I think Github is just confused about the CI state here as it looks good if I click through, so I'll merge.

PR #828, opened 2020-05-03
allow pasting into inputs

despairblue:
This allows pasting into inputs from the clip board into the current position.

I'd also like to ask for some general feedback. I'd like to add things like 

* <KBD>CTRL-Z</KBD> support
	* How much history would we like to keep around?
	* Do we want to group inputs similar to what content editables in browser do? E.g. typing one word really fast and hitting the undo shortcut will delete the whole word instead of just the last character. 
* <KBD>CTRL-A</KBD> and selection support
	* We probably need to decide on colors for that? Probably just inverting foreground and background color? But how do I get the background color of my parent?

Just some thoughts. I can open individual issues for those.

### DIFF
--- a/src/UI_Components/Input.re
+++ b/src/UI_Components/Input.re
@@ -104,9 +104,12 @@ let removeCharacterAfter = (word, cursorPosition) => {
   (newString, cursorPosition);
 };
 
-let addCharacter = (word, char, index) => {
-  let (startStr, endStr) = getStringParts(index, word);
-  (startStr ++ char ++ endStr, String.length(startStr) + 1);
+let insertString = (currentValue, insertion, index) => {
+  let (startStr, endStr) = getStringParts(index, currentValue);
+  (
+    startStr ++ insertion ++ endStr,
+    String.length(startStr) + String.length(insertion),
+  );
 };
 
 module Constants = {
@@ -288,10 +291,18 @@ let%component make =
     dispatch(TextInput(value, cursorPosition));
   };
 
+  let paste = (currentValue, currentCursorPosition) => {
+    let (newValue, newCursorPosition) =
+      Sdl2.Clipboard.getText()
+      |> Option.value(~default="")
+      |> insertString(currentValue, _, currentCursorPosition);
+    update(newValue, newCursorPosition);
+  };
+
   let handleTextInput = (event: NodeEvents.textInputEventParams) => {
     resetCursor();
     let (value, cursorPosition) =
-      addCharacter(value, event.text, cursorPosition);
+      insertString(value, event.text, cursorPosition);
     update(value, cursorPosition);
   };
 
@@ -319,6 +330,19 @@ let%component make =
       update(value, cursorPosition);
 
     | v when Key.Keycode.escape == v => Focus.loseFocus()
+
+    | v
+        when
+          118 == v
+          && Environment.os === Mac
+          && Sdl2.Keymod.isGuiDown(event.keymod) =>
+      paste(value, cursorPosition)
+    | v
+        when
+          118 == v
+          && Environment.os !== Mac
+          && Sdl2.Keymod.isControlDown(event.keymod) =>
+      paste(value, cursorPosition)
 
     | _ => ()
     };
--- a/src/UI_Components/Input.re
+++ b/src/UI_Components/Input.re
@@ -307,44 +307,35 @@ let%component make =
   };
 
   let handleKeyDown = (event: NodeEvents.keyEventParams) => {
+    open Key;
+
     resetCursor();
     onKeyDown(event);
 
-    switch (event.keycode) {
-    | v when Key.Keycode.left == v =>
+    let code = event.keycode;
+    if (code == Keycode.left) {
       let cursorPosition = getSafeStringBounds(value, cursorPosition, -1);
       update(value, cursorPosition);
-
-    | v when Key.Keycode.right == v =>
+    } else if (code == Keycode.right) {
       let cursorPosition = getSafeStringBounds(value, cursorPosition, 1);
       update(value, cursorPosition);
-
-    | v when Key.Keycode.delete == v =>
+    } else if (code == Keycode.delete) {
       let (value, cursorPosition) =
         removeCharacterAfter(value, cursorPosition);
       update(value, cursorPosition);
-
-    | v when Key.Keycode.backspace == v =>
+    } else if (code == Keycode.backspace) {
       let (value, cursorPosition) =
         removeCharacterBefore(value, cursorPosition);
       update(value, cursorPosition);
-
-    | v when Key.Keycode.escape == v => Focus.loseFocus()
-
-    | v
-        when
-          118 == v
-          && Environment.os === Mac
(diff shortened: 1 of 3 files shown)


Et7f3:
For limit history I think a attribute: `~limit=Some(50)` as default is a good standard. A user can specify `None` for no limit.

src/UI_Components/Input.re
```diff
@@ -326,6 +337,19 @@ let%component make =
...
+          118 == v
+          && Environment.os === Mac
+          && Sdl2.Keymod.isGuiDown(event.keymod) =>
+      paste(value, cursorPosition)
+    | v
+        when
+          118 == v
+          && Environment.os !== Mac
+          && Sdl2.Keymod.isControlDown(event.keymod) =>
+      paste(value, cursorPosition)
```
glennsl, line 351: These magic number should be added to `Key.Keycode`, as otherwise I have no clue what they're referring to. I'd also suggest matching on `Environment.os`, although this whole `switch` is kind of useless as it doesn't currently pattern match on anything. It's just a sequence of if conditionals, which tend to be better written as... a sequence of if conditionals.

despairblue: I made a PR to [reason-sdl2](https://github.com/revery-ui/reason-sdl2/pull/71). I also added the definition to `Key.rei`. Now I have the problem that `Key.Keycode` also exports a type named `t`, but it will also export a magic number called `t` for the letter `t`. 

I could prefix all letters with `lettera` similar to `digit0`. Problem there is:

1. It's a breaking change for everybody using `Key.Keycode.c`
2. There should be a delimiter between the word `letter` and the actual letter: `letter_a` but that makes it inconsistent with `digit0`.

I don't really mind which way to go. Let me know what you think @glennsl

glennsl: >  Now I have the problem that `Key.Keycode` also exports a type named `t`, but it will also export a magic number called `t` for the letter `t`.
Not sure I see the problem with that. Types and variables have separate namespaces, so there should be no conflict unless there's code generation associated with the type.

despairblue: I added the keycodes to the keycode modules' types in Key.rei:

```re

let t: int;


let a: int;
// ...
let t: int;
// ...
```

With that I got a problem in `examples/Calculator.re`


I'm currently not at my computer. I'll follow up with the exact error.

glennsl: Right, yeah I see several collisions there because of poor variable naming combined with the module being opened.

Can be fixed either by renaming `e` and `v` to something sensible like `event` and `key`, or by removing this:

https://github.com/revery-ui/revery/blob/43a29184211ea259ddaa2e18d3a44535e59a51a2/examples/Calculator.re#L253

and fully qualifying the keycodes.

Et7f3: I think we can do the other way: `NodeEvents.e`

despairblue: I renamed `e` to `event`:

```re
    let respondToKeys = (event: NodeEvents.keyEventParams) => {
```

despairblue: I rewrote this part @glennsl. Not really happy with the condition with the `Environment.os` but rewriting it with a switch inside wouldn't make it more readable, it would just add more noise in my opinion, but we would get exhaustiveness checks. Your call.

 I'd like to have it like this:

```re
      if (
        (Environment.os === Mac && Sdl2.Keymod.isGuiDown(event.keymod)) ||
        (Environment.os !== Mac && Sdl2.Keymod.isControlDown(event.keymod)
      ) {
        paste(value, cursorPosition);
      };
```

But refmt thinks otherwise :)

despairblue:
I'll rebase.

src/UI_Components/Input.re
```diff
@@ -294,39 +297,51 @@ let%component make =
...
+    } else if (code == Keycode.escape) {
+      Focus.loseFocus();
+    } else if (code == Keycode.v) {
+      if (Environment.os === Mac
+          && Sdl2.Keymod.isGuiDown(event.keymod)
+          || Environment.os !== Mac
+          && Sdl2.Keymod.isControlDown(event.keymod)) {
+        paste(value, cursorPosition);
+      };
     };
```
glennsl, line 345: This conditional is about confusing. What if we pull some of this verbosity out into a couple of variables:

```reason
let mac = Environment.os == Mac;
let super = Sdl2.Keymod.isGuiDown(event.keymod);
let ctrl = Sdl2.Keymod.isControlDown(event.keymod);
```
then we can just do

```suggestion
    } else if (mac && super && code == Keycode.v) {
        paste(value, cursorPosition);
    } else if (!mac && ctrl && code == Keycode.v) {
        paste(value, cursorPosition);
    };
```

glennsl:
Also, you need to run `./update-lockfiles.sh` and commit all the updated lockfiles after modifying dependencies.

despairblue:
@glennsl done and done :100:

zbaylin:
@despairblue you have some merge conflicts again. If you fix those I can merge this.

despairblue:
@zbaylin Did so :)

despairblue:
@glennsl @zbaylin Should I rebase this again?

zbaylin:
😳 Sorry @despairblue! I got distracted with other things, I guess. My bad! Yeah if you could rebase I'll make sure nothing major has changed architecturally and then merge. Thanks again!

Et7f3:
Sorry for the double force-push (it was to rebase and In first rebase one commit lost his author name...)

Et7f3:
I think it is a issue with my computer only: it doesn't work. I build with the old HEAD (maybe we introduced bug in repo merge).

src/UI_Components/Input.re
```diff
@@ -288,39 +291,52 @@ let%component make =
     dispatch(TextInput(value, cursorPosition));
   };
 
+  let paste = (currentValue, currentCursorPosition) => {
+    let (newValue, newCursorPosition) =
+      Sdl2.Clipboard.getText()
+      |> Option.value(~default="")
+      |> insertString(currentValue, _, currentCursorPosition);
+    update(newValue, newCursorPosition);
```
Et7f3, line 299:
```suggestion
    switch (Sdl2.Clipboard.getText()) {
    | None => ()
    | Some(data) =>
      let (newValue, newCursorPosition) =
        insertString(currentValue, data, currentCursorPosition);
      update(newValue, newCursorPosition);
    };
```
It help to skip update if no data.

Et7f3:
This commit work: 630b01aaf358afafc84d4faf0b8104079849c53f.
I will try to cherry-pick one by one

despairblue:
> I think it is a issue with my computer only: it doesn't work. I build with the old HEAD (maybe we introduced bug in repo merge).

Let me try to look at the old version and see what may have broken it.

Et7f3:
e5f363a1 seem guilty

src/UI_Components/Input.re
despairblue: Good idea!

Et7f3:
Thanks for the feature and for finding this typo :)

PR #829, opened 2020-05-04
Add minimize function to Window

zbaylin:
Binds to reason-sdl2's `Sdl2.Window.minimize`


zbaylin:
Woah -- I didnt expect all those commits to get there. The last one is the only one that matters. Sorry about that!

zbaylin:
This won't pass the CI because it's not used within Revery, but is needed for completing onivim/oni2#1717

Et7f3:
If you have all those commit you can create a new local branch cherry pick the three last commit `git cherry-pick sha` then force push to this branch and history will be clean. You can even merge your formatting commit with your code.
src/Core/Windows.re have a interface file. Since OCaml know the shape of the module it can know such value are unused (because not exported nor used in files). I have exported it so it won't complain anymore.

PR #831, opened 2020-05-06
Events: Implement FileDrop event

zbaylin:
This is modeled after the Mouse events, which I think are most similar
to what we're trying to accomplish here.


package.json
```diff
@@ -54,7 +54,8 @@
     "@glennsl/timber": "1.0.0"
   },
   "resolutions": {
-    "@opam/conf-pkg-config": "1.1"
+    "@opam/conf-pkg-config": "1.1",
+    "reason-sdl2": "revery-ui/reason-sdl2#737721d"
```
glennsl, line 58: Just a note that this ought to be removed before merging, in favor of doing a proper release of reason-sdl2.

src/Core/Window.re
```diff
@@ -372,6 +374,15 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
 
   | Sdl2.Event.WindowFocusGained(_) => Event.dispatch(v.onFocusGained, ())
   | Sdl2.Event.WindowFocusLost(_) => Event.dispatch(v.onFocusLost, ())
+  | Sdl2.Event.DropFile({x, y, file, _}) =>
+    Event.dispatch(
+      v.onFileDropped,
+      {
+        mouseX: float_of_int(x),
+        mouseY: float_of_int(y),
```
glennsl, line 382: Handy short-hand:

```suggestion
        mouseX: float(x),
        mouseY: float(y),
```

src/UI/FileDrop.re
```diff
@@ -0,0 +1,44 @@
...
+
+let getPositionFromFileDropEvent = (evt: Events.internalFileDropEvents) =>
+  switch (evt) {
+  | InternalFileDropped({mouseX, mouseY, _}) => (mouseX, mouseY)
+  };
+
+let isFileDroppedEv =
+  fun
+  | FileDropped(_) => true
+  | _ => false;
```
glennsl, line 22: This doesn't seem to be used anywhere. Should be removed if not expected to be needed later.

src/UI/FileDrop.re
```diff
@@ -0,0 +1,44 @@
...
+
+module Log = (val Log.withNamespace("Revery.UI.FileDropped"));
+
+let internalToExternalEvent = (evt: Events.internalFileDropEvents) =>
+  switch (evt) {
+  | InternalFileDropped(evt) =>
+    FileDropped({mouseX: evt.mouseX, mouseY: evt.mouseY, path: evt.path})
+  };
+
+let getPositionFromFileDropEvent = (evt: Events.internalFileDropEvents) =>
```
glennsl, line 14: Since this is in a `FileDrop` module, I think that context makes repeating it in the function names redundant. But what _is_ missing is that it operates on the _internal_ event.

```suggestion
let getPositionFromInternalEvent = (evt: Events.internalFileDropEvents) =>
```

src/UI/Node.re
```diff
@@ -368,6 +368,8 @@ class node (()) = {
     | (TextEdit(_), _)
     | (KeyDown(_), _)
     | (KeyUp(_), _) => ()
+    | (FileDropped(e), {onFileDropped: Some(cb)}) => cb(e)
```
glennsl, line 371:
```suggestion
    | (FileDropped(e), {onFileDropped: Some(cb), _}) => cb(e)
```

You'll get a warning for this if you run `dune build @check`

src/UI/Ui.re
```diff
@@ -38,7 +36,7 @@ let start = (window: Window.t, element: React.element(React.reveryNode)) => {
   let container = Container.create(rootNode);
   let ui = RenderContainer.create(window, rootNode, container, mouseCursor);
 
-  let _ignore = Window.onExposed(window, () => {uiDirty := true});
+  let _ignore = Window.onExposed(window, () => uiDirty := true);
```
glennsl, line 39: Out of scope for this PR, but just as an opportunity to learn: using `_ignore` is setting yourself up for accidental partial application. If an argument is added `Window.onExposed` this will still continue to compile but it will not be called, because it will instead return a partially applied function that's waiting for the last argument to be given to it. Normally this would be caught because stand-alone "statements" are required to evaluate to `unit`. But since this explicitly ignores and discards the result, it's telling the compiler "trust me, I know what I'm doing". Which unfortunately isn't the case here. And in fact if you just use `_` the compiler will still emit a warning. Using `_ignore` instead is even more explicit, demonstrating that the warning has been noted but is still being actively ignored. Which is terrible practice in general.

What you should do instead in cases like this is either:

```suggestion
  let _: uint => unit = Window.onExposed(window, () => uiDirty := true);
```
or

```suggestion
  ignore(Window.onExposed(window, () => uiDirty := true): unit => unit);
```

(Or better yet, if possible, change the `Window` API to better account for this use case)

Assuming the return type is `unit => unit`. I haven't bothered to check. But in either case since this is explicit about the type returned, if that type changes, because an argument is added or because it's changed in other ways, this will fail to compile and force you to fix and check that it is still correct.

src/UI/Ui.re
```diff
@@ -120,9 +118,24 @@ let start = (window: Window.t, element: React.element(React.reveryNode)) => {
...
+          Revery_Core.Events.InternalFileDropped({
+            mouseX: f.mouseX,
+            mouseY: f.mouseY,
+            path: f.path,
+          });
+        FileDrop.dispatch(evt, rootNode);
+      },
+    );
+
+  let _ignore =
```
glennsl, line 135: This one is in scope though. But it also demonstrates how having this kind of code around makes it seem "normal" and causes it to get abused by contributors who don't know better and just copy the conventions that seem to be used. Which, to be clear, is good. Existing conventions should be copied. The problem is solely that some existing conventions are bad. Very bad.

src/Core/Events.re
```diff
@@ -38,3 +38,12 @@ type internalMouseEvents =
...
   | InternalMouseOut(mouseMoveEvent);
+
+type fileDropEvent = {
+  mouseX: float,
+  mouseY: float,
+  path: string,
+};
+
+type internalFileDropEvents =
+  | InternalFileDropped(fileDropEvent);
```
glennsl, line 49: Why is this a variant if there's just a single case? Will there be more?

package.json
zbaylin: Yep! Totally agree -- I just wanted to get started on this before a new version of reason-sdl2 was pushed.

src/Core/Window.re
zbaylin: Interesting -- is this new? I've always used `float_of_int`. TIL. Thanks!

src/UI/FileDrop.re
zbaylin: It can be removed for now. It's a relic of an implementation I changed before committing.

src/UI/FileDrop.re
zbaylin: That's fine -- I thought it was a bit redundant too. I was taking a lot of the naming schemes from `Mouse.re`, i.e.
https://github.com/revery-ui/revery/blob/cd5911c823b4f98a0d46e27ac8bb44f09dc1f8d9/src/UI/Mouse.re#L269

Up to you if you think it's worth keeping for consistency, or changing for clarity.

src/UI/Node.re
zbaylin: 😮 not sure how I missed that! Thanks!

src/UI/Ui.re
zbaylin: Thanks @glennsl! I've definitely been burned by partial application bugs before. I will open a separate PR for annotating the types of all these variables.

src/UI/Ui.re
zbaylin: Sounds good -- I will annotate this to make sure it's correct

src/Core/Events.re
zbaylin: The plan is to eventually include an active drag event. Unfortunately SDL2 doesn't provide this in their SDL_Event abstractions, so we (or I) will have to make a custom SDL2 event that is triggered by platform specific events. When that is merged into reason-sdl2, that variant will provide a clear/easy step to adding the "a file is being dragged" event. I can get rid of it for now, but that was the rationale.

src/UI/Node.re
zbaylin: I fixed this in the latest push! Thanks!

src/UI/FileDrop.re
zbaylin: I decided that I think the naming is worth changing. I did so in one of the latest commits. Thanks!

src/UI/FileDrop.re
zbaylin: Removed in the latest commit!

src/Core/Window.re
zbaylin: Changed in the latest commit! Thanks!

glennsl: I don't think it is. At least it doesn't have a `since` annotation in the docs, implying it's very old. I think most people expect it to be named `float_of_int` though, as that's the general convention. But the short-hand really is very convenient!

src/Core/Events.re
glennsl: Cool. Sounds like a good plan!

src/UI/FileDrop.re
glennsl: Good call!

examples/FileDragAndDrop.re
```diff
@@ -0,0 +1,44 @@
...
+      fontSize(14.),
+    ];
+};
+
+let%component dnd = () => {
+  let%hook (text, setText) = Hooks.state("Nothing dropped yet!");
+
+  <View style=Styles.outer>
+    <Text style=Styles.text text />
+    <View style=Styles.droppable onFileDropped={e => setText(_ => e.path)}>
```
bryphe, line 38: Thanks for adding the example, Zach!

bryphe:
Oh no, lockfiles got out-of-date!

FYI - the quickest way to update them (that I know) is:
- `git merge master`
- `git checkout master *esy.lock`
-  `./update-lockfiles.sh`

This merges master, reverts local lockfile changes, and updates them with esy install.

zbaylin:
✅ CI has passed, looks like we can merge this in now!

PR #832, opened 2020-05-06
chore: remove pkg-config workaround

glennsl:
Root cause fixed in https://github.com/esy-ocaml/esy-opam-override/pull/62


glennsl:
Bump. Need an approval for this. Docs action failed, but then it always does it seems.

PR #833, opened 2020-05-07
FileDrop: collect multiple files dropped at once

zbaylin:
Changes the file drop event to store a list of strings/paths.
The event handler "collects" all the files dropped between BEGIN and
COMPLETE and sends the populated list out

Example:

screencast 2020-05-07 12-52-58

### DIFF
--- a/examples/FileDragAndDrop.re
+++ b/examples/FileDragAndDrop.re
@@ -30,12 +30,15 @@ module Styles = {
     ];
 };
 
+let filesToText = files =>
+  List.map(file => <Text text=file style=Styles.text />, files);
+
 let%component dnd = () => {
-  let%hook (text, setText) = Hooks.state("Nothing dropped yet!");
+  let%hook (files: list(string), setFiles) = Hooks.state([]);
 
   <View style=Styles.outer>
-    <Text style=Styles.text text />
-    <View style=Styles.droppable onFileDropped={e => setText(_ => e.path)}>
+    {React.listToElement(filesToText(files))}
+    <View style=Styles.droppable onFileDropped={e => setFiles(_ => e.paths)}>
       <Text style=Styles.text text="Drop here!" />
     </View>
   </View>;
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -378,7 +378,7 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
   | Sdl2.Event.DropBegin(_) => v.dropState = Some([])
   | Sdl2.Event.DropFile({file, _}) =>
     switch (v.dropState) {
-    | Some(list) => v.dropState = Some([Option.get(file), ...list])
+    | Some(list) => v.dropState = Some([file, ...list])
     | None =>
       Log.warn("Received drop file event without preceding drop begin")
     }
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -63,7 +63,7 @@ type t = {
   mutable requestedHeight: option(int),
   // True if composition (IME) is active
   mutable isComposingText: bool,
-  mutable dropState: dropState,
+  mutable dropState,
   titlebarStyle: WindowStyles.titlebar,
   onBeforeRender: Event.t(unit),
   onAfterRender: Event.t(unit),
@@ -384,19 +384,22 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
   | Sdl2.Event.DropBegin(_) => v.dropState = Dropping([])
   | Sdl2.Event.DropFile({file, _}) =>
     switch (v.dropState) {
-      | Dropping(list) => v.dropState = Dropping([Option.get(file), ...list])
-      | Idle => Log.warn("Received drop file event without preceding drop begin")
+    | Dropping(list) => v.dropState = Dropping([Option.get(file), ...list])
+    | Idle =>
+      Log.warn("Received drop file event without preceding drop begin")
     }
   | Sdl2.Event.DropComplete({x, y, _}) =>
     switch (v.dropState) {
-    | Idle | Dropping([]) => Log.warn("Received drop complete event without preceding drop events");
+    | Idle
+    | Dropping([]) =>
+      Log.warn("Received drop complete event without preceding drop events")
     | Dropping(list) =>
       v.dropState = Idle;
       Event.dispatch(
         v.onFileDropped,
-        {mouseX: float(x), mouseY: float(y), paths: List.rev(list)}
-      )
-    };
+        {mouseX: float(x), mouseY: float(y), paths: List.rev(list)},
+      );
+    }
   | Sdl2.Event.Quit => ()
   | _ => ()
   };
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -43,10 +43,6 @@ module WindowMetrics = {
   };
 };
 
-type dropState =
-  | Idle
-  | Dropping(list(string));
-
 type t = {
   mutable backgroundColor: Color.t,
   sdlWindow: Sdl2.Window.t,
@@ -63,7 +59,7 @@ type t = {
   mutable requestedHeight: option(int),
   // True if composition (IME) is active
   mutable isComposingText: bool,
-  mutable dropState,
+  mutable dropState: option(list(string)),
   titlebarStyle: WindowStyles.titlebar,
   onBeforeRender: Event.t(unit),
   onAfterRender: Event.t(unit),
@@ -379,20 +375,20 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
 
   | Sdl2.Event.WindowFocusGained(_) => Event.dispatch(v.onFocusGained, ())
   | Sdl2.Event.WindowFocusLost(_) => Event.dispatch(v.onFocusLost, ())
-  | Sdl2.Event.DropBegin(_) => v.dropState = Dropping([])
+  | Sdl2.Event.DropBegin(_) => v.dropState = Some([])
   | Sdl2.Event.DropFile({file, _}) =>
     switch (v.dropState) {
-    | Dropping(list) => v.dropState = Dropping([Option.get(file), ...list])
-    | Idle =>
+    | Some(list) => v.dropState = Some([Option.get(file), ...list])
+    | None =>
       Log.warn("Received drop file event without preceding drop begin")
     }
   | Sdl2.Event.DropComplete({x, y, _}) =>
     switch (v.dropState) {
-    | Idle
-    | Dropping([]) =>
+    | None
+    | Some([]) =>
       Log.warn("Received drop complete event without preceding drop events")
-    | Dropping(list) =>
-      v.dropState = Idle;
+    | Some(list) =>
+      v.dropState = None;
       Event.dispatch(
         v.onFileDropped,
         {mouseX: float(x), mouseY: float(y), paths: List.rev(list)},
@@ -526,7 +522,7 @@ let create = (name: string, options: WindowCreateOptions.t) => {
     requestedHeight: None,
 
     isComposingText: false,
-    dropState: Idle,
+    dropState: None,
 
     forceScaleFactor: options.forceScaleFactor,
 
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -132,8 +132,6 @@ let onSizeChanged = w => Event.subscribe(w.onSizeChanged);
 let onMoved = w => Event.subscribe(w.onMoved);
 let onFileDropped = w => Event.subscribe(w.onFileDropped);
 
-let dropList: ref(list(string)) = ref([]);
-
 let getUniqueId = (w: t) => w.uniqueId;
 
 let isDirty = (w: t) =>
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -385,7 +385,7 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
       Log.warn("Received drop file event without preceding drop begin");
     }
   | Sdl2.Event.DropComplete({x, y, _}) =>
-    if (v.isDropping && List.length(dropList^) > 0) {
+    if (v.isDropping && dropList^ == []) {
       v.isDropping = false;
       Event.dispatch(
         v.onFileDropped,
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -59,6 +59,7 @@ type t = {
   mutable requestedHeight: option(int),
   // True if composition (IME) is active
   mutable isComposingText: bool,
+  mutable isDropping: bool,
   titlebarStyle: WindowStyles.titlebar,
   onBeforeRender: Event.t(unit),
   onAfterRender: Event.t(unit),
@@ -126,6 +127,8 @@ let onTextInputCommit = w => Event.subscribe(w.onTextInputCommit);
 let onSizeChanged = w => Event.subscribe(w.onSizeChanged);
 let onMoved = w => Event.subscribe(w.onMoved);
 let onFileDropped = w => Event.subscribe(w.onFileDropped);
+
+let dropList : ref(list(string)) = ref([]);
 
 let getUniqueId = (w: t) => w.uniqueId;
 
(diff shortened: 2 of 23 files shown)


src/Core/Window.re
```diff
@@ -374,15 +377,32 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
-    )
+  | Sdl2.Event.DropBegin(_) => v.isDropping = true
+  | Sdl2.Event.DropFile({file, _}) =>
+    if (v.isDropping) {
+      dropList := List.append(dropList^, [Option.value(file, ~default="")]);
+    } else {
+      Log.error("Received drop file event without preceding drop begin");
+    }
+  | Sdl2.Event.DropComplete({x, y, _}) =>
+    if (v.isDropping && List.length(dropList^) > 0) {
```
glennsl, line 388: You should never use `List.length` to check for emptiness as it's very inefficient compared to structural comparison against an empty list. The reason is that, unlike an array, a list does not have a `length` field. It has to actually traverse the entire list to get its length. Every time. It's also unnecessarily verbose and indirect when you have structural comparison, as we do in functional-land.

```suggestion
    if (v.isDropping && dropList^ != []) {
```

On a more out-of-scope note, the `v ` variable here apparently refers to a window, but I had to scroll around way too much to figure that out. We ought to chuck that into the bucket of stuff that ought to be cleaned up around here as well.

src/Core/Window.re
```diff
@@ -374,15 +377,32 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
-      {
-        mouseX: float_of_int(x),
-        mouseY: float_of_int(y),
-        path: Option.value(file, ~default=""),
-      },
-    )
+  | Sdl2.Event.DropBegin(_) => v.isDropping = true
+  | Sdl2.Event.DropFile({file, _}) =>
+    if (v.isDropping) {
+      dropList := List.append(dropList^, [Option.value(file, ~default="")]);
```
glennsl, line 383: Why are we adding empty strings to the list if there's no `file`? Is that useful for anything? In which scenarios can `file` not be present?

Also, you should generally avoid using `List.append`, `@` or appending to the end of a list in any other way as it's very inefficient. Each time you append, the entire list will have to be traversed and every node reallocated. Instead it's almost always better to prepend (`[item, ...list]`) and then reverse the list when you're done, in this case on `DropComplete`.

src/Core/Window.re
```diff
@@ -374,15 +377,32 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
+    } else {
+      Log.error("Received drop file event without preceding drop begin");
```
glennsl, line 385: I think this is more of a warning than an error. It's suspicious, but might not actually cause a problem. See the [Oni2 convention for log levels](https://onivim.github.io/docs/for-developers/contributing#log-levels)

src/Core/Window.re
```diff
@@ -374,15 +377,32 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
+  | Sdl2.Event.DropComplete({x, y, _}) =>
+    if (v.isDropping && List.length(dropList^) > 0) {
+      v.isDropping = false;
+      Event.dispatch(
+        v.onFileDropped,
+        {mouseX: float(x), mouseY: float(y), paths: dropList^},
+      );
+      dropList := [];
+    } else {
+      Log.error("Received drop complete event without preceding drop events");
```
glennsl, line 396: Also here.

src/Core/Window.re
```diff
@@ -127,6 +128,8 @@ let onSizeChanged = w => Event.subscribe(w.onSizeChanged);
 let onMoved = w => Event.subscribe(w.onMoved);
 let onFileDropped = w => Event.subscribe(w.onFileDropped);
 
+let dropList: ref(list(string)) = ref([]);
+
```
glennsl, line 132: Not a big fan of global mutable variables, and would like to contain this as much as possible. Restrict who can mutate it, and in which way it can be mutated.

A few ideas:
- At the very least, I would think it could be moved into the window state so that different windows can't interfere with each other. However unlikely that might seem.
- The list could be combined with the ìsDropping` field as a single value to represent the entire state of the drop event. Something like ``[`NotDropping | `Dropping(list(string))]``. This kind of refactoring is often called "makin illegal states unrepresentable" btw. A good phrase to google.

src/Core/Window.re
```diff
@@ -374,15 +377,32 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
+      Log.error("Received drop complete event without preceding drop events");
+    }
+  /* Event.dispatch(
+       v.onFileDropped,
+       {
+         mouseX: float_of_int(x),
+         mouseY: float_of_int(y),
+         path: Option.value(file, ~default=""),
+       },
+     ) */
```
glennsl, line 405: Why is this commented out?

src/Core/Window.re
zbaylin: File is `NULL` in C with `DROPBEGIN` and `DROPCOMPLETE` events. SDL2 decided to use the same event structure for all drop events, so I just made it an option in reason-sdl2. The `file` field should **never** be null in `DROPFILE`, so maybe it makes more sense to use `Option.get` which throws an exception (`Not_found`) when the option is `None`

Let me know your thoughts!

src/Core/Window.re
zbaylin: Sounds good, and thanks for the reference!

src/Core/Window.re
zbaylin: I agree with you! I think we can change `window.isDropping` to `window.dropState` with a variant like you described. I'll work on that asap!

src/Core/Window.re
zbaylin: My bad -- this was relic code from the PR that had weird behavior with multiple files. I'll get rid of it.

src/Core/Window.re
zbaylin: Interesting -- I guess it makes sense but I always assumed OCaml's lists were backed by arrays. Thanks for the info. I can change that now.

zbaylin:
@glennsl my latest commits address all your previous comments I think -- let me know if I missed anything! I'm curious to know what you think of how I restructured the drop state API.

Thanks for all the notes!

src/Core/Window.re
glennsl: Not at all. And it's not just another collection data structure, it is fundamentally linked with the design of the language. It goes all the way back to LISP and and its "cons" cell, which was one of its most fundamental building blocks. In LISP it was basically just a pair, a tuple of two values. Like OCaml it had functions to access the first and second value, but instead of `fst` and `snd` they were called `CAR` and `CDR`, short for "Contents of the Address Register" and "Contents of the Decrement Register", which hints that this was pretty fundamental even to the computer architecture. To this day most lisps still use this convoluted and confusing naming scheme. Possibly because those weren't the only utility functions for cons cells. You also had `caar`, `cadr`, `cddr`, `caadar`, `cdddar` and so on, which are sequences of `car` and `cdr` calls corresponding to the position of `a`s and `d`s in the middle. Why all these convoluted functions for such a simple little data structure? Because it's incredibly useful and efficient. At least if that's the only primitive compound data structure you have, and you have to build all kinds of custom data structures with it.

The list `[1, 2, 3]` was represented in LISP as `(1, (2, (3, nil)))`, and to get the second element you might do `(car (cdr list))` or `(cadr list)`. To get an element by index dynamically you'd have to apply `car` and `cdr` to it recursively. The fact that it's a _recursive_ data structure makes it particularly suitable for a functional language which has recursive algorithms as a fundamental primitive.

Fortunately someone eventually discovered types, and in particular sum types, so the definition of `list` in OCaml is something like:

```reason
type list('a) = Cons('a, list('a)) | Nil;
```

Can you see the similarity with the `cons` cell and `nil` in LISP?

But there's more. Because of immutability parts of these data structures can be reused. In particular, when prepending an item to an existing list, the existing list can be reused in its entirety. Because it's guaranteed to not change. All you need to do is allocate one cons cell, then put the value in the first slot and a pointer to the existing list in the second. That's it, which makes prepending extremely cheap. On the other hand, to append you have to allocate a new cons cell for the new item, then for the last value of the existing list to have it point to the new cons cell, because you can't mutate it remember, then the next to last to point to the new last cell and so on all the way through the entire list.

So that's why lists are cool and appending is not!

glennsl: Also, I'm sure this is better explained in a million other places on the web. But I was on a roll...

src/Core/Window.re
glennsl: Ideally this would be abstracted away already in `reason-sdl2`. If the file field is `null`, which it should never be, just don't emit an event. Is that possible?

src/Core/Window.re
```diff
@@ -43,6 +43,10 @@ module WindowMetrics = {
   };
 };
 
+type dropState =
+  | Idle
+  | Dropping(list(string));
```
glennsl, line 48: `Idle` might be a bit too general for a variant case at the top level, as it could collide with other variant cases, or lead to them having to be awkwardly named to try to avoid collision and still follow the same kind of convention. Imagine adding another kind of event `state` that could also be in an `Idle` state... That's why I suggested using a polymorphic variant type instead, as those are structurally typed, hence they don't collide. And they're usually fine to use internally and for short-lived and ad hoc values.

zbaylin: > Other than the minor note on the variant type below, this is exactly how I imagined it! What do you think about it? Do you find this style more or less intuitive now that you know it's actually a possibility? It looks like it came quite naturally to you.

Haha thanks! I've written enough code like this before to know what I was doing. I guess I've just been writing so much C for school recently that my mind first went to the first approach. It's definitely more intuitive for OCaml/Reason. Sorry that I didn't think of it sooner!

zbaylin: I can change it to something other than `Idle`. The reason I didn't make it polymorphic was because I couldn't think of a situation where it wouldn't be a binary on/off (dropping or not dropping). I can change it if you think a polymorphic variant fits better (you definitely know better than me haha).

src/Core/Window.re
zbaylin: Hmm I didn't think of it like that. As far as I know right now, all events dispatched from SDL on the C side are isomorphic to those dispatched from the reason-sdl2 side (meaning if a C event is dispatched, a reason event always is). Also, removing the option would mean separate event types for `DROPBEGIN` and `DROPCOMPLETE` (which is totally fine, just something I thought was worth mentioning). Worse comes to worst if we merge this now I/someone else can change how the drop events work upstream in reason-sdl2.

Let me know what you think is best!

examples/FileDragAndDrop.re
```diff
@@ -30,12 +30,15 @@ module Styles = {
...
+  List.map(file => <Text text=file style=Styles.text />, files);
+
 let%component dnd = () => {
-  let%hook (text, setText) = Hooks.state("Nothing dropped yet!");
+  let%hook (files: list(string), setFiles) = Hooks.state([]);
 
   <View style=Styles.outer>
-    <Text style=Styles.text text />
-    <View style=Styles.droppable onFileDropped={e => setText(_ => e.path)}>
+    {React.listToElement(filesToText(files))}
```
bryphe, line 40: Cool 👍

src/Core/Window.re
zbaylin: > Also, I'm sure this is better explained in a million other places on the web. But I was on a roll... 🤷

Haha maybe so but I really enjoyed reading that -- thanks for the history lesson! I wish they taught us this kind of stuff at my university, but alas...

src/Core/Window.re
bryphe: If it's a binary state - it might be reasonable to model it as `option(list(string))` as well - that means you can use the `Option.map`/`Option.iter` methods with it. The downside is it isn't quite explicit what `None` means, but the upside is you can use the full suite of utilities that work with option / maybe monads.

src/Core/Window.re
```diff
@@ -374,15 +379,25 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
...
+  | Sdl2.Event.DropComplete({x, y, _}) =>
+    switch (v.dropState) {
+    | Idle
+    | Dropping([]) =>
+      Log.warn("Received drop complete event without preceding drop events")
+    | Dropping(list) =>
+      v.dropState = Idle;
+      Event.dispatch(
+        v.onFileDropped,
+        {mouseX: float(x), mouseY: float(y), paths: List.rev(list)},
```
bryphe, line 398: Nice! From a revery-consumer perspective, this API is much easier to digest - only one event to worry about instead of three 👍

src/Core/Window.re
bryphe: Good catch @glennsl ! Indeed, we'd want this state to be captured per-window

src/Core/Window.re
glennsl: Yeah `option` fits well here I think. Good call!

src/Core/Window.re
bryphe: Wow, what a great explanation @glennsl !

Understanding the difference between `list` and `array` is important for squeezing the most out of performance / minimizing allocations 👍

src/Core/Window.re
glennsl: Accurately modelling the possible states is generally more important than isomorphism I think. But as this is fairly low-level and hidden from most users, I don't mind merging this before fixing it upstream. I'd still like it noted that this isn't actually a possible state though. Using `Option.value` is definitely safer just in case, but pulling it out into a separate variable and adding a comment explaining that it should never be `None`. Although it might still crash downstream if we somewhere assume it is a valid path.

zbaylin: Not a problem at all! I can make a quick PR to `reason-sdl2` to change that behavior really quickly. Should be very little work.

zbaylin: > but pulling it out into a separate variable and adding a comment explaining that it should never be `None`.

Not quite sure what you mean here -- can you give me an example? Sorry about that!

glennsl: If you're poinf to fix it upstream anyway it doesn't really matter, but I was just thinking this:

```reason
| Some(list) =>
  let file = Option.value(file, ~default=""); // Will always be `Some`
  v.dropState = Some([file, ...list])
```

Just to make it clear that this only applies to `file`

zbaylin: I just made revery-ui/reason-sdl2#69 to address this! @bryphe since you just published 2.10.3027 today you could probably revoke that and republish it with these small changes, but it's (obviously) up to you. Sorry to create all this confusion!

glennsl: Also, while we're not on the subject but recently were and this reminded me of it: One aspect of Reason I don't like is how it intentionally makes lists look like arrays for familiarity (and it's not just lists/arrays it does this with, also functions/currying for example). In OCaml, the syntax for `[file, ...list]` is `file :: list`, and if you have more elements it would be `file1 :: file2 :: list`, which is equivalent to `file1 :: (file2 :: list)`, which makes the underlying cons cell representation much more apparent. `::` is really just the `Cons` constructor as an infix operator. And `[]` is just sugar for `Nil`. In fact, in a fairly recent version of OCaml you're even allowed to override this syntax for your own types using something like this:

```ml
type t('a) =
  | []
  | (::) of 'a * t
```
(which is effectively the actual definition of list)

src/Core/Window.re
```diff
@@ -378,7 +378,7 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
   | Sdl2.Event.DropBegin(_) => v.dropState = Some([])
   | Sdl2.Event.DropFile({file, _}) =>
     switch (v.dropState) {
-    | Some(list) => v.dropState = Some([Option.get(file), ...list])
+    | Some(list) => v.dropState = Some([file, ...list])
```
bryphe, line 381: Nice!

zbaylin:
Alright I've updated to use reason-sdl2@2.10.3028! It was a single line change to the new API, which is always nice :). Once this passes CI I can merge this into the other PR, and merge *that* PR into master 🎆🎆🎆

bryphe:
@zbaylin - if you want, since this is going in the other PR anyway - we can just merge this now, and let the CI run on the 'main' PR. Might be more streamlined than having to wait twice 😄 

I think you have permission to override for this PR, but if not let me know 👍

zbaylin:
@bryphe I realized the same thing after I posted that 😆. Sounds good - I can do that now ✅

PR #834, opened 2020-05-10
fix(input): handle mouse capture properly

glennsl:
**Issue:** Sliders, scrollbars and other mouse capture consumers would not track the mouse correctly when the mouse goes outside the window. This could also sometimes cause the capture to get irreversibly stuck and perhaps other erratic behaviour.

**Defect:** The old capture API would intentionally release capture when the mouse moved outside the window (although this seems to not always have worked correctly) and did not have the ability to track the mouse outside the window anyway.

**Fix:** Use SDL's `SDL_CaptureMouse` function to properly capture the mouse, and release capture only if the window loses focus (or if manually released, of course).

Since there is no "current window" API, this requires the window to be explicitly passed into `setCapture`, which is a breaking change. For users to be able to obtain the current window it is passed through the mouse button event, effectively meaning that the mouse can only be captured during these events, which is consistent with the browser's pointer events API (except instead of being explicit it's just a noop if it is called outside "active button state").

Also, I think the ideal capture API would just continue to send events to the current node, instead of passing in a bunch of random callbacks. That would make it possible for the node to clean it up when it gets unmounted instead of the consumer having to manually do so using an effect hook. Unfortunately I couldn't see how to implement this robustly with the current architecture.

### DIFF
--- a/src/UI/Mouse.re
+++ b/src/UI/Mouse.re
@@ -173,6 +173,7 @@ let noop0 = () => ();
 let noop1 = _ => ();
 
 let releaseCapture = () => {
+  ignore(Sdl2.Mouse.capture(false): int);
   capturedEventStateInstance := None;
 };
 
@@ -189,8 +190,7 @@ let setCapture =
       ~onMouseLeaveWindow=noop0,
       (),
     ) => {
-  // If there was a previous capture - release
-  releaseCapture();
+  ignore(Sdl2.Mouse.capture(true): int);
 
   capturedEventStateInstance :=
     Some({
--- a/src/UI/Mouse.re
+++ b/src/UI/Mouse.re
@@ -10,20 +10,18 @@ module Log = (val Log.withNamespace("Revery.UI.Mouse"));
 module Cursor = {
   /* State needed to track on the cursor */
   type t = {
-    x: ref(float),
-    y: ref(float),
+    mutable x: float,
+    mutable y: float,
   };
 
-  let make = () => {
-    let ret: t = {x: ref(0.), y: ref(0.)};
-    ret;
-  };
+  let make = () =>
+   {x: 0., y: 0.};
 
-  let get = cursor => (cursor.x^, cursor.y^);
+  let get = cursor => (cursor.x, cursor.y);
 
   let set = (~x, ~y, c) => {
-    c.x := x;
-    c.y := y;
+    c.x = x;
+    c.y = y;
   };
 };
 
--- a/src/UI/Mouse.re
+++ b/src/UI/Mouse.re
@@ -61,7 +61,7 @@ let releaseCapturedNode = node =>
   | _ => ()
   };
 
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -53,6 +53,9 @@
     "revery-text-wrap": "revery-ui/revery-text-wrap#966383e",
     "@glennsl/timber": "1.0.0"
   },
+  "resolutions": {
+    "reason-sdl2": "revery-ui/reason-sdl2#355cb28"
+  },
```
glennsl, line 58: TODO: Remove. Need a new release of reason-sdl2

glennsl:
Actually, I'm going to find a way to make the node API work, because I've scratched my head for a while now trying to get the callback approach to work, and it just doesn't without nasty hacks to work around that the callbacks will close over the environment when the capture is initiated, which is very counter-intuitive from a React POV.

src/UI_Components/Slider.re
```diff
@@ -69,71 +68,44 @@ let%component make =
...
-            if (isCaptured) {
-              Mouse.releaseCapture();
-            },
-        );
-      },
+          Some(origin);
+        },
+      ~onMouseUp=(_, _) => None,
+      (),
     );
```
glennsl, line 91: This hook API is pretty nice I think. It requires the initial state to be passed when calling `captureMouse`, then passes that to each capture event handler, which can either return a new state or `None` or release the capture.

src/UI/Mouse.re
```diff
@@ -10,261 +10,103 @@ module Log = (val Log.withNamespace("Revery.UI.Mouse"));
...
-      onMouseOver(evt);
-      true;
-    | MouseOut(evt) =>
-      onMouseOut(evt);
-      true;
-    | _ => false
-    }
-  };
-};
+let releaseCapture = Capture.release;
```
bryphe, line 108: Thanks for the clean-up here, @glennsl ! It looks much simpler now.

src/UI_Components/Slider.re
bryphe: I agree - this is much nicer than what we had before, and addresses the closure issue. Thanks for improving this API!

test/UI/MouseTest.re
```diff
@@ -528,34 +545,57 @@ describe("Mouse", ({describe, test, _}) => {
...
 
-      Mouse.dispatch(cursor, InternalMouseDown({button: BUTTON_LEFT}), node);
+      expect.int(uncapturedMoves^).toBe(1);
+      expect.int(capturedMoves^).toBe(0);
 
-      expect.int(nodeCount^).toBe(0);
-      expect.int(captureCount^).toBe(1);
+      Mouse.setCapture(
+        ~onMouseMove=_ => incr(capturedMoves),
+        Window.create("", WindowCreateOptions.default),
```
bryphe, line 577: Thanks for updating  and improving these tests!

PR #835, opened 2020-05-11
Native: Add Shell module with openItem function

zbaylin:
Similar to https://www.electronjs.org/docs/api/shell#shellopenitemfullpath

### DIFF
--- a/src/Native/shell_cocoa.c
+++ b/src/Native/shell_cocoa.c
@@ -1,22 +1,29 @@
 #ifdef __APPLE__
 #import <Cocoa/Cocoa.h>
 
-int revery_openItemAtURL_cocoa(const char *url_string) {
+int revery_openURL_cocoa(const char *url_string) {
     NSString *nsString = [NSString stringWithCString:url_string encoding:NSUTF8StringEncoding];
-
-    // This is required for non "file://"-prefixed URLs to work
-    NSURL *nsURL;
-    if([nsString characterAtIndex:0] == '/') {
-        nsURL = [NSURL fileURLWithPath:nsString];
-    } else {
-        NSString *nsEncodedString = [nsString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLFragmentAllowedCharacterSet]];
-        nsURL = [NSURL URLWithString:nsEncodedString];
-        [nsEncodedString release];
-    }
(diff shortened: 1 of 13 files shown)


zbaylin:
Alright I've now implemented this on Windows, macOS, and Linux! 🎉  It's universally a very simple API, which is always nice. Pending review it's ready to be merged.

src/Native/shell_cocoa.c
```diff
@@ -0,0 +1,23 @@
...
+#import <Cocoa/Cocoa.h>
+
+int revery_openItemAtURL_cocoa(const char *url_string) {
+    NSString *nsString = [NSString stringWithCString:url_string encoding:NSUTF8StringEncoding];
+
+    // This is required for non "file://"-prefixed URLs to work
+    NSURL *nsURL;
+    if([nsString characterAtIndex:0] == '/') {
+        nsURL = [NSURL fileURLWithPath:nsString];
+    } else {
```
glennsl, line 11: A file path that is not prefixed with `file://` is not a URL. So either this should be handled at a higher level so that the function is passed a valid URL, or the function name should be changed to reflect what it actually does. Approximate naming just leads to confusing APIs where it's not clear where the exceptions are actually handled without following the code.

zbaylin: That's fine, I can get rid of it. The reason I did that was because Windows will handle absolute paths without having to prepend `file://` to the path, but it simplifies things for me so that's totally okay haha. I can separate the API into `Shell.openFile` and `Shell.openURL` if that works.

bryphe: That sounds reasonable - Electron has separate APIs for opening a path (https://www.electronjs.org/docs/api/shell#shellopenitemfullpath) and URL (https://www.electronjs.org/docs/api/shell#shellopenexternalurl-options)

zbaylin:
Alright I separated the APIs into `openFile` and `openURL` and tested it on all three OSes

glennsl:
:100:

zbaylin:
I'm not sure why the Linux CI is failing. I've included GTK in the file but it's giving me an implicit declaration warning? Could it be the version of GTK on the CI's VM?

zbaylin:
These last two commits should fix CI. I had to switch to `gtk_show_uri` since `gtk_show_uri_on_window` is pretty new and isn't widely supported.

PR #840, opened 2020-05-13
refactor(window): Update Window API to differentiate between scaled/unscaled sizes

bryphe:
Motivated by https://github.com/onivim/oni2/pull/1752 - this addresses an inconsistency in the Revery `Window` API in talking about sizes - there are two window sizes that SDL gives us - the 'screen space' size, and the 'framebuffer' size. On top of that, we also account for display scaling (like frameworks like Electron handle). So that makes three 'spaces' for window sizes:

1) Screen space size
2) Screen space size, _scaled based on display settings_
3) Framebuffer pixel size (the size of the 'backbuffer' that we render to - this is in pixels can vary depending on the device pixel ratio of the screen. For example, retina displays would be some multiple of the screen space size).

For the purposes of window management, we only care about 1 & 2 - framebuffer size is important for rendering though.

We run into case 2 in Linux with `GDK_SCALE` applied, or zoomed display settings on Windows. When the `scaleFactor` is 1, cases 1/2 are equivalent.

The problem was, the API surface wasn't consistent in handling scaling - `create`, `getSize`, and `setSize` should all be consistent in the space they use - which they weren't. `create` would use coordinates in the 'display scaled' space (2), but getSize/getRawSize would get the dimensions in unscaled space (1) - and `setSize` would also retrieve dimensions in unscaled space. This is confusing and inconsistent, and caused problems in https://github.com/onivim/oni2/pull/1752, because we'd use `getSize` to read the window dimensions (without scaling applied), and then on startup we'd apply scaling to them, causing the window to grow in size, because each `create` would re-apply scaling.

__TODO:__
- [x] Test on Linux with `GDK_SCALE` set
- [x] Test on retina OSX display
- [x] Test on non-retina OSX display
- [x] Test on unscaled Windows display
- [x] Test on Windows display with display scaling
(test example app, example with zoom, Onivim 2 text rendering, Onivim 2 persistence round trip)

### DIFF
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -2,6 +2,7 @@ open Events;
 
 type unsubscribe = unit => unit;
 
+[@deriving show({with_path: false})]
 type size =
   Sdl2.Size.t = {
     width: int,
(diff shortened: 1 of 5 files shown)


src/Core/Window.re
```diff
@@ -101,6 +113,135 @@ module Internal = {
...
+          ~forceScaleFactor=w.forceScaleFactor,
+          w.sdlWindow,
+        ),
+      zoom: previousZoom,
+    };
+    w.areMetricsDirty = false;
+    Log.trace(
+      "updateMetrics - new metrics: " ++ WindowMetrics.toString(w.metrics),
+    );
+  };
```
glennsl, line 205: This seems like a very error-prone way of dealing with the metrics. Firstly because there are a bunch of invariants here that aren't checked at all. Secondly because everything can be changed anytime, anywhere. Thirdly because there's no restriction on using invalid/dirty metrics. And fourthly (?) because it's not clear what can and should be changed directly and what will be computed, because everything's just been thrown in the same heap.

So I would suggest that the type itself is made private so it can't be manipulated directly, to move the dirty flag into the record, and to expose functions to set the fields that can be changed, update the metrics, set the dirty flag automatically, update the metrics and maintain every invariant. And to make sure invalid data isn't used it might be an idea to replace the dirty flag with a "state" variant with the metrics as its payload.

src/Core/Window.re
```diff
@@ -15,29 +15,41 @@ module WindowMetrics = {
      * and the _framebuffer_ width/height
      * Some more info here: http://www.glfw.org/docs/latest/window_guide.html
      */
```
glennsl, line 17: This comment is outdated

src/Core/Window.re
```diff
@@ -101,6 +113,135 @@ module Internal = {
...
+        });
+      };
+    };
+  };
+
+  let resizeIfNecessary = (w: t) =>
+    switch (w.requestedWidth, w.requestedHeight) {
+    | (Some(width), Some(height)) => setRawSize(w, width, height)
+    | _ => ()
+    };
```
glennsl, line 244: `requestedWidth` and `requestedHeight` is treated as a single field. So it should be a single field.

src/Core/Window.rei
```diff
@@ -38,9 +38,31 @@ let onFileDropped: (t, fileDropEvent => unit) => unsubscribe;
...
+  will actually set the window to a height of 800 pixels wide by 600 pixels high. This is usually
+  the behavior you want. To directly set the size, without considering display scaling, use [setUnscaledSize]
+*/
+let setSize: (~width: int, ~height: int, t) => unit;
+
+/**
+  [setUnscaledSize(~width, ~height, window)] sets the window size, without accounting for display scaling.
+*/
+let setUnscaledSize: (~width: int, ~height: int, t) => unit;
+
```
glennsl, line 65: Very nice API and doc comments. But why can you set the size unscaled? When will that be useful?

src/UI/Render.re
```diff
@@ -28,7 +28,7 @@ let render =
   );
 
   /* Layout */
-  let size = Window.getRawSize(window);
+  let size = Window.getUnscaledSize(window);
   let pixelRatio = Window.getDevicePixelRatio(window);
   let scaleAndZoomFactor = Window.getScaleAndZoom(window);
   let canvasScalingFactor = pixelRatio *. scaleAndZoomFactor;
```
glennsl, line 34: Why is `getUnscaledSize` used here, and the immediately scaled? Instead of just using `getSize` directly.

src/Core/Window.rei
bryphe: Good point. I'll simply remove this from the public API to avoid confusion.

src/Core/Window.re
bryphe: Cleaned this up a bit

src/UI/Render.re
bryphe: Good catch, seems like we can do the same calculation with scaled size / zoom factor.

src/Core/Window.re
bryphe: I agree 100% that this could be improved and with your feedback. I'm just concerned about the additional risk of changing this (it looks like new code... but it's actually just moved) and introducing an additional bug / regression (when we're already trying to fix a regression).

I have to work on another fix today too for syntax highlighting (been a pretty rough week for regressions...) but I can see a few options:

- __Option 1:__ We can defer this once we've dug ourselves out of the hole of regressions and I can test it a bit more thoroughly, and bring in what we have now to unblock a fix for the persistence, and refactor this, later as a separate change, once we're in a better state.

- __Option 2:__ I can work on this, but I might not complete it today - which means we may want to revert the width/height persistence piece in Onivim 2, and then once we've addressed this and are confident in the fix - we could re-enable the window width/height persistence in Onivim 2.

- __Option 3:__ if you want to give this a shot, and you have time today - I can do my test pass with those changes... and fingers crossed everything just works... 🤞 

Let me know what you'd prefer @glennsl

glennsl: Most of the suggestions aren't all that risky as it's just making the "contract" explicit. I.e. instead. of manipulating the record directly wherever, add a function to do exactly the same, but with the direct manipulation being internal. So it could be done incrementally as well if you'd prefer, and just doing that would resolve three of the four points I think

Otherwise I'm not working on anything particularly time-sensitive, so I could certainly work on this instead if you want.

src/Core/Window.re
```diff
@@ -637,14 +664,9 @@ let hide = w => {
...
-let getSize = window => Sdl2.Window.getSize(window.sdlWindow);
-
-let getRawSize = (w: t) => {
-  let width = w.metrics.size.width;
-  let height = w.metrics.size.height;
+let getSize = ({metrics, _}) => metrics.scaledSize;
 
-  {width, height};
-};
+let getUnscaledSize = ({metrics, _}) => metrics.unscaledSize;
```
glennsl, line 669: This doesn't seem to be used anymore. Should produce a warning.

src/Core/Window.re
```diff
@@ -148,147 +285,34 @@ let setTitle = (v: t, title: string) => {
...
 
-  setRawSize(win, adjWidth, adjHeight);
+  Internal.setRawSize(win, adjWidth, adjHeight);
+};
+
+let setUnscaledSize = (~width: int, ~height: int, win: t) => {
+  Log.tracef(m => m("setUnscaledSize - calling with: %ux%u", width, height));
+
+  Internal.setRawSize(win, width, height);
 };
```
glennsl, line 304: This doesn't seem to be used either.

src/Core/Window.re
bryphe: > Otherwise I'm not working on anything particularly time-sensitive, so I could certainly work on this instead if you want.

If you wanted it to give it a shot - that would be really helpful for me! I'll come back and test this in a bit

src/Core/Window.re
bryphe: Ah, right, it looks like `esy dune build @check` isn't running on CI - added it 👍

src/Core/Window.re
bryphe: Cool, thanks for taking care of this @glennsl ! I did a sanity check and seems like everything is good - going to work on the full test pass now.

bryphe:
Tested across platforms with scaling - looks like everything is still good 👍

PR #841, opened 2020-05-13
refcactor: encapsulate WindowMetrics better

glennsl:
As discussed in #840 (and merges into that, of course)

This does everything I suggested, except replacing the dirty flag with a variant to prevent invalid data from being accessed, as that would be pretty invasive. All this currently does is move code around to make the contract explicit.


bryphe:
Looks great to me, @glennsl - thanks for the help, appreciate it!

PR #844, opened 2020-05-14
refactor(dependency): Remove rely as top-level dependency

bryphe:


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "esy b",
-    "build:release": "esy b dune build --profile=release --root . -j4",
+    "build:release": "esy b dune build -p Revery -j4",
```
Et7f3, line 11: it is unrelated ?

bryphe: This scopes it so that the package we consume is just the top-level `Revery` package, and not the test runner / benchmark application. If we keep this with `--root .`, we'll still need the top-level `rely` dependency, because it'll be build the test runner too.

The motivation for this change is I had a collision with some overrides with `rely` in Onivim 2

PR #845, opened 2020-05-14
fix(test/#843): Fix windows output for test cases

### LINKED ISSUE #843: Regression: test doesn't display something
image

- Operating System: Windows
- Revery Version: 4e85cf05b82b5a92b8b8076d51b8ff899896fda4
- OCaml version: 4.08.1

- Link to github repo: https://github.com/revery-ui/revery

- Steps to reproduce: git clone & esy @test run

- Actual Result: nothing is displayed
- Expected Result: rely output should be seen

- Additional Information: 

- Search terms used:

bryphe:
__Issue:__ `esy @test run` stopped producing output on all the windows devices I tried.

__Defect:__ Linking in `reason-sdl2` brings in the `-mwindows` flag, which causes the application to not allocate or attach console.

__Fix:__ Explicitly attach a console for the test runner.

Fixes #843  - I guess we both hit this at the same time, @Et7f3 !

### DIFF
--- a/test/TestRunner.re
+++ b/test/TestRunner.re
@@ -1,3 +1,10 @@
+// On Windows, because this is linked against the '-mwindows' flag
+// (for a GUI app), we need to manually allocate a console.
+if (Sys.win32) {
+  let _: int = Sdl2.Platform.win32AttachConsole();
+  Pastel.setMode(Pastel.Disabled);
+};
+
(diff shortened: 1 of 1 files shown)


Et7f3:
Ah yeah make sense. I think I will upstream to rely.

PR #846, opened 2020-05-15
feat(components): Add TextButton and Link components

zbaylin:
@bryphe and I were discussing adding a Link component that takes an `href` now that we have `Native.Shell.openURL`. However, I considered the fact that in many cases one wouldn't want to trigger side effects from the UI (like in Oni2), so I made a `TextButton` component that's essentially the same thing except it takes an `onClick` instead.


zbaylin:
I just made examples for both components:
- `Link` in `Hello.re` -- a "View on GitHub" button
- `TextButton` in `FileDragDrop.re` -- all dropped files are linkable/clickable

bryphe:
This is awesome Zach - thanks for adding this!

src/UI_Components/TextButton.re
```diff
@@ -0,0 +1,20 @@
+open Revery_UI;
+open Revery_Core;
+open Revery_UI_Primitives;
+
+module Hooks = Revery_UI_Hooks;
+
+let noop = () => ();
+
+let%component make = (~text, ~inactiveStyle, ~activeStyle, ~onClick=noop, ()) => {
```
bryphe, line 9: One minor naming thing - the `Button` in `TextButton` makes me think of a button element, with a border, etc - maybe something like `ClickableText` would be more accurate?

src/UI_Components/TextButton.re
```diff
@@ -0,0 +1,20 @@
...
+  let%hook (isHovered, setHovered) = Hooks.state(false);
+
+  let outerStyle = Style.[cursor(MouseCursors.pointer)];
+  let onMouseEnter = _ => setHovered(_ => true);
+  let onMouseLeave = _ => setHovered(_ => false);
+  let onMouseUp = _ => onClick();
+
+  <View style=outerStyle onMouseUp onMouseEnter onMouseLeave>
+    <Text text style={isHovered ? activeStyle : inactiveStyle} />
+  </View>;
```
bryphe, line 19: I like the way you split up the functionality between this and `<Link />` - good idea!

src/UI_Components/TextButton.re
glennsl: Or ´LinkButton`? That has precedent from ASP.NET at least.

src/UI_Components/TextButton.re
```diff
@@ -0,0 +1,20 @@
...
+module Hooks = Revery_UI_Hooks;
+
+let noop = () => ();
+
+let%component make = (~text, ~inactiveStyle, ~activeStyle, ~onClick=noop, ()) => {
+  let%hook (isHovered, setHovered) = Hooks.state(false);
+
+  let outerStyle = Style.[cursor(MouseCursors.pointer)];
+  let onMouseEnter = _ => setHovered(_ => true);
+  let onMouseLeave = _ => setHovered(_ => false);
```
glennsl, line 14: This is so common it ought to be made into a hook. Something like:

```
let%hook (isHovered, onMouseEnter, onMouseLeave) = Hooks.hover();
```

Not a request, just a thought that flew past my head just now, that I felt ought to be written down somewhere.

src/UI_Components/TextButton.re
zbaylin: I think `ClickableText` is the clearest name. I'll change it now.

PR #848, opened 2020-05-19
refactor(dependency): reason-sdl2 -> 2.10.3032

bryphe:


zbaylin:
This can be closed since I bumped the `reason-sdl2` dependency in #850

PR #849, opened 2020-05-19
feat(macOS): allow opening of files through the AppDelegate

zbaylin:
This adds a `onFileOpen` Event API to `App` which is dispatched whenever a file is opened through the AppDelegate.

This is kind of a tricky one to test, but if you're trying to do so I recommend using Oni2.


src/Native/cocoa/ReveryAppDelegate_func.c
```diff
@@ -0,0 +1,28 @@
...
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
+#include <caml/threads.h>
+#include <caml/alloc.h>
+
+#include "caml_values.h"
+
+#include "utilities.h"
+
+void appDelegate_openFile(const char *path) {
```
bryphe, line 14: We may want to prefix this with `revery_*`, just to minimize the chance of global name collisions

src/Native/Revery_Native.re
```diff
@@ -4,3 +4,5 @@ module Notification = Notification;
 module Shell = Shell;
 
 include Initialization;
+
+external log: string => unit = "revery_log";
```
bryphe, line 8: Would be great to add a comment here about how this is different from `Console`, or `print_endline` - that it is specifically logging to the system log. 

It doesn't look its used either, so we could potentially remove it for now

src/Native/cocoa/ReveryAppDelegate.c
```diff
@@ -59,6 +61,19 @@
     UNUSED(notification);
     return YES;
 }
+
+/* openFile
+  We call into the CAML function `revery_dispatchFileOpen`
+  Unfortunately because of namespacing issues with `alloc`,
```
bryphe, line 67: Interesting! What was the issue with the alloc collision?

src/Native/cocoa/ReveryAppDelegate_func.c
```diff
@@ -0,0 +1,28 @@
...
+    static const value *dispatchFileOpen = NULL;
+    if (dispatchFileOpen == NULL) {
+        dispatchFileOpen = caml_named_value("revery_dispatchFileOpen");
+    }
+    // Call only if the value was gotten
+    if (dispatchFileOpen != NULL) {
+        caml_c_thread_register();
+        caml_acquire_runtime_system();
```
bryphe, line 22: Nice work wiring this up 👍

src/Native/cocoa/ReveryAppDelegate_func.c
```diff
@@ -0,0 +1,28 @@
...
+        caml_callback(*dispatchFileOpen, caml_copy_string(path));
```
bryphe, line 23: Since we're allocating here - do we need the `CAMLParam`/`CAMLreturn` macros? 🤔

src/Core/App.rei
```diff
@@ -62,6 +62,13 @@ let onBeforeQuit: (t, unit => unit) => unsubscribe;
 */
 let onIdle: (t, unit => unit) => unsubscribe;
 
+/** [onFileOpen(app, f) registers a callback [f] that is called when the host OS is trying to open a file]
+
+  Note that as of now, this function only is called on macOS, as other OS's have other
+  mechanisms for opening files (usually just through CLI args)
+*/
+let onFileOpen: (t, string => unit) => unsubscribe;
```
bryphe, line 70: Thanks for the descriptive comment here 👍

src/Native/cocoa/ReveryAppDelegate.c
```diff
@@ -59,6 +61,19 @@
...
+  we have to put the function in a separate file.
+*/
+- (BOOL)application:(NSApplication *)sender
```
bryphe, line 70: Do we need to handle `openFiles` as well?

src/Native/Revery_Native.re
zbaylin: Oh my bad I left this in there from my debugging yesterday. I'll get rid of it now.

src/Native/cocoa/ReveryAppDelegate.c
zbaylin: Both OCaml and Obj-C have functions named `alloc`, which confuses the Objective-C runtime leading to either build errors (preferable) or runtime errors (not so preferable).

src/Native/cocoa/ReveryAppDelegate.c
zbaylin: If `openFiles` is undefined then `openFile` is just called n times. Up to you if you want to have a separate Revery event to handle the case where multiple files are opened but I couldnt think of a reason to.

src/Native/cocoa/ReveryAppDelegate_func.c
bryphe, line 23: One thing we could consider, to be safe, is refactor the ocaml-runtime stuff to a helper method, so it would look like here:
```
caml_c_thread_register();
caml_acquire_runtime_system();
revery_dispatch_openFile(path);
caml_release_runtime_system();
```

And then:
```
CAMLprim value revery_dispatch_openFile(const char *path) {
   CAMLparam0(); // The parameter isn't an ocaml value, so don't need to wrap
   CAMLlocal1(str);
   str = caml_copy_string(path);
   
    static const value *dispatchFileOpen = NULL;
    if (dispatchFileOpen == NULL) {
        dispatchFileOpen = caml_named_value("revery_dispatchFileOpen");
    }
     caml_callback(*dispatchFileOpen, str);
      CAMLreturn(Val_unit);
}
```

bryphe: I don't think this would harm anything, and it could be safer

src/Native/cocoa/ReveryAppDelegate.c
bryphe: Oh, cool! That sounds perfect 👍

src/Native/cocoa/ReveryAppDelegate.c
bryphe: ouch! Nice catch.

It'd be helpful to me to add that extra detail here (that `alloc` is a collision between Obj-C `alloc` and OCaml `alloc`) just to jog my memory next I read through.

zbaylin: Sounds good! This is something a bunch of people have asked about on Discord too when writing their own native bindings. Definitely an annoying state of affairs!

src/Native/cocoa/ReveryAppDelegate_func.c
zbaylin: Great idea! Stupid question -- since this is already a helper function though, is there any harm in just adding the `CAMLparam0` and `CAMLreturn` calls directly to this one? Im probably missing something but I cant think of what.

Let me know! Thanks again!

bryphe: I believe the macros might depend on the runtime already being acquired, but I could be mistaken. 

A good way to test this to add a `Gc.full_major()` call in the handler - we wouldn't want to check that in, but it can exercise a crash when the primitives aren't used properly - might be a good litmus test

zbaylin: I just pushed something very similar to what you just suggested. Let me know if that works!

PR #853, opened 2020-05-20
feat(window): bind Node mouseBehavior to SDL hit-tests

zbaylin:
This adds an attribute to all nodes that changes their mouse
behavior based on SDL2's hit test results.

i.e.
```reason
<View
  mouseBehavior=Draggable
  ...
/>
```

GIF:
revery-hit-tests-example

I want to make sure this works cross-platform before I merge it in. I've tested in on macOS and can test on Linux. @bryphe would you be able to test on Windows? That would be super helpful 😃.

- [x] Test on macOS
  - [x] `Draggable`
  - [ ] `Resize`
- [x] Test on Linux
  - [x] `Draggable`
  - [x] `Resize`
- [x] Test on Windows
  - [x] `Draggable`
  - [x] `Resize`

From what I can tell the `Resize` types don't work on macOS. I don't think this is evidence of something wrong either here or in `reason-sdl2`. I'll do some research to see if they're supposed to work.

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -155,6 +155,11 @@ let examples = [
     render: _ => URLFileOpen.render(),
     source: "URLFileOpen.re",
   },
+  {
+    name: "Window: Hit Tests/Zones",
+    render: _ => HitTests.render(),
+    source: "HitTests.re",
+  },
 ];
 
 let getExampleByName = name =>
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -157,7 +157,7 @@ let examples = [
   },
   {
     name: "Window: Hit Tests/Zones",
-    render: _ => HitTests.render(),
+    render: w => HitTests.render(w),
     source: "HitTests.re",
   },
 ];
@@ -271,13 +271,15 @@ let init = app => {
   |> (ignore: Revery.App.unsubscribe => unit);
 
   let initialExample = ref("Animation");
+  let decorated = ref(true);
   Arg.parse(
     [
       ("--trace", Unit(() => Timber.App.setLevel(Timber.Level.trace)), ""),
+      ("--no-decoration", Unit(() => decorated := false), ""),
       ("--example", String(name => initialExample := name), ""),
     ],
     _ => (),
-    "There is only --trace and --example",
+    "There is only --trace, --example, and --no-decoration",
   );
   let initialExample = initialExample^;
 
@@ -298,6 +300,7 @@ let init = app => {
           ~maximized,
           ~titlebarStyle=Transparent,
           ~icon=Some("revery-icon.png"),
+          ~decorated=decorated^,
           (),
         ),
       app,
--- a/examples/HitTests.re
+++ b/examples/HitTests.re
@@ -1,0 +1,44 @@
+open Revery;
+open Revery.UI;
+
+module Styles = {
+  open Style;
+
+  let outer = [
+    position(`Absolute),
+    top(0),
+    bottom(0),
+    left(0),
+    right(0),
+    justifyContent(`Center),
+    alignItems(`Center),
+  ];
+
+  let draggable = [
+    width(100),
+    height(100),
+    backgroundColor(Colors.blue),
+    justifyContent(`Center),
+    alignItems(`Center),
+  ];
+
+  let text =
+    Style.[
+      color(Colors.white),
+      fontFamily("Roboto-Regular.ttf"),
+      fontSize(14.),
+    ];
+};
+
+let hitTests = () =>
+  <View style=Styles.outer>
+    <View style=Styles.draggable mouseBehavior=Draggable>
+      <Text
+        text="Drag window from here!"
+        style=Styles.text
+        mouseBehavior=Draggable
+      />
+    </View>
+  </View>;
+
+let render = () => <hitTests />;
--- a/examples/HitTests.re
+++ b/examples/HitTests.re
@@ -32,20 +32,18 @@ module Styles = {
 
 let hitTests = (~window, ()) =>
   <View style=Styles.outer>
-    {
-      Window.isDecorated(window) ?
-        <Text
-          text="Window decorations must not be enabled.\nRestart with --no-decoration"
-          style=Styles.text
-        /> :
-        <View style=Styles.draggable mouseBehavior=Draggable>
-          <Text
-            text="Drag window from here!"
-            style=Styles.text
-            mouseBehavior=Draggable
-          />
-        </View>
-    }
+    {Window.isDecorated(window)
+       ? <Text
+           text="Window decorations must not be enabled.\nRestart with --no-decoration"
+           style=Styles.text
+         />
+       : <View style=Styles.draggable mouseBehavior=Draggable>
+           <Text
+             text="Drag window from here!"
+             style=Styles.text
+             mouseBehavior=Draggable
+           />
+         </View>}
   </View>;
 
 let render = window => <hitTests window />;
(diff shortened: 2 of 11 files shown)


zbaylin:
Just tested it on Linux and both work there!

src/UI/Ui.re
```diff
@@ -36,6 +49,11 @@ let start = (window: Window.t, element: React.element(React.reveryNode)) => {
   let container = Container.create(rootNode);
   let ui = RenderContainer.create(window, rootNode, container, mouseCursor);
 
+  Sdl2.Window.setHitTest(
+    Window.getSdlWindow(window),
+    HitTest.windowCallback(rootNode, window),
+  );
```
bryphe, line 55: I wonder if we should make this an option on WindowCreation? The drawback with this, at least on Linux and Windows, is you have to mark your own areas for resize at the borders of the window.

src/UI/Ui.re
```diff
@@ -20,6 +20,19 @@ let _activeWindow: ref(option(Window.t)) = ref(None);
 
 type renderFunction = React.element(React.reveryNode) => unit;
 
+type mouseBehvaior =
```
bryphe, line 23: Minor type: `mouseBehvaior` -> `mouseBehavior`

src/UI/Ui.re
```diff
@@ -20,6 +20,19 @@ let _activeWindow: ref(option(Window.t)) = ref(None);
...
+    | Normal
+    | Draggable
+    | ResizeTopLeft
+    | ResizeTop
+    | ResizeTopRight
+    | ResizeRight
+    | ResizeBottomRight
+    | ResizeBottom
+    | ResizeBottomLeft
+    | ResizeLeft;
```
bryphe, line 34: Thanks for bringing these over!

src/UI/Ui.re
zbaylin: Oof my bad! Thanks for pointing that out! I'm surprised my spell-checker didn't catch that for me 🤔

src/UI/HitTest.re
```diff
@@ -0,0 +1,21 @@
...
+
+module Log = (val Revery_Core.Log.withNamespace("Revery.Ui.HitTest"));
+
+let windowCallback = (node, window) =>
+  Some(
+    (sdlWindow, mouseX, mouseY) =>
+      Sdl2.Window.(
+        if (sdlWindow == Window.getSdlWindow(window)) {
+          let deepestNode =
+            getTopMostNode(node, float(mouseX), float(mouseY));
```
bryphe, line 12: I'm seeing intermittent exceptions when starting up with this PR, like this: 
```
[DEBUG]    +0ms Revery.Core.Window : Metrics: { scaledSize = { width = 800; height = 480 };
  unscaledSize = { width = 800; height = 480 };
  framebufferSize = { width = 800; height = 480 }; devicePixelRatio = 1.;
  scaleFactor = 1.; zoom = 1.; isDirty = false }
[INFO]     +0ms Revery.Core.Window : Using vsync: Synchronized
Fatal error: exception Revery_UI__Node.NoDataException("getBoundingBoxClipped")
```

I think what's happening is that we can't make this query if we haven't actually done layout yet - so there is a race between the first mouse event and layout occurring.

We should check node#hasRendered() prior to running this (maybe it makes sense to just push that into `getTopMostNode`), so that we can ensure that layout happening and we have the node dimensions before running this

examples/HitTests.re
```diff
@@ -0,0 +1,44 @@
...
+    ];
+};
+
+let hitTests = () =>
+  <View style=Styles.outer>
+    <View style=Styles.draggable mouseBehavior=Draggable>
+      <Text
+        text="Drag window from here!"
+        style=Styles.text
+        mouseBehavior=Draggable
```
bryphe, line 39: Thanks for adding the example, Zach!

src/UI/HitTest.re
zbaylin: I think I got that exception once but I couldn't replicate it, but your explanation makes total sense. I'll add that really quick!

src/UI/Ui.re
bryphe: Actually, that might only be if we make the window undecorated, and unrelated to this - testing though on Windows 👍

zbaylin: Oh I wasn't sure what you mean at first but now I get it. Yeah, I'm pretty sure just enabling the hit tests doesn't change any default behavior -- that only occurs if you remove the decorations. Linux also doesn't change, so I think we're good on that.

src/UI/HitTest.re
zbaylin: Fixed in the latest commit!

src/UI/Ui.re
zbaylin: Fixed in the latest commit!

src/UI/Ui.re
bryphe: Unfortunately, on Windows - it actually does disable the default resize behavior of the window. The window is still draggable with the titlebar, but it is unable to be resized when you move the edge of the screen.

So I think we will need to make this a toggleable option on the window - whether to support custom mouse behavior.

The only other idea I have is we could add default `<View/>`s to handle resizing on the borders. But that seems more complex.

bryphe: (The drag zone itself, though, works great!)

zbaylin: Sounds good! I fixed this in the latest commit:
- Hit-tests are only enabled if `decorated` is false
- The examples now have a `--no-decoration` CLI arg so you can test the hit-tests within the examples
  - if the Window decorations are off, the example shows a message about restarting with `--no-decoration`

examples/Examples.re
```diff
@@ -271,13 +271,15 @@ let init = app => {
   |> (ignore: Revery.App.unsubscribe => unit);
 
   let initialExample = ref("Animation");
+  let decorated = ref(true);
   Arg.parse(
     [
       ("--trace", Unit(() => Timber.App.setLevel(Timber.Level.trace)), ""),
+      ("--no-decoration", Unit(() => decorated := false), ""),
```
bryphe, line 278: Great idea!

examples/HitTests.re
```diff
@@ -30,15 +30,22 @@ module Styles = {
...
-      <Text
-        text="Drag window from here!"
-        style=Styles.text
-        mouseBehavior=Draggable
-      />
-    </View>
+    {
+      Window.isDecorated(window) ?
+        <Text
+          text="Window decorations must not be enabled.\nRestart with --no-decoration"
```
bryphe: I was wondering how we'd update the example - thanks for adding this 👍

src/Core/WindowCreateOptions.re
```diff
@@ -14,6 +14,9 @@ type t = {
   maximized: bool,
   /**
     If [decorated] is true, the window will be framed with the OS-specific adorners.
+    If false, the default chrome behavior on the OS will be disabled and (if you want
+    to maintain that funtionality) you must implement it yourself through the
+    mouseBehavior attribute on Views
```
bryphe, line 19: Thanks for adding the additional details 👍

src/Core/WindowCreateOptions.re
```diff
@@ -14,6 +14,9 @@ type t = {
   maximized: bool,
   /**
     If [decorated] is true, the window will be framed with the OS-specific adorners.
+    If false, the default chrome behavior on the OS will be disabled and (if you want
+    to maintain that funtionality) you must implement it yourself through the
```
bryphe, line 18: Minor typo; `funtionality` -> `functionality`

zbaylin: Oops! Will fix that now!

PR #856, opened 2020-05-23
Stop separately tracking focus for Input

SirFurness:
Instead of tracking `isFocused` in `Input`, it checks with `Focus` if it is focused.

### DIFF
--- a/src/UI/Focus.re
+++ b/src/UI/Focus.re
@@ -29,15 +29,15 @@ let focus = (node: Node.node) => {
   focused := Some({handler: node#handleEvent, id: node#getInternalId()});
 };
 
+let isFocused = (node: Node.node) =>
+  switch (focused^) {
+  | Some({id, _}) => node#getInternalId() === id
+  | None => false
+  };
+
 /* TODO perform checks if a node can be focused ? */
 let dispatch = (node: Node.node) =>
-  switch (focused^) {
-  | Some({id, _}) =>
-    if (node#getInternalId() === id) {
-      ();
-    } else {
-      loseFocus();
-      focus(node);
-    }
-  | None => focus(node)
+  if (!isFocused(node)) {
+    loseFocus();
+    focus(node);
   };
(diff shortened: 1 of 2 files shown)


src/UI/Focus.re
```diff
@@ -29,15 +29,15 @@ let focus = (node: Node.node) => {
   focused := Some({handler: node#handleEvent, id: node#getInternalId()});
 };
 
+let isFocused = (node: Node.node) =>
+  switch (focused^) {
+  | Some({id, _}) => node#getInternalId() === id
```
glennsl, line 34: Not really related to this change, but this is a really weird use of physical equality. It shouldn't really matter for `int`s, but it is generally more error-prone and should almost never be used.

src/UI/Focus.re
```diff
@@ -29,15 +29,15 @@ let focus = (node: Node.node) => {
...
-      ();
-    } else {
-      loseFocus();
-      focus(node);
-    }
-  | None => focus(node)
+  if (!isFocused(node)) {
+    loseFocus();
+    focus(node);
   };
```
glennsl, line 43: Hmm, I don't understand why this exists. Or rather, I don't understand why `focus` exists and is used in place of this, as I can't imagine when you'd want `focus` to not trigger `blur`, and especially why you';d want that to be the default behaviour. This is probably causing bugs already.

src/UI_Components/Input.re
```diff
@@ -350,6 +350,13 @@ let%component make =
     };
   };
 
+  let handleRef = node => {
+    clickableRef := Some(node);
+    if (autofocus) {
+      Focus.focus(node);
```
glennsl, line 356: Why, for example, does this cause `focus` instead of `dispatch`? Is it not a bug that this does not trigger a `Blur` if some other node has focus?

I realize you didn't make these decisions, and most likely don't have an answer to any of these questions @SirFurness. But if no-one else does either, maybe we should do something about it to remove this source of bugs?

Et7f3:
The autopush haven't push because you have unused variable.

PR #858, opened 2020-05-24
fix(events): fix mouse positions not scaling correctly

zbaylin:
This totally flew under the radar because I'm on a Mac and (I assume) @bryphe and I have our scaling set to 1.0 on Windows, but I didn't scale the mouse coordinates properly on the HitTest params and FileDrop events.

@CrossR noticed this in onivim/oni2#1801, and I think this fixes that.

@CrossR, if you could resolve Revery in your local Oni `package.json` to the latest commit here and see if that fixes the issue, that would be super helpful!

Sorry this was overlooked in the first place! Luckily it doesnt seem to be _too_ hard a fix!

### DIFF
--- a/src/UI/HitTest.re
+++ b/src/UI/HitTest.re
@@ -8,8 +8,13 @@ let windowCallback = (node, window) =>
     (sdlWindow, mouseX, mouseY) =>
       Sdl2.Window.(
         if (sdlWindow == Window.getSdlWindow(window) && node#hasRendered()) {
+          let scaleAndZoomFactor = Window.getScaleAndZoom(window);
           let deepestNode =
-            getTopMostNode(node, float(mouseX), float(mouseY));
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks @CrossR for catching this, and @zbaylin - thanks for fixing it so quickly! Glad it was a straightforward fix 🎉 

> @bryphe and I have our scaling set to 1.0 on Windows, but I didn't scale the mouse coordinates properly on the HitTest params and FileDrop events.

Indeed, I should've thought to test this too!

> Its a shame there isn't any easy way to test any of this scaling related issues, but something like this especially really can't be automatically tested in an easy way.

Totally. The scaling issues have been tough - they happen to be platform specific (different behavior on Windows, Mac, and Linux). I'd love to have a way to have some automated tests to cover it.

PR #859, opened 2020-05-24
Stop separately tracking focus for Input

SirFurness:
I renamed `dispatch` to `focus` and `focus` to `focusWithoutBlur`.

(I accidentally screwed up the other pull request.)

I ran `esy format` and it didn't change anything, but hopefully it's happy now. I also am not sure what the unused variable mentioned is.

### DIFF
--- a/src/UI/Focus.re
+++ b/src/UI/Focus.re
@@ -29,15 +29,15 @@ let focus = (node: Node.node) => {
   focused := Some({handler: node#handleEvent, id: node#getInternalId()});
 };
 
+let isFocused = (node: Node.node) =>
+  switch (focused^) {
+  | Some({id, _}) => node#getInternalId() === id
+  | None => false
+  };
+
 /* TODO perform checks if a node can be focused ? */
 let dispatch = (node: Node.node) =>
-  switch (focused^) {
-  | Some({id, _}) =>
-    if (node#getInternalId() === id) {
-      ();
-    } else {
-      loseFocus();
-      focus(node);
-    }
-  | None => focus(node)
+  if (!isFocused(node)) {
+    loseFocus();
+    focus(node);
   };
--- a/src/UI/Focus.re
+++ b/src/UI/Focus.re
@@ -23,7 +23,7 @@ let loseFocus = () => {
   // If there is an active window, with text input active, turn off text input
 };
 
-let focus = (node: Node.node) => {
+let focusWithoutBlur = (node: Node.node) => {
   Log.trace("focus()");
   node#handleEvent(Focus);
   focused := Some({handler: node#handleEvent, id: node#getInternalId()});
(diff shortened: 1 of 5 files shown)


src/UI_Components/Input.re
```diff
@@ -108,9 +105,7 @@ let addCharacter = (word, char, index) => {
 
 let reducer = (action, state) =>
   switch (action) {
-  | Focus => {...state, isFocused: true}
-  | Blur => {...state, isFocused: false}
-  | TextInput(value, cursorPosition) => {...state, value, cursorPosition}
+  | TextInput(value, cursorPosition) => {value, cursorPosition}
   };
```
Et7f3, line 109: This reducer can now be rewritten with a state hook because one case switch ... is not very useful.

glennsl: They're not actually equivalent, because there's a bug with stale hook setters that don't affect reducers. I'm not sure whether this actually triggers here, but it might, if a key is held down and the key repeat interval is sufficiently short for example.

https://github.com/briskml/brisk-reconciler/issues/74

src/UI/Focus.re
```diff
@@ -23,21 +23,21 @@ let loseFocus = () => {
   // If there is an active window, with text input active, turn off text input
 };
 
-let focus = (node: Node.node) => {
+let focusWithoutBlur = (node: Node.node) => {
   Log.trace("focus()");
   node#handleEvent(Focus);
   focused := Some({handler: node#handleEvent, id: node#getInternalId()});
 };
```
glennsl, line 30: Thanks for fixing this. I'm not sure this should exist as a separate function though. It doesn't really make any sense to me, nor does the name really.

Also, it would be nice to have an interface for the whole `Focus` module that hides all of its internals, as anyone anywhere could just reassign `focused` directly.

src/UI/Focus.rei
```diff
@@ -0,0 +1,15 @@
+open NodeEvents;
+
+type active = {
+  handler: event => unit,
+  id: int,
+};
+type focused = ref(option(active));
+
+let getFocused: unit => option(active);
```
glennsl, line 9: Does this really need to be exposed? I think we can get away with just having a `dispatch` function that calls the current handler for us, if set.

SirFurness: Good point, that definitely makes things nicer.

src/UI/Keyboard.re
```diff
@@ -2,39 +2,33 @@ open Revery_Core;
...
+      KeyDown({
+        keycode: e.keycode,
+        scancode: e.scancode,
+        repeat: e.repeat,
+        keymod: e.keymod,
+        ctrlKey: Key.Keymod.isControlDown(e.keymod),
+        altKey: Key.Keymod.isAltDown(e.keymod),
+        shiftKey: Key.Keymod.isShiftDown(e.keymod),
+      }),
+    )
```
glennsl, line 32: Minor nit, and technically unrelated to this PR, but this really performs two separate actions. It first maps the internal event to an external event, and then dispatches it using `Focus.dispatch`. So the code for it should really be separated as well. The switch ought to just translate the events, and the call to `Focus.dispatch` ought to be factored out. I think `Mouse` has a separate top-level function to do the event translation.

src/UI/Focus.rei
glennsl: Indeed. Much better! Thanks!

PR #860, opened 2020-05-24
Use Hooks.mouseCapture

SirFurness:
Use `Hooks.mouseCapture` in `Clickable` instead of using its own system.

~~It also adds `onMouseEnter` and `onMouseLeave` to mouse capturing so that it can track when the mouse is over the button or not.~~

It releases the mouse on mouse up instead of mouse leave.
This allows mouse down, moving the mouse off, bringing the mouse back, and mouse up to register as a click.
Keeping the mouse captured until mouse up will also help with highlighting text in an input.

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -90,7 +90,6 @@ let%component make =
   <View
     style
     onMouseDown
-    onMouseUp
     onMouseEnter
     onMouseLeave
     ?onBlur
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -37,6 +37,8 @@ let%component make =
                 ~children,
                 (),
               ) => {
+  let%hook (isHovered, startHover, endHover) = Hooks.hover();
+
   let%hook mouseDownTimes = Hooks.ref(Constants.initialMouseDownTimes);
 
   let isDoubleClick = () =>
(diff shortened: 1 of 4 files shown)


src/UI_Components/Clickable.re
```diff
@@ -47,51 +43,47 @@ let%component make =
...
+    None;
+  };
+
+  let%hook (captureMouse, captureState) =
+    Hooks.mouseCapture(
+      ~onMouseUp,
+      ~onMouseEnter=(_state, _evt) => Some(true),
+      ~onMouseLeave=(_state, _evt) => Some(false),
+      (),
+    );
```
glennsl, line 75: This will never release the capture (unless the window as a whole loses focus). You release the capture by retuning `None`. My bad for not documenting that though. I'll add a doc comment to the hook explaining this.

src/UI_Components/Clickable.re
```diff
@@ -47,51 +43,47 @@ let%component make =
...
+
+    onAnyClick(mouseEvt);
+  };
+
+  let onMouseUp = (isInButton, evt: NodeEvents.mouseButtonEventParams) => {
+    if (isInButton) {
+      onMouseClick(evt);
     };
+    None;
+  };
```
glennsl, line 67: I found it a bit confusing that this was defined outside the hook. Could you assign the function to `~onMouseUp` directly instead?

src/UI_Components/Clickable.re
```diff
@@ -47,51 +43,47 @@ let%component make =
...
+
+  let%hook (captureMouse, captureState) =
+    Hooks.mouseCapture(
+      ~onMouseUp,
+      ~onMouseEnter=(_state, _evt) => Some(true),
+      ~onMouseLeave=(_state, _evt) => Some(false),
```
glennsl, line 73: This doesn't do what you expect it to do. `MouseEnter` and `MouseLeave` are triggered on the window, not on the node. And there's no reason to use capture for enter and leave events, as they are always triggered on the node itself. That's why they weren't included in the API.

src/UI_Components/Clickable.re
```diff
@@ -47,51 +43,47 @@ let%component make =
...
+    | MouseButton.BUTTON_LEFT =>
+      if (isDoubleClick()) {
+        resetMouseDownTimes();
+        onDoubleClick();
+      } else {
+        onClick();
+      }
+    | MouseButton.BUTTON_RIGHT => onRightClick()
+    | _ => ()
     };
```
glennsl, line 57: I'm not sure this needs to be a separate function. It's also a bit ambiguous to have both `onClick` and `onMouseClick`, where one triggers the other. Given a less ambiguous name it would be fine though.

SirFurness:
It seems the issue is knowing whether or not the mouse is over the button when mouse up occurs.

I saw that you mentioned a `Hooks.hover` earlier and I think that would clear this up.

I am curious what you think a good way of implementing this would be given that some components don't have access to their mouse enter and leave events (e.g. anything using a `Clickable`). I was looking at `UiEvents.isNodeImpacted` and `Mouse.storedNodesUnderCursor`, but neither feels like it was intended to be used this way.

glennsl:
Sorry for the late response. I'm not sure I understand what you mean though. `Clickable` uses `View`, which has `onMouseEnter` and `onMouseLeave`.

Also, `Hooks.hover` would just be for convenience, to remove a little bit of boilerplate. It wouldn't enable anything that's not already possible.

SirFurness:
Got it, thank you. I misunderstood how it worked.

SirFurness:
Unfortunately, that by itself didn't work because once the capture took the mouse move event the mouse enter/leave events would not be calculated.

I changed it so `bubble` events are blocked by mouse capturing but non-`bubble` events are not blocked by capturing.

I also added `setCallbacks` to mouse capture because otherwise if a callback depended on a hook, it would only use the value of that hook when the mouse was captured. (In this case, that was `isHovered` staying true because it was always true when the mouse was captured).

SirFurness:
This feels like a bad way to do things. I'll keep looking at it.

SirFurness:
Things might be best left as they are now. `Input` can always capture the mouse itself.

PR #862, opened 2020-05-26
fix(Windows): ensure WS_CAPTION is set on undecorated windows

zbaylin:
People in the Oni Discord were noting that an undecorated Window was covering the taskbar on Windows. Luckily, I found that Chrome had that same problem: https://chromium.googlesource.com/chromium/src.git/+/46.0.2478.0/chrome/browser/ui/views/apps/chrome_native_app_window_views_win.cc#71

This fixes that by adding an `Revery_Native.initWindow` function that sets the flag after all the SDL modifications are done. 

This also renames `Revery_Native`'s `init` to `initApp` to avoid confusion.

### DIFF
--- a/src/Native/Revery_Native.c
+++ b/src/Native/Revery_Native.c
@@ -11,6 +11,7 @@
 #ifdef WIN32
 #include "ReveryWin32.h"
 #include <combaseapi.h>
+#include <windows.h>
 #elif __APPLE__
 #include "ReveryCocoa.h"
 #import "ReveryAppDelegate.h"
@@ -18,7 +19,7 @@
 #include "ReveryGtk.h"
 #endif
 
-CAMLprim value revery_initialize() {
+CAMLprim value revery_initializeApp() {
 #ifdef __APPLE__
     SDLAppDelegate *sdlDelegate = [NSApp delegate];
     ReveryAppDelegate *delegate = [ReveryAppDelegate newWithSDLDelegate:sdlDelegate];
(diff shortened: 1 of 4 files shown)


bryphe:
> Luckily, I found that Chrome had that same problem: https://chromium.googlesource.com/chromium/src.git/+/46.0.2478.0/chrome/browser/ui/views/apps/chrome_native_app_window_views_win.cc#71

Awesome find, Zach!

src/Native/Revery_Native.c
```diff
@@ -32,9 +33,23 @@ CAMLprim value revery_initialize() {
...
 }
+
+
+CAMLprim value revery_initializeWindow(value vWin) {
+    CAMLparam1(vWin);
+    void *win = (void *)vWin;
+#ifdef WIN32
+    HWND window = (HWND)win;
+    int current_style = GetWindowLong(window, GWL_STYLE);
+    SetWindowLong(window, GWL_STYLE, current_style | WS_CAPTION);
```
bryphe, line 50: Might be worth adding a link to that chrome source URL you found: https://chromium.googlesource.com/chromium/src.git/+/46.0.2478.0/chrome/browser/ui/views/apps/chrome_native_app_window_views_win.cc#71

zbaylin:
@CrossR I fixed that issue you mentioned in onivim/oni2#1823 about the window being maximized on start. Check this out locally and see if this fixes it!

CrossR:
Tested with your Oni2 PR again and it looks like its working now! Thanks again for the fix!

vhanla:
I don't know if minimize from taskbar is not allowed by default on undecorated windows, but I suppose that can be modified by end user(developer) on his own if he/she wishes to do.



As shown in the GIF above, if you also add SysMenu and MinimizeBox styles \`SetWindowLong(window, GWL\_STYLE, current\_style | WS\_CAPTION | WS\_MINIMIZEBOX | WS\_SYSMENU);\` it will allow to minimize Examples' window by clicking on its icon in Windows taskbar, and also by right clicking it while holding shift key, it will show the system contextual menu, which currently is not possible in Oni2, unfortunately that won't fix the lack of `A`lt + `S`pace functionality (to resize/move manually using only keyboard).

PR #863, opened 2020-05-29
feat(hotreload): add hotreloadable components

zbaylin:
⚠️ This is total WIP at the moment ⚠️

Some of the things that need to be done:

- [ ] Get bytecode builds working again
- [ ] Move to a `libuv`/`luv`-based event loop
- [ ] Use `luv` for filewatching
- [ ] Create a seamless hotreloadable-component API
  - What does this mean?
  - Should we use functors?
- ...
- Profit!

### DIFF
--- a/examples/hotreload/HotreloadExample.re
+++ b/examples/hotreload/HotreloadExample.re
@@ -1,0 +1,30 @@
+open Revery;
+open Revery.UI;
+open Revery.UI.Components;
+
+module Example = {
+  module Styles = {
+    open Style;
+    let text = [
+      color(Colors.white),
+      fontFamily("Roboto-Regular.ttf"),
+      fontSize(14.),
+    ];
+    let outer = [
+      position(`Absolute),
+      top(0),
+      bottom(0),
+      left(0),
+      right(0),
+      justifyContent(`Center),
+      alignItems(`Center),
+    ];
+  };
+
+  let render = () =>
+    <View style=Styles.outer>
+      <Text style=Styles.text text="This component was dynamically loaded!" />
+    </View>;
+};
+
+let () = HotReload.setModule("HotreloadExample", (module Example));
--- a/examples/hotreload/HotreloadExample.re
+++ b/examples/hotreload/HotreloadExample.re
@@ -5,6 +5,12 @@ module Example = {
   module Styles = {
     open Style;
     let text = [
+      color(Colors.blueViolet),
+      fontFamily("Roboto-Regular.ttf"),
+      fontSize(14.),
+    ];
+
+    let text2 = [
       color(Colors.white),
       fontFamily("Roboto-Regular.ttf"),
       fontSize(14.),
@@ -21,13 +27,17 @@ module Example = {
   };
 
   let text = "ABC";
-  let render = () =>
-    <View style=Styles.outer>
+  let%component hot = () => {
+    let%hook (style, setStyle) = Hooks.state(true);
+
+    <View style=Styles.outer onMouseUp={_ => setStyle(n => !n)}>
       <Text
-        style=Styles.text
-        text="This component was ynamically loaded! Cool!"
+        style={style ? Styles.text : Styles.text2}
+        text="This component was loaded! Wow! Check this out!"
       />
     </View>;
+  };
+  let render = () => <hot />;
 };
 
 let () = HotReload.setModule("HotreloadExample", (module Example));
--- a/examples/hotreload/HotreloadExample.re
+++ b/examples/hotreload/HotreloadExample.re
@@ -1,6 +1,5 @@
 open Revery;
 open Revery.UI;
-open Revery.UI.Components;
 
 module Example = {
   module Styles = {
@@ -21,9 +20,13 @@ module Example = {
     ];
   };
 
+  let text = "ABC";
   let render = () =>
     <View style=Styles.outer>
-      <Text style=Styles.text text="This component was dynamically loaded!" />
+      <Text
+        style=Styles.text
+        text="This component was ynamically loaded! Cool!"
+      />
     </View>;
 };
 
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "build:release": "esy b dune build -p Revery -j4",
     "test": "esy b dune runtest",
     "format": "esy #{os == 'windows' ? 'b' : ''} bash .ci/format.sh #{os}",
-    "run": "esy x Examples"
+    "run": "esy x Examples",
+    "run-bytecode": "esy x Examples.bc"
   },
   "homepage": "https://github.com/bryphe/revery#readme",
   "esy": {
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@opam/lwt": "^4.0.0",
     "@opam/lwt_ppx": "^1.1.0",
     "@opam/ppx_deriving": "*",
+    "@opam/luv": "~0.5.1",
     "@brisk/brisk-reconciler": "briskml/brisk-reconciler#10cab2d",
     "flex": "1.2.3",
     "reperf": "^1.4.0",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "build": "esy b",
+    "build-watch": "esy b dune build --watch",
     "build:release": "esy b dune build -p Revery -j4",
     "test": "esy b dune runtest",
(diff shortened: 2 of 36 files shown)


src/Core/App.re
```diff
@@ -298,5 +302,14 @@ let start = init => {
...
+  let idle = Luv.Idle.init() |> Result.get_ok;
+
+  ignore @@
+  Luv.Idle.start(idle, () =>
+    if (appLoop()) {
+      Luv.Idle.stop(idle) |> ignore;
+    }
+  );
+
+  Luv.Loop.run() |> (ignore: bool => unit);
```
bryphe, line 314: It might be worth moving the `Luv` change to  a separate PR (and testing individually) - there's a few things that I want to make sure - namely, there is no perf impact and that that some of the perf saving we do (like switching to 'idle' mode using `SDL_WaitTimeout`) still work as expected. 

It seems like we'd also need a timer to ping every so often, as I believe the `idle` callback only gets called once per event loop - if there is nothing happening in the event loop, it might not get called.

package.json
```diff
@@ -39,6 +41,7 @@
     "@opam/lwt": "^4.0.0",
     "@opam/lwt_ppx": "^1.1.0",
     "@opam/ppx_deriving": "*",
+    "@opam/luv": "~0.5.1",
```
bryphe, line 44: We'll actually need to pick up a newer version (for Onivim)
```
 "@opam/luv": "aantron/luv:luv.opam#dfab9df",
```

src/Core/App.re
```diff
@@ -298,5 +302,14 @@ let start = init => {
     false;
   };
 
-  Sdl2.renderLoop(appLoop);
+  let idle = Luv.Idle.init() |> Result.get_ok;
+
+  ignore @@
```
bryphe, line 307: This style of `ignore` can cause issues if the signature changes - more details here: https://onivim.github.io/docs/for-developers/style-guide#ignoring-values

src/Core/App.re
```diff
@@ -255,6 +255,10 @@ let start = init => {
 
   Revery_Native.initApp();
 
+  Log.infof(m =>
+    m("Running in %s mode", Dynlink.is_native ? "native" : "bytecode")
+  );
```
bryphe, line 260: Good idea to log this!

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
...
+        }
+      | [@implicit_arity] Ok(path, events) => {
+          let filename = Filename.basename(path);
+          if (List.mem(`CHANGE, events) && Hashtbl.mem(table, filename)) {
+            Log.infof(m => m("File %s was changed.", path));
+            let command =
+              switch (Sys.getenv_opt("REVERY_HOTRELOAD_COMMAND")) {
+              | Some(c) => c
+              | None => "dune build"
+              };
```
bryphe, line 59: We probably don't want to get involved with having to manage the build here - it'd be simpler to let another mechanism handle that (ie, `dune watch`)

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
+open Revery_UI;
```
bryphe, line 1: Would be great to have a `rei` file to see a birds-eye view of the API (and what is important for the consumer, and what are internal details)

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
...
+
+module Log = (val Revery_Core.Log.withNamespace("Revery.Hotreload"));
+
+type obscure;
```
bryphe, line 5: What's this used for?

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
...
+
+let registerFile = filename => {
+  let path = filename |> getPath |> Dynlink.adapt_filename;
+  Dynlink.loadfile_private(path);
+};
+
+let start = () =>
+  if (!Dynlink.is_native) {
+    let dir = Revery_Core.Environment.getWorkingDirectory();
+    let watcher = Luv.FS_event.init() |> Result.get_ok;
```
bryphe, line 37: I wonder if we could move this into 'registerFile' - it'd be better to defer running this until the user asks to hot-reload. If we run `start()` at the toplevel, and it crashes, we won't get much error information.

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
+open Revery_UI;
+
+module Log = (val Revery_Core.Log.withNamespace("Revery.Hotreload"));
+
+type obscure;
+
+module type hotreload = {
+  let text: string;
+  let render: unit => React.element(React.reveryNode);
+};
```
bryphe, line 10: I wonder if we could make this more general?

src/HotReload/Revery_HotReload.re
```diff
@@ -0,0 +1,73 @@
...
+      dir,
+      fun
+      | Error(e) => {
+          Log.warnf(m =>
+            m("Could not watch %s: %s", dir, Luv.Error.strerror(e))
+          );
+          ignore(Luv.FS_event.stop(watcher));
+          Luv.Handle.close(watcher, ignore);
+        }
+      | [@implicit_arity] Ok(path, events) => {
```
bryphe, line 51: The `[@implicit_arity]` isn't really need here

examples/hotreload/HotreloadExample.re
```diff
@@ -0,0 +1,43 @@
...
+      <Text
+        style={style ? Styles.text : Styles.text2}
+        text="This component was loaded! Check this out!"
+      />
+    </View>;
+  };
+  let render = () => <hot />;
+};
+
+let () = HotReload.setFile("HotreloadExample.re", (module Example));
```
bryphe, line 43: It's a bit confusing to me that Hotreload is working against a source file.

The API I think that would be most intuitive for the hotreload layer would be something like:
```
HotReload.watch("HotreloadExample.cma", ((module Example)) => {
   // Do something with new example, or maybe not
});
```

bryphe: `setFile` is a bit confusing for me, it's not clear what we're 'setting' in this case.

zbaylin: I totally agree, I had it this way because I originally had the source file watching, but got rid of it in commits I havent pushed yet switching to the bytecode watching

bryphe: This means that instead of forcing the UI to 'pull' for new changes - we can explicitly trigger a re-render when there is a new version of the component available.

bryphe, line 43: It'd be really cool if we could push this up even higher in the Example app - like here:
https://github.com/revery-ui/revery/blob/e5f363a1ac1d2b3a91f30ce936efa18478fedb1a/examples/Examples.re#L342

such that everything could be reloaded.

IE:

```
let update = UI.start(window, <ExampleHost window initialExample />);

let _: unit => unit = HotReload.watch("Example.cma", ((module ExampleHost)) => {
   update(<ExampleHost window initialExample />)
});
```

bryphe:
So exciting, this could be the start of a really great developer experience for Revery. Thanks for your investigation on this, @zbaylin !

src/Core/App.re
tcoopman: Just a side remark. I was wondering why not just do `let () = f x` instead of explicitly annotating the result?

src/HotReload/Revery_HotReload.re
zbaylin: Oops, that's my bad. Didn't mean to leave that in there!

examples/hotreload/HotreloadExample.re
zbaylin: > It's a bit confusing to me that Hotreload is working against a source file.
> 
> The API I think that would be most intuitive for the hotreload layer would be something like:
> 
> ```
> HotReload.watch("HotreloadExample.cma", ((module Example)) => {
>    // Do something with new example, or maybe not
> });
> ```

When you say this, do you mean you want the callback to be in the file that's being reloaded or the "root" of the project that you're reloading from? The latter makes more sense in my head, but before I change it I want to make sure I understand you correctly.

zbaylin: Ah, this clears up my previous comment. Sounds good!

zbaylin:
Time to rebase this 😅

zbaylin:
That was easier than I thought 👍

PR #864, opened 2020-05-30
Ocaml/4.10

Et7f3:
This PR make Revery buildable with OCaml 4.10

Linked PR:
- [x] https://github.com/onivim/reason-native-crash-utils/pull/1
- [x] https://github.com/revery-ui/reason-font-manager/pull/9 (should to be released if we don't want to have a sha)
- [x] https://github.com/revery-ui/reason-skia/pull/73


Et7f3:
The failure aren't caused by the change in our code source but it is caused by upstream change. And the two PR are still needed to build with 4.10. And a ci check failed with a 503 so not related to real failure.

bryphe:
Thanks for upgrading this, @Et7f3 !

>  revery-ui/reason-font-manager#9 (should to be released if we don't want to have a sha)

I'd like to 'monorepo' this - bring it into this repo. Will make it easier to manage.

The only concerning issue is the build failure on Windows - I see this failure: 
```
  build log:
    # esy-build-package: building: @opam/ocaml-secondary-compiler@opam:4.08.1-1
```

I'm not sure why we need the 4.08.1 secondary compiler for this?

Probably related to this:
https://github.com/esy-ocaml/esy-opam-override/commit/1aca5f5e4dc473c5b04edd3f7a7cec4aaab4d8c7

Which might be fixed by:
https://github.com/esy-ocaml/esy-opam-override/commit/b17938187580b2fce35e138a53754b7092a21937

@prometheansacrifice might have ideas for us to work around this. In the meantime, for Onivim, I had to pin dune to `2.5.0`: https://github.com/onivim/oni2/blob/a46192af82cd5c9f317ba0b55a8e0344142cb3f4/package.json#L209 - we could do that here as well, until the issues with the override get sorted out.

Et7f3:
We can try with dune <=2.5.0. Dune 2.5.1 require second compiler to be installed with old ocaml. And it is transient I though @manas has removed this override

prometheansacrifice:
While the opam constraints only pull in the secondary compiler on 4.06.1, esy seems to pull it in for higher compilers as well. I'm not entirely sure why esy is doing that.

prometheansacrifice:
I've made some progress btw on the Windows build of the secondary compiler. We should be seeing them live in sometime

prometheansacrifice:
@Et7f3 Could you try running the CI once again? The overrides are on master now (not sure if it was during the last run)

prometheansacrifice:
Clarifying: the overrides are only for Dune 2.5.1 and ocaml-secondary-compiler.4.08.1-1

Et7f3:
I have waited https://github.com/esy-ocaml/esy-opam-override/pull/63 to be closed before relaunch

PR #869, opened 2020-06-01
fix(primitives): canvas matrix operations did not work as expected

glennsl:
Using `CanvasContext.setMatrix` in a canvas rendering function produced unexpected transformations. I believe this is because the root transform was not being set on the canvas context, and so any matrix operation would override the transform of the element itself. But because of all the mutation going on here I'm not sure whether doing this may have other side-effects.

### DIFF
--- a/src/UI/CanvasNode.re
+++ b/src/UI/CanvasNode.re
@@ -28,6 +28,7 @@ class canvasNode (()) = {
     | Some(r) =>
       let _ret: int = CanvasContext.save(canvas);
       CanvasContext.setMatrix(canvas, world);
+      CanvasContext.setRootTransform(world, canvas);
       Overflow.render(canvas, LayoutTypes.Hidden, dimensions, () => {
         // canvas save
         // canvas set transform
--- a/examples/CanvasExample.re
+++ b/examples/CanvasExample.re
@@ -24,6 +24,10 @@ module Sample = {
       <Canvas
         style=outerBox
         render={canvasContext => {
+          let transform = Skia.Matrix.make();
+          Skia.Matrix.setIdentity(transform);
+          CanvasContext.setMatrix(canvasContext, transform);
+
           let paint = Skia.Paint.make();
           Skia.Paint.setColor(
             paint,


src/UI/CanvasNode.re
```diff
@@ -28,6 +28,7 @@ class canvasNode (()) = {
     | Some(r) =>
       let _ret: int = CanvasContext.save(canvas);
       CanvasContext.setMatrix(canvas, world);
+      CanvasContext.setRootTransform(world, canvas);
```
bryphe, line 31: I don't think this is correct - this will break scaling. The root transform handles the scale factor, so if we reset it, it will break for scaled devices. We want a hierarchy of transforms instead.

Example app scaled 1.5, expected:

image

Example app scaled 1.5, PR:

image

Since our root transform is overridden, the scaling is not properly accounted for, so the canvas ends up being shifted (and the wrong size).

Do you have a repro I can try where things are working incorrectly?

glennsl: Of course, scaling! :man_facepalming: 

> Do you have a repro I can try where things are working incorrectly?

Yeah, just remove this line. I added an identity transform to the canvas example for testing, which shouldn't affect anything, but without setting the root transform here it does.

bryphe: Ah, got it.

This code: 
```
	   let transform = Skia.Matrix.make();
          Skia.Matrix.setIdentity(transform);
          CanvasContext.setMatrix(canvasContext, transform);
```
won't work as expected. `setMatrix` actually overrides the entire transform with a new matrix - so any parent transforms won't get applied. `setMatrix` isn't a great name, but it's what Skia uses, so we inherit.

For example, if we have a node hierarchy:

- Root (`scaleTransform`)
  - Node 1 (`transform1`)
     - Node 2 (`transform2`)

We use the transform: `transform2 * transform1 * scaleTransform` to position the canvas. If we use `setMatrix(identity)`, we'll lose all the parent transforms (aside from the root one). What we'd actually want is `setMatrix(identity * transform2 * transform1 * scaleTransform`).

There's an operation skia exposes `concat`, which appends the new transform matrix to the existing transform (`identity * existingTransform`), which I think is what we need here:
https://github.com/revery-ui/esy-skia/blob/c8bded8894f6624a0c1b0b1b2736630754bef62b/src/c/sk_canvas.cpp#L153 (and stubbed here: https://github.com/revery-ui/reason-skia/blob/5e94192b90b71122fa821e9c4e6b979b15c7df83/src/Skia.rei#L371)

bryphe: We're missing `concat` on `CanvasContext` but we should add it:
```
let concat = (~transform: Skia.Matrix.t, {canvas}: t) => {
  Skia.Canvas.concat(canvas, transform);
}
```

bryphe: And then if we use it in the example:
```
          let transform = Skia.Matrix.make();
          Skia.Matrix.setIdentity(transform);
          CanvasContext.concat(~transform, canvasContext);
```

Seems like it works OK!

glennsl: Ah, yeah `concat` should work (also not a great name really, but that seems par for the course with skia unfortunately). And then just use `save` and `restore` to get back to the previous transform I'm guessing?

Issue #867, opened 2020-05-31
0.31 is broken after breaking change in reason-sdl2


glennsl:
Because `reason-sdl2` is not pinned to a specific version and does not follow semver with regards to breaking changes because the version is set up to track the version of `sdl2`, Revery 0.31 is now broken.

Given the difficulty of making `reason-sdl2` follow semver properly, we might want to consider specifying a specific version at least for releases.

It would also be nice if another release could be published to have the latest release in working order.

Original report: https://github.com/revery-ui/revery-quick-start/issues/72

bryphe:
Now that we've "monorepo'd" SDL2 here - this sort of version mismatch should no longer be an issue-

PR #870, opened 2020-06-01
refactor(monorepo): Move in reason-skia

bryphe:
This brings in `reason-skia` to be 'monorepo'd` here - I think this will make it easier to iterate on Skia+Revery (for example, if we want to bring new primitives in - it's much easier to do it in a single PR and a codebase than to manage resolutions while developing).

This doesn't preclude `reason-skia` being released independently as a package, either - we could release it by itself like the [`reason-native`](https://github.com/facebookexperimental/reason-native) monorepo does.

This also streamlines maintenance - we don't have to port CI changes, or dependency changes, across multiple repos.

__Next steps:__
- Bring in `reason-harfbuzz` to the revery repo (@zbaylin - this should make it easier for us to augment harfbuzz for font-fallback next month)
- Bring in `reason-font-manager` to the revery repo (@manuhornung - hopefully this doesn't interfere with your work!)
- Bring in `reason-sdl2` to help prevent issues like #867 - since these dependencies don't respect `semver` anyway, baking them in can simplify our versioning story.


manuhornung:
No interference so far, and now that I know it, I can keep it in mind! I like the monorepo idea, should make things easier to follow and reduce overhead indeed!

bryphe:
Trying a more streamlined approach in #880

PR #874, opened 2020-06-01
feat(style): migrate font family API to use actual families

zbaylin:


examples/Border.re
```diff
@@ -30,7 +30,7 @@ let innerStyle =
 let textStyle =
   Style.[
     fontSize(20.),
-    fontFamily("Roboto-Regular.ttf"),
+    fontFamily(FontFamily.asset("Roboto-Regular.ttf")),
```
bryphe, line 33: Since we're breaking the API anyway - I wonder if we could move these out of `Style`, and just have them be properties on a `<Text />` component - like `<Text fontSize={20.0} color=Colors.black fontFamily =FontFamily.asset("Roboto-Regular.ttf") />`

bryphe: The `Style` API unfortunately comes with some disadvantages - like the monster type error you ran into!

I see moving away from the `Style` API to first-class properties on the component (at least for `<Text />`, `<Image />` for now)

src/UI/Node.re
```diff
@@ -137,7 +137,7 @@ class node (()) = {
       };
   };
   pub setStyle = style =>
-    if (style != _style) {
+    if (style !== _style) {
```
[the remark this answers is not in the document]
bryphe: My preference would actually be to move the `fontFamily` out of the style prop, which would side-step these issues! This change would make check the more expensive (`!=` is reference equality, whereas `!==` is structural equality).

src/UI/Style.re
```diff
@@ -1,8 +1,23 @@
...
 
 open Revery_Core;
-
-type fontFamily = string;
+open Revery_Font;
+
+type fontFamily = (Weight.t, bool, bool) => string;
+
+module FontFamily = {
+  let asset = (~variant: option(fontFamily)=?, default: string): fontFamily =>
```
bryphe, line 9: `asset` seems a bit ambiguous to me - it sounds like something bundled along with the app. I think `fromFile` or `fromPath` might be clearer.

examples/Border.re
zbaylin: I think that would be a pretty frictionless change. The only problem I could foresee is you have to technically have to write 3x as many attrs for each text element. Maybe something like a `fontProps` would be more appropriate? IIRC Flutter's component model works this way.

bryphe: I'm open to something like this 👍

bryphe: What's the API you have in mind?

zbaylin: I'm on my phone, but what I was thinking was a record API wrapped with a function. Something like:
`FontProps.re`
```reasonml
type t = {
  family: (Weight.t, bool, bool) => string,
  weight: Weight.t,
  italicized: bool,
  monospaced: bool
};

let make = (~weight=Weight.Normal, ~italicized=false, ~monospaced=false, ~fontFamily as family) => {
  family,
  weight,
  italicized,
  monospaced,
}

...
```

Then a Text element would take a `FontProps.t` instead.

Sorry for any formatting errors! Coding with softkeys is hard lol

src/UI/Style.re
bryphe: It also seems like we should have this in `Revery_Font` - this would let us have an abstract type `t` like `FontFamily.t`.

We could also have a method like `FontFamily.resolve(~fontFamily, ~weight, ~italicized, ~monospaced)` that returns a `Font.t` - it'd be what the `Text` component uses to find the actual font instance for rendering.

examples/Border.re
bryphe: Hmm, one thing I'd like to see is for us to abstract the font family - like a `FontFamily.t`.

src/UI/Style.re
bryphe: Also functions like:
```
let system: (~name: string) => FontFamily.t;
let fromFile: (~filePath: string, ~variants: (weight, ~italicized, ~monospace) => string)) => FontFamily.t;
...
```

examples/Border.re
zbaylin: What information do you think that type should contain? I was thinking about abstracting it out, but the function type seemed to naturally result from the API you suggested. Nothing we can't change, though!

src/UI/Style.re
zbaylin: > Also functions like:
> ```
> let system: (~name: string) => FontFamily.t;
> let fromFile: (~filePath: string, ~variants: (weight, ~italicized, ~monospace) => string)) => FontFamily.t;
> ...
> ```

Tangentially -- the `system` function I think should also take the monospaced, italicized, etc. options, because font discovery requires those params as well. I haven't pushed the system implementation yet, but it fit well into that function type as well

src/UI/Node.re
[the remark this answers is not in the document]
zbaylin: Works for me!

src/UI/Style.re
zbaylin: > Tangentially -- the `system` function I think should also take the monospaced, italicized, etc. options, because font discovery requires those params as well. I haven't pushed the system implementation yet, but it fit well into that function type as well

Scratch that, I was tired and I didnt really realize what I was saying 😳. My point I guess was that `FontFamily.t` should maybe be a function that takes the same arguments as the variants function you described.

examples/Border.re
zbaylin: Looking at this again, I think I see what you're saying. What if we made `FontFamily.t` that function type but memoized in some way? I think this is our best bet to maximize performance because we aren't doing the expensive task of font discovery for all variants up-front. The caching/memoization would just be an added optimization to make sure we only discover/resolve once.

bryphe: > What if we made FontFamily.t that function type but memoized in some way?

It's fine for `FontFamily.t` to still be a function (with the same implementation). The main issue is that the _consumer_ of a FontFamily shouldn't have to know about the internal details of whether it is a function or not. I suspect though we might need at least a hashtable or some additional context for that type - along with the function - to implement the caching.

> I think this is our best bet to maximize performance because we aren't doing the expensive task of font discovery for all variants up-front. The caching/memoization would just be an added optimization to make sure we only discover/resolve once.

Definitely! Thanks for thinking about performance.

src/UI/Style.re
bryphe: I think we're conflating two things. The first is the _definition_ of a `FontFamily` - ie, `FontFamily.system("Arial")`. This is just a declarative description of a font-family.

Then, there is the actual resolution of the font - where we take the `FontFamily.t`, along with some options (`weight`, `italicized`, `monospace`), and resolve it into a font instance for rendering. 

So the complete interface would look like:
```
module FontFamily = 
type t;

// Definition of font families
let system: (~name: string) => FontFamily.t;
let fromFile: (~filePath: string, ~variants: (weight, ~italicized, ~monospace) => string)) => FontFamily.t;

// Resolution to a concrete font instance (for rendering)
let resolve: (~weight, ~italicized, ~monospace, FontFamily.t) => Font.t
};
```

When we use a component, we'd do:
```
// We should re-use this family instance, so we can benefit from caching
let arialFamily = FontFamily.system("Arial");

<Text fontFamily=arialFamily weight=FontWeight.bold italic=true/>
```

And inside `TextNode`, we'd use `FontFamily.resolve`, based on the component parameters, to find a concrete instance of `Font.t` for rendering.

examples/Border.re
zbaylin: > It's fine for FontFamily.t to still be a function (with the same implementation). The main issue is that the consumer of a FontFamily shouldn't have to know about the internal details of whether it is a function or not.

Definitely -- I see what you mean now. 

> I suspect though we might need at least a hashtable or some additional context for that type - along with the function - to implement the caching.

In the latest commit you can see how I used an LRU cache for the system half of it. We can use a hashtable too, that would definitely simplify things a little bit.

src/UI/Style.re
zbaylin: I see what you mean now. The only detail left I'm a little unclear about is what `Font.t` is. Right now we use `FontCache.load` to get the font information from Harfbuzz and Skia. Would `resolve` just be a helper function that goes from `fontFamily(weight, italicized, monospaced) |> FontCache.load`?

zbaylin:
The nice thing about this API is that now we can have a default font family for each system! This fixes that bug that everyone seems to run across with empty font families.

src/UI/Style.re
bryphe: > Would resolve just be a helper function that goes from fontFamily(weight, italicized, monospaced) |> FontCache.load?

Sounds like we're on the same page! Yes, essentially - but there may be additional indirection to implement caching. But that would be the initial implementation. Hiding the implementation behind an abstract `t` gives us flexibility with what the actual implementation is - we're not coupled to it being to a function, at least from a consumer's point of view.

zbaylin:
I'm going to close this out because I'm reworking it so that the `fontProp`s are intrinsic to the `TextNode`s themselves rather than Style. There's been too much changed here to go that route.

PR #875, opened 2020-06-01
Add RichText component

### LINKED ISSUE #654: RichText model
It would be useful to have a model that associates text with text style properties, and can associatively combine into a sequence (a semigroup).

This doesn't need to be fancy to begin with, just the properties we have today (fontFamily, fontSize, color) would be useful, and we'd then have a nice base to extend with other properties.

Something like this:

```reason
module RichText : {
  type t;

  let create: (~fontFamily: string, ~fontSize: int, ~color: Color.t, string) => t;
  let (++): (t, t) => t;
  let measure: t => Dimensions.t;
}
```

The data structure could be implemented as a tree:

```reason
type t =
  | Leaf({ fontFamily: string, fontSize: int, color: Color.t, text: string })
  | Node(t, t);
```

And then it would be nice to have a `RichText` component which turned the model into a sequence of `Text` elements with the appropriate styles set.

SirFurness:
Adds a basic `RichText` component. Closes #654 

It also adds a simple example.

Having different font sizes causes problems with alignment. ``alignItems(`Baseline)`` would solve this but I don't think that currently exists here.


SirFurness:
The way I'm doing this seems a bit foolish as every tree concatenation eventually becomes a list concatenation in the `toList` function. How can I avoid doing this?

src/UI_Components/RichText.re
```diff
@@ -0,0 +1,74 @@
...
+      <Text
+        style=Style.[
+          fontFamily(text.fontFamily),
+          fontSize(text.fontSize),
+          color(text.color),
+        ]
+        text={text.text}
+      />,
+    ]
+  | Node(left, right) => toList(left) @ toList(right)
```
kyldvs, line 33: I think this will be slow, @ on lists is O(n) and it happens at each node, something like this is probably what you want:

```
let rec inOrder = (fn: ('acc, text) => 'acc, accumulation: 'acc, node: t): 'acc => {
  switch (node) {
  | Leaf(text) => fn(accumulation, text);
  | Node(left, right) =>
    let leftAcc = inOrder(fn, accumulation, left);
    let rightAcc = inOrder(fn, accumulation, right);
    rightAcc;
  };
};

let rec toList = (node: t) => {
  let itemsInReverse = inOrder((list, text) => [text, ...list], [], node);
  let itemsInCorrectOrder = List.rev(itemsInReverse);
  // convert them to <Text>
};
```

SirFurness: Definitely much better, thanks!

src/UI_Components/RichText.re
```diff
@@ -0,0 +1,78 @@
...
+    (~fontFamily: string, ~fontSize: float, ~color: Color.t, text: string) =>
+  Leaf({fontFamily, fontSize, color, text});
+
+let (++) = (left: t, right: t) => Node(left, right);
+
+let rec inOrder = (fn: ('acc, text) => 'acc, accumulator: 'acc, richtext: t) =>
+  switch (richtext) {
+  | Leaf(text) => fn(accumulator, text)
+  | Node(left, right) =>
+    let rightAcc = inOrder(fn, accumulator, right);
```
kyldvs, line 27: `inOrder` is a specific kind of tree-traversal: [wikipedia](https://en.wikipedia.org/wiki/Tree_traversal#In-order_(LNR))

By doing right, then left this is not longer an in-order traversal. I think it's more helpful to have an `inOrder` function and then deal with the list reversals for perf, but if you prefer to keep this version it should be renamed to something like `reverseOrder`.

SirFurness: Good point, I'll switch the names. We can always add an `inOrder` later if we need it.

src/UI/RichTextModel.re
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+let rec changeTextInfo = (updateLeaf: textInfo => t, richtext: t) =>
+  switch (richtext) {
+  | Leaf(textInfo) => updateLeaf(textInfo)
+  | Node(left, right) =>
+    let newLeft = changeTextInfo(updateLeaf, left);
+    let newRight = changeTextInfo(updateLeaf, right);
+    Node(newLeft, newRight);
+  };
```
glennsl, line 68: This is usually called `map` 😉 

You seem to be rediscovering all the standard functional data structure operations. And implementing them pretty much perfectly too, except for being overly specific in naming them. Good job!

src/UI/RichTextModel.re
```diff
@@ -0,0 +1,107 @@
...
+let fontWeight = (fontWeight: Weight.t, richtext: t) =>
+  richtext |> changeTextInfo(textInfo => Leaf({...textInfo, fontWeight}));
+let italicized = (italicized: bool, richtext: t) =>
+  richtext |> changeTextInfo(textInfo => Leaf({...textInfo, italicized}));
+let monospaced = (monospaced: bool, richtext: t) =>
+  richtext |> changeTextInfo(textInfo => Leaf({...textInfo, monospaced}));
+let fontSize = (fontSize: float, richtext: t) =>
+  richtext |> changeTextInfo(textInfo => Leaf({...textInfo, fontSize}));
+let smoothing = (smoothing: Smoothing.t, richtext: t) =>
+  richtext |> changeTextInfo(textInfo => Leaf({...textInfo, smoothing}));
```
glennsl, line 80: These would be a bit smarter and nicer to use if they weren't just dumb setters. Realistically you're not ever going to set italic to false, at least not using this DSL, so you could just have `italic` and `monospaced` just setting those properties to `true`. And instead of `fontWeight` and `smoothing` you could have separate functions for each value of the property variant, like `bold`, `bolder` etc.

Also, as a side-note, I like to implement DSL in a separate `DSL` module, and include that in the main module, just because they tend to have short function names that are very common and might be shadowed by local variable names and such. In that case it's nice to be able to do `DSL.color(...)` to access the DSL function.

src/UI_Components/RichText.re
```diff
@@ -0,0 +1,31 @@
+open Revery_UI;
+open Revery_UI_Primitives;
+
+let toList = (richtext: RichTextModel.t) =>
```
glennsl, line 4: `toList` is a bit vaguely named I think. It's also more of a `render` function really, that just happens to return the result as a list. I'm not really sure this needs to be a function at all. It could just be inlined.

SirFurness:
I tried moving `RichText` to `Core` but it creates a dependency cycle. 

`RichText` depends on `Font` which depends on `Core`.

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+
+let (++): (t, t) => t;
+
+let foldRight: (('acc, textInfo) => 'acc, 'acc, t) => 'acc;
+
+let map: (textInfo => t, t) => t;
+
+let measure: t => Dimensions.t;
+
+module type DSLType = {
```
glennsl, line 39: This is a bit too much DRY I think. There's nothing wrong with copy-pasting code when it increases clarity. Especially in interface files.

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+  let fontWeight: (Weight.t, t) => t;
+  let thin: t => t;
+  let ultralight: t => t;
+  let light: t => t;
+  let normal: t => t;
+  let medium: t => t;
+  let semibold: t => t;
+  let bold: t => t;
+  let ultrabold: t => t;
+  let heavy: t => t;
```
glennsl, line 49: Nice!

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+
+  let smoothing: (Smoothing.t, t) => t;
+  let noSmoothing: t => t;
+  let antialiased: t => t;
+  let subpixelAntialiased: t => t;
```
glennsl, line 54: These are very strange properties to have in a text model. I would consider it a property of the renderer instead. Is that possible to do?

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+
+  let fontFamily: (Family.t, t) => t;
+  let italicized: (~italicized: bool=?, t) => t;
+  let monospaced: (~monospaced: bool=?, t) => t;
```
glennsl, line 58: These should be separate functions instead of taking `bool`s.

(I consider `bool` arguments a bit of a code smell in general, although labelled arguments do often help a great deal)

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+    ~fontFamily: Family.t=?,
+    ~fontWeight: Weight.t=?,
+    ~italicized: bool=?,
+    ~monospaced: bool=?,
+    ~fontSize: float=?,
+    ~smoothing: Smoothing.t=?,
+    ~color: Color.t=?,
+    string
+  ) =>
+  t;
```
glennsl, line 29: I would consider this part of the DSL as well.

src/UI/RichText.rei
```diff
@@ -0,0 +1,63 @@
...
+type textInfo = {
+  fontFamily: Family.t,
+  fontWeight: Weight.t,
+  italicized: bool,
+  monospaced: bool,
+  fontSize: float,
+  text: string,
+  smoothing: Smoothing.t,
+  color: Color.t,
+};
```
glennsl, line 13: This could be inlined into the `Leaf` node I think. Is there a reason to keep it separate?

examples/RichTextExample.re
```diff
@@ -2,8 +2,6 @@ open Revery;
 open Revery.UI;
 open Revery.UI.Components;
 
-open Revery_Font;
-
```
glennsl: Did you remove this because you had an issue opening it here?

glennsl:
> I tried moving RichText to Core but it creates a dependency cycle.
>
> RichText depends on Font which depends on Core.

Ah, I see, that's unfortunate. Thanks for trying though!

src/UI/RichText.rei
SirFurness: It's used in `foldRight` and `map`.

examples/RichTextExample.re
SirFurness: It's back! I don't think there was reason.

src/UI/RichText.rei
SirFurness: I moved it to `RichTextView`. This now makes the assumption that all of the `RichText` will use the same `smoothing`, but I think that's a reasonable assumption.

src/UI/RichText.re
```diff
@@ -0,0 +1,129 @@
...
+  let italicized = (richtext: t) =>
+    richtext |> map(textInfo => Leaf({...textInfo, italicized: true}));
+  let monospaced = (richtext: t) =>
+    richtext |> map(textInfo => Leaf({...textInfo, monospaced: true}));
+  let fontSize = (fontSize: float, richtext: t) =>
+    richtext |> map(textInfo => Leaf({...textInfo, fontSize}));
+  let color = (color: Color.t, richtext: t) =>
+    richtext |> map(textInfo => Leaf({...textInfo, color}));
+};
+include DSL;
```
Et7f3, line 129: Why create a module to include it just after ?

examples/RichTextExample.re
```diff
@@ -0,0 +1,42 @@
...
+    top(0),
+    bottom(0),
+    left(0),
+    right(0),
+    alignItems(`Center),
+    justifyContent(`Center),
+    flexDirection(`Column),
+  ];
+
+let font = Family.system("Arial");
```
Et7f3, line 18: Thanks to @zbaylin in #879 we have a default one in https://github.com/revery-ui/revery/pull/879/files/25dfb38b8e57c35647ada63b3958b59d71e21fce#diff-ab5cacadbfac4f862c73d2dc57d40aff

src/UI/RichText.re
SirFurness: The idea is to allow, for example, `DSL.color` when the variable `color` is already being used.

examples/RichTextExample.re
SirFurness: I switched to Arial because of the bug that's fixed by #895. Otherwise, bold text was not working on Mac.

src/UI/RichText.re
zbaylin: Including just dumps all the values from one module into another. If you can access `DSL` to include it, you can access `DSL.color` without including it. Unless I'm misunderstanding you!

examples/RichTextExample.re
zbaylin: Thanks for reminding me to merge that @SirFurness! You can merge master now

src/UI/RichText.re
SirFurness: The goal was to keep `color` because of how short and easy it is to call while keeping `DSL.color` in case `color`, being such a short and common name, was already used.

For example,
```
let color = Colors.black;
RichText.(
  text("hello")
  |> DSL.color(color)
  |> fontSize(20.)
);
```

The `include` allows `fontSize` to be used without doing `DSL.fontSize` while having the `DSL` module allows `DSL.color` to be used since `color` already exists.

I can just leave it all in `DSL` if it's too weird.

zbaylin: I think it's better to keep it all in `DSL`. The user can choose to open `DSL` themselves or do
```
RichText.(
  DSL.(
    ...
  )
)
```

I dont think you should overwrite names like that, because it's not very consistent. If someone adds a conflicting name to DSL it would go unnoticed.

SirFurness: Sounds good!

Et7f3: Ah good args to prevent shading thanks.

glennsl: > I dont think you should overwrite names like that, because it's not very consistent. If someone adds a conflicting name to DSL it would go unnoticed.

It'll produce an unused variable warning.

The idea behind the `DSL` module is just to disambiguate when necessary when opening `RIchText`. Otherwise, if `DSL` is not included in `RichText`, opening `RichText.DSL` instead of `RichText` has the exact same problem.

glennsl: I would recommend reverting https://github.com/revery-ui/revery/pull/875/commits/17e54452473fbbf3e8e42ea60ba0f5ff270692a3

SirFurness: Reverted

glennsl:
:tada:

glennsl:
Thanks for doing an awesome job with this @SirFurness! :tada:

lessp:
ping @thangngoc89 😄

PR #876, opened 2020-06-01
chore(ci): Fix up triggers to reduce backlogs of builds in azure pipeline queue

bryphe:
This sets up triggers to auto-cancel PR builds when a new commit has been pushed

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,6 +4,21 @@
 # https://aka.ms/yaml
 
 name: $(Build.SourceVersion)
+
+# Master build triggers
+trigger: 
+- master
+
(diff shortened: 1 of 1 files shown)


Et7f3:
Can't you stop a build ? like we can do with GitHub Action ?

bryphe:
> Can't you stop a build ? like we can do with GitHub Action ?

We can manually, but what happens is we end up with backlog of builds - like, if multiple commits are pushed for a PR, each one triggers a build. This isn't ideal because of the time it takes windows to build, and the fact that we only have a fixed number of machines available. It's better if we just build the latest commits in a PR for azure pipelines:
image

PR #879, opened 2020-06-02
feat(components): move font styling out of Style and use families

zbaylin:
This creates a couple API changes:
- the introduction of a `FontFamily` with a `t` that can be made with `system` and `fromFile`
- the introduction of `FontProps` with family, weight, italicized, and monospaced information
- the migration to specifying these `FontProps` directly on the `Text` component/node


src/Font/FontFamily.rei
```diff
@@ -0,0 +1,12 @@
+open FontManager;
+
+type t = (FontWeight.t, bool, bool) => string;
```
bryphe, line 3: Can this be an abstract type, now? Just `type t;`? Or does some downstream code depend on this being a function?

src/Font/FontFamily.rei
```diff
@@ -0,0 +1,12 @@
...
+
+type variantSolver =
+  (~fontWeight: FontWeight.t, ~italicized: bool, ~monospaced: bool, unit) =>
+  string;
+
+let fromFile: (~variant: variantSolver=?, string) => t;
```
bryphe, line 9: IMO, I think inlining this type could be cleaner:
```
let fromFile: (
   ~variant=(~weight: FontWeight.t, ~italicized: bool, ~monospaced: bool) => string,
   string,
) => t;
```

Otherwise, glancing at the signature, it's not clear that `variantSolver` is only important for the `fromFile` signature, and not needed for others.

src/Font/FontFamily.re
```diff
@@ -0,0 +1,78 @@
...
+
+  let hash = Hashtbl.hash;
+};
+
+module FontDescriptorWeight = {
+  type t = FontManager.FontDescriptor.t;
+  let weight = _ => 1;
+};
+
+module FontFamilyCache = Lru.M.Make(FontFamilyHashable, FontDescriptorWeight);
```
bryphe, line 43: Good idea to use the LruCache, to help bound the number of familes we keep in memory!

src/Font/FontFamily.re
```diff
@@ -0,0 +1,78 @@
...
+  | Some(solver) => (
+      (w, i, m) =>
+        switch (w, i, m) {
+        | (FontWeight.Normal, false, false) => default
+        | _ => solver(~fontWeight=w, ~italicized=i, ~monospaced=m, ())
+        }
+    )
+  };
+
+module FontFamilyHashable = {
```
bryphe, line 21: Nice 👍  It's great that these implementation details aren't exposed to consumers of the `FontFamily` module - this is all tucked away behind the interface.

src/UI/TextNode.re
```diff
@@ -15,6 +15,10 @@ class textNode (text: string) = {
   val mutable _isMeasured = false;
   val mutable _lines: list(string) = [];
   val mutable _smoothing = Smoothing.default;
+  val mutable _fontFamily = Family.default;
+  val mutable _fontWeight = Weight.Normal;
+  val mutable _italicized = false;
+  val mutable _monospaced = false;
```
bryphe, line 21: Thanks for adding these!

src/Font/FontFamily.re
```diff
@@ -0,0 +1,78 @@
...
+  let weight = _ => 1;
+};
+
+module FontFamilyCache = Lru.M.Make(FontFamilyHashable, FontDescriptorWeight);
+
+let cache = FontFamilyCache.create(~initialSize=8, 64);
+
+let system = (familyName): t =>
+  (weight, italicized, monospaced) => {
+    let font: FontFamilyHashable.t = {
```
bryphe, line 49: Minor: `font` seems a bit unclear to me, maybe `fontDescription`?

src/UI/FontProps.re
```diff
@@ -0,0 +1,29 @@
...
+};
+
+let default = {
+  family: Family.default,
+  weight: Weight.Normal,
+  italicized: false,
+  monospaced: false,
+};
+
+let make =
```
bryphe, line 17: Thinking about this more, it's not really clear to me why this would be more convenient than just having the properties on the component:

ie:

`<Text fontProps=FontProps.make(~family="Arial", ~size=12, ~weight=FontWeight.Bold, ~italicized=true, ()) />`

vs

`<Text fontFamily="Arial" size=12 weight=FontWeight.Bold italicized=true/>`

I might be missing something, though. Do you have an example from flutter? Or a case where it'd be more convenient than just having them as properties on `<Text />`?

zbaylin: Thanks, from our DMs on Discord I realized this is better. This is changed in the latest commit!

src/Font/FontFamily.rei
zbaylin: That sounds good to me!

zbaylin: Fixed in the latest commit!

src/Font/FontFamily.rei
zbaylin: Fixed in the latest commits!

src/Font/FontFamily.re
zbaylin: Fixed in the latest commit! Thanks! I renamed it `fontDescr`

src/Font/FontFamily.re
```diff
@@ -0,0 +1,92 @@
...
+
+let fromFile =
+    (
+      ~variant:
+         option(
+           (
+             ~fontWeight: FontWeight.t,
+             ~italicized: bool,
+             ~monospaced: bool,
+             unit
```
bryphe, line 17: Since there aren't any optional parameters, we could remove the `unit`.

The `option` also shouldn't be necessary here, since we're using `=?`

zbaylin: Oh thanks for the reminder! Fixed in the latest commit!

src/UI/TextNode.re
```diff
@@ -131,10 +165,21 @@ class textNode (text: string) = {
       _this#markLayoutDirty();
     };
   pub setSmoothing = smoothing => _smoothing = smoothing;
-  pub setFontFamily = fontFamily => _fontFamily = fontFamily;
+  pub setFontFamily = fontFamily => {
+    if (_fontFamily !== fontFamily) {
+      _this#markLayoutDirty();
+    };
+    _fontFamily = fontFamily;
+  };
```
bryphe, line 173: Thanks for all the work to wire this up!

src/UI/TextNode.re
```diff
@@ -153,15 +198,27 @@ class textNode (text: string) = {
...
   pub handleTextWrapping = (width, style) => {
-    let {textWrap, fontFamily, fontSize, lineHeight, _}: Style.t = style;
+    let {textWrap, lineHeight, _}: Style.t = style;
     let lineHeightPx =
-      lineHeight *. Text.getLineHeight(~fontFamily, ~fontSize, ());
+      lineHeight
+      *. Text.getLineHeight(
+           ~fontFamily=
+             Family.toPath(
+               _fontFamily,
```
bryphe, line 207: One thing to think about is, maybe later, we could have the `Family` return a `Font.t` directly. Not a blocker for this change, though.

src/UI/TextNode.re
```diff
@@ -19,6 +19,7 @@ class textNode (text: string) = {
   val mutable _fontWeight = Weight.Normal;
   val mutable _italicized = false;
   val mutable _monospaced = false;
+  val mutable _fontSize = 0.;
```
bryphe, line 22: Is this a safe default? I haven't set font size to `0`, really, I'm not sure how Skia handles it 🤔

zbaylin: Oh, I set it to that knowing that it would never get that far, because `configureInstance` sets it up with a default of `14.`. I can change it here too though, in case we ever get some strange edge case!

src/Font/FontFamily.re
```diff
@@ -0,0 +1,91 @@
...
+    weight: FontWeight.t,
+    italicized: bool,
+    monospaced: bool,
+  };
+
+  let equal = (a, b) =>
+    String.equal(a.familyName, b.familyName)
+    && a.weight == b.weight
+    && a.italicized == b.italicized
+    && a.monospaced == b.monospaced;
```
glennsl, line 38: Just out of curiosity, why not just `a == b`? Does `String.equal` differ in behaviour from `==`?

PR #880, opened 2020-06-02
refactor(monorepo): Bring in reason-skia (round 2)

bryphe:


package.json
```diff
@@ -25,16 +23,18 @@
       }
     },
     "build": [
-      "dune build -p font-manager,harfbuzz,Revery -j4"
+      "dune build -p font-manager,harfbuzz,skia,Revery -j4"
     ],
     "install": [
       "esy-installer font-manager.install",
       "esy-installer harfbuzz.install",
+      "esy-installer skia.install",
```
Et7f3, line 31: Shouldn't we pick skia.dll now ?

bryphe: Ya, good point @Et7f3 , we should copy the `skia.dll` too

PR #881, opened 2020-06-03
Add windows, centos, archlinux in GHA CI

Et7f3:
Windows is added so we know we don't break major OS.
Centos is added to have parity with azure pipeline.
ArchLinux is added because it is a rolling release.

This PR also fix push docs from other non-master.
This PR also cache store from docker :) We can't see it because it isn't merged.

You can check it build here: https://github.com/revery-ui/revery/runs/732801762?check_suite_focus=true because without cache it is too much slow.

With this the build matrix get high. If we plan on adding build for IOS we might get another build: so we have to decide strategy ? Does it is fine that we wait a bit. Today we have 3 active PR so I have to kill some CI run. I haven't really this before so what do you think ?

### DIFF
--- a/scripts/docker/archlinux/Dockerfile
+++ b/scripts/docker/archlinux/Dockerfile
@@ -1,0 +1,14 @@
+FROM archlinux:latest
+
+RUN pacman -Syu --noconfirm
+
+RUN pacman -S --noconfirm base base-devel git npm
+
+RUN mkdir "${HOME}/.npm-packages"
+RUN npm config set prefix "${HOME}/.npm-packages"
+RUN echo export PATH=\"\$PATH:\${HOME}/.npm-packages/bin\" >> .bashrc
+RUN npm install -g esy
+
+RUN npm --version
+RUN git --version
+RUN esy --version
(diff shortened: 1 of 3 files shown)


scripts/docker/archlinux/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM archlinux:latest
```
bryphe, line 1: Nice! Thanks for setting this up @Et7f3

bryphe:
Thanks for setting this up, @Et7f3 !

It really is a lot faster with the cache. Very promising.

> This PR also fix push docs from other non-master.

Cool!

> This PR also cache store from docker :) We can't see it because it isn't merged.

Seems like it helps a lot with the CentOS Dockerfile. Is there anything special we need to do when we change the Dockerfile? Or is it smart to update when the Dockerfile changes?

Et7f3:
It only cache esy store ~/.esy/3_/i. The docker is rebuild all the time atm.

bryphe:
> f we plan on adding build for IOS we might get another build: so we have to decide strategy ? Does it is fine that we wait a bit. 

Ya, I think it's best to wait - we can manually kill jobs one-off if we need too.

Et7f3:
Merge is blocked: do you want I do something more before merging it ?

PR #883, opened 2020-06-03
refactor(monorepo): Move SDL2 in

bryphe:
This will help with issues like #867 , to avoid juggling version dependencies of libraries like `reason-sdl2` that don't respect semver.


Et7f3:
It doesn't seem to build sdl2 and we should know pick sdl2.dll like the skia one.

bryphe:
> It doesn't seem to build sdl2 and we should know pick sdl2.dll like the skia one.

Regarding build - I just included it in the `Revery` package (`Revery.sdl2`), I don't picture us exporting it as as separate project, as it's pretty specialized for Revery at this point (more likely - replace it with tsdl, eventually).

Thanks for catching the dll again! Copying it too on Windows now.

PR #886, opened 2020-06-03
feat(components): add a Markdown component

zbaylin:
This uses [omd](https://github.com/ocaml/omd)'s generic markdown parser to generate Markdown using Revery components.
image

To-do list:
- Blocks:
  - [ ] Paragraph
  - [ ] List
  - [ ] Blockquote
  - [ ] Thematic_break
  - [ ] Heading
  - [ ] Code block
  - [ ] HTML block (rendered as text)
  - [ ] Tag block
- Inlines:
  - [ ] Text
  - [ ] Emph
  - [ ] Code
  - [ ] Hard break
  - [ ] Soft break
  - [ ] Link
  - [ ] Ref
  - [ ] HTML (rendered as text)
  - [ ] Tag


src/UI_Components/Markdown.re
```diff
@@ -0,0 +1,226 @@
...
+      fontSize
+      fontFamily={styles.fontFamily}
+      fontWeight
+      italicized={isItalicized(attrs)}
+      monospaced={isMonospaced(attrs)}
+    />
+  };
+};
+
+let rec _generateInline = (inline, styles, attrs) => {
```
glennsl, line 116: You shouldn't use `_` to denote internal function as that has special meaning to the compiler. As with any other variable it disables dead code checking, which tends to make a mess over time, and it also tends to gray them out (depending on syntax theme of course), making them harder to find.

In this case I think you could just nest it inside `generateInline`. Alternatively you can use an `Internal` module or you can use `'` as a suffix, although I'd recommend being more specific.

bryphe:
So cool! Great use of the new `FontFamily` capabilities

src/UI_Components/Markdown.re
zbaylin: Thanks for the naming tips. I settled on using the `'` suffix, just because I think it makes the most intuitive sense ("function _prime_")

examples/MarkdownExample.re
```diff
@@ -0,0 +1,84 @@
...
+    alignItems(`FlexStart),
+  ];
+
+  let whiteText = [color(Colors.white)];
+  let linkActive = [color(Colors.blue)];
+  let linkInactive = [color(Colors.lightBlue)];
+};
+
+let markdown =
+  Markdown.make(
```
OhadRau, line 25: I'm not sure if there's a reason this would be better or worse in this scenario, but I feel like the typical way of doing something like this in OCaml is with functors:

```ocaml
module MD = Markdown.Make({
  include Markdown.DefaultStyle;
  let whiteText = [color(Colors.white)];
  let h1Style = whiteText;
});
```

I can't remember if Revery plays nice with module JSX though (IIRC it's the way all of ReasonReact works so syntactically at least it should be possible). I think apart from what's idiomatic OCaml, the main advantage is that you can reuse/inherit from styles (this can also be solved with records FWIW). But for example, you could have a standard style in your app and then in one place if you just want to change one thing you can include the original style and change the one property.

(Or this could just be me over-analyzing it since I guess normally these would just be passed in with a few props, but in that case I'd just rename the function to `markdown` so you can do `<markdown ~h1Style=...>`)

examples/MarkdownExample.re
```diff
@@ -0,0 +1,84 @@
...
+    ~h5Style=Styles.whiteText,
+    ~h6Style=Styles.whiteText,
+    ~activeLinkStyle=Styles.linkActive,
+    ~inactiveLinkStyle=Styles.linkInactive,
+  );
+
+let example = () =>
+  <View style=Styles.outer>
+    <markdown
+      markdown="
```
OhadRau, line 40: Maybe nitpicky but should probably use something like `{markdown|` ... `|markdown}` instead of quotes -- it works better for multiline strings & theoretically editors can plug into this to interpret the string as markdown. Also, is there a way to pull this string in from a file (other than reading in the file and providing it to Revery)?

src/Core/TextWrapping.re
```diff
@@ -6,7 +6,13 @@ type wrapType =
 let wrapText = (~text, ~measureWidth, ~maxWidth, ~mode) => {
   switch (mode) {
   | NoWrap => [text]
-  | Wrap => Wrap.wrap(text, ~max_width=maxWidth, ~width_of_char=measureWidth)
+  | Wrap =>
+    Wrap.wrap(
+      text,
+      ~max_width=maxWidth,
+      ~width_of_char=measureWidth,
+      ~ignore_preceding_whitespace=false,
```
OhadRau, line 14: I might be wrong but this probably shouldn't be enabled globally -- might make more sense to split this into a different `Wrap` mode (`WrapPrecedingWhitespace` etc.)

src/UI_Components/Markdown.re
```diff
@@ -0,0 +1,531 @@
...
+      fontFamily={styles.fontFamily}
+      fontWeight
+      style={selectStyleFromKind(attrs.kind, styles)}
+      italicized={isItalicized(attrs)}
+      monospaced={isMonospaced(attrs)}
+    />
+  };
+};
+
+let rec generateInline' = (inline, styles, attrs, dispatch, state) => {
```
OhadRau, line 229: Nitpicky again but I'd just nest this into `generateInline` since nothing else needs this function AFAICT. I think it looks a little cleaner to show what scope it belongs to

src/UI_Components/Markdown.re
```diff
@@ -0,0 +1,531 @@
...
+       |> React.listToElement;
+
+     | (_, _) => <View />
+     }}
+  </View>;
+};
+
+let generateList =
+    (
+      blist: Omd__Ast.Block_list.t(Omd.block(Omd.inline)),
```
OhadRau, line 360: This type is annoying enough to probably just get its own name (`omd_inline_block` or something)

src/UI_Components/Markdown.re
```diff
@@ -0,0 +1,531 @@
...
+         </View>;
+       },
+       blist.blocks,
+     )
+     |> React.listToElement}
+  </View>;
+};
+
+let generateBlockquote =
+    (blocks, styles, highlighter, dispatch, state, generateMarkdown') => {
```
OhadRau, line 404: Is the `generateMarkdown'` arg just to get around recursion restrictions? If so you could use `let rec ... and` to clean it up a little (though order matters a little bit since the first one needs to be declared recursive.

examples/MarkdownExample.re
zbaylin: Oh I see what you mean now, it wasn't clear to me when we were talking privately. `make` is just a function used by the `brisk-reconciler` PPX when you write JSX, i.e.
```
<View /> == View.make()
```
The reason I partially applied it like that there was because originally I had multiple markdown components, and I didn't want to type out all of the styles for all of them. Since font styling isn't covered by our `Style` API anymore, in order to generalize like that you have to partially apply, or copy & paste. The former obviously preferable to the latter.

I guess the long winded way to answer your question is that the `make` function is required to be like that, and I should un-partially apply it in the example.

Thanks!

examples/MarkdownExample.re
zbaylin: That last part is a good question. Theoretically there obviously is a way to pull it from a file, but Im not sure if that's the correct level of abstraction. To me I think it makes more sense to pass in strings. We could also have a different component that took a stream or something, that would be interesting. But as for right now @bryphe will have to let me know if that's a proper level of abstraction.

src/UI_Components/Markdown.re
zbaylin: Yeah it is, that's nice to know about `rec` needing to be the first declaration. I tried it with it being 1..nth and it didn't work. Thanks for that!

src/UI_Components/Markdown.re
zbaylin: Im not sure it makes sense to alias it since we only use it once, but I know what you mean. I don't really understand why OMD structured their types like that to be honest.

examples/MarkdownExample.re
OhadRau: Eh nah I think it's fine as is, I just got a little confused since I haven't worked with Revery in a bit. But take a look at https://github.com/facebook/reason/pull/1927, you can do something like `<Markdown(DefaultStyle)>` and `<Markdown(MyFancyStyle)>` with functors here, or you could just give each one a shorthand name `<DefaultMarkdown>` and `<FancyMarkdown>`

Again, the main diff from what you have now is just the ability to inherit styles (which isn't really a possibility with props) and if you're making a module to store all the styles anyways this is pretty similar. One other thing I can think of is you could associate custom syntax highlighters with this module since those are also modules.

examples/MarkdownExample.re
OhadRau: Right, it might be better to leave this feature provided by the user manually (there's always the option for ppx_include I think to do this statically)

src/UI_Components/Markdown.re
OhadRau: Yep np. I think there's a compiler option to fix this but it's not enabled by default. I wonder if this is something that Reason should do (or maybe I could write a quick PPX for it)? Shouldn't be hard since the syntax is still valid in either case, it's just a matter of telling the typechecker which one actually has to be recursive.

examples/MarkdownExample.re
Et7f3: I don't think hardcode fromFile feature is required: we might provide another function file=> string so we can compose it. But if we add direct support then we will have to do this for almost all component m.

src/UI_Components/Markdown.re
Et7f3: Tips from octachron that work because we have ocaml >=4.08
```reason
open {
    let smth = value
}```
smth is hidden from public interface and only accessible in this module.

Comment (Et7f3) [reply src/UI_Components/Markdown.re:229]: And we could safely use that since oni2 require 4.09. I don't know the lower bound of reason parser but shouldn't be an issue seen we require >=3.6.0

File: src/UI_Components/Markdown.re (hunk 6 above)
Comment (Et7f3) [reply src/UI_Components/Markdown.re:360]: You could reduce type size with a `let open Omd;` just after the = and you will need {}

File: src/UI_Components/Markdown.re (hunk 7 above)
Comment (Et7f3) [reply src/UI_Components/Markdown.re:404]: Order doesn't matter when we use `let rec a = [1, ...b] and b = [1, ...a]`

File: examples/MarkdownExample.re (hunk 2 above)
Comment (zbaylin) [reply examples/MarkdownExample.re:25]: Cool, I'm gonna resolve this because I think it's out of this PR's scope but it's definitely a cool idea and something I'll think about!

File: examples/MarkdownExample.re (hunk 3 above)
Comment (OhadRau) [reply examples/MarkdownExample.re:40]: This is true, I just think that in the case of markdown it's similar to how images work -- it's typically an outside document that you're pulling in (and doesn't apply to general properties).

File: src/UI_Components/Markdown.re (hunk 6 above)
Comment (OhadRau) [reply src/UI_Components/Markdown.re:360]: Or `Omd__Ast.Block_list.t(Omd.(block(inline))` if I'm understanding correctly

File: src/UI_Components/Markdown.re (hunk 7 above)
Comment (OhadRau) [reply src/UI_Components/Markdown.re:404]: Yes it doesn't matter if both are recursive, unfortunately if the first isn't recursive OCaml will emit a warning (which will get escalated to an error). E.g.

```
let rec x = 0
and y = y + x
```
will not compile

File: examples/MarkdownExample.re (hunk 3 above)
Comment (zbaylin) [reply examples/MarkdownExample.re:40]: The main use case for this right now is hover support in Oni, where the extensions push the strings from the extension host. But yeah, definitely Markdown is typically loaded from the disk.

File: src/Core/TextWrapping.re (hunk 4 above)
Comment (zbaylin) [reply src/Core/TextWrapping.re:16]: I fixed this in the latest commit. Thanks for reminding me!

File: src/UI_Components/Markdown.re (hunk 5 above)
Comment (zbaylin) [reply src/UI_Components/Markdown.re:239]: I changed this around in the latest commit. Thanks @OhadRau and @Et7f3 !

File: src/UI_Components/Markdown.re (hunk 6 above)
Comment (zbaylin) [reply src/UI_Components/Markdown.re]: I realized this can be cleaned up a lot with `Block_list.t(block(inline))` because I already opened `Omd`. Thanks guys!

File: src/UI_Components/Markdown.re (hunk 7 above)
Comment (zbaylin) [reply src/UI_Components/Markdown.re]: I changed this around when I changed the `generateInline` functions. Thanks!

File: src/UI_Components/Markdown.re (hunk 6 above)
Comment (Et7f3) [reply src/UI_Components/Markdown.re]: @OhadRau not yeah possible https://github.com/ocaml/ocaml/issues/9484

File: src/UI_Components/Markdown.rei [hunk 8]
```diff
@@ -0,0 +1,36 @@
+module SyntaxHighlight: {
```
bryphe, line 1: Minor - some of the names are a bit confusing for me here. If type `t` is the main type of the module - this module seems like it'd really be a `SyntaxHighlighter` - something that produces syntax highlights - versus a 'highlight' itself.

The `block` sounds like it'd be a code-block, which could encompass multiple colors - it seems like `highlight` may be a more accurate name for this.

zbaylin: I fixed this up in the latest commit, sorry for the naming confusion. It's always weird when you have to subdivide concepts further and further 😆

PR #887, opened 2020-06-03
chore: fix gcc 10 issue with secondary compiler

glennsl:
This pins `dune` to 2.5.0 because 2.5.1 requires `ocaml-secondary-compiler`, which is on 4.8.1, which unfortunately doesn't build with gcc 10.


package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@opam/ctypes": "0.15.1",
-    "@opam/dune": ">=1.6.0",
+    "@opam/dune": "2.5.0",
```
Et7f3, line 40:
```suggestion
    "@opam/dune": ">=1.6.0 <=2.5.0",
```
or maybe we expose the tips to "fix" failure because it is safe here. (not for esy-sdl2) https://github.com/revery-ui/revery/issues/866
because we might forget to use the range version when releasing and then we will have 0.3.2+ with very strange deps constraints: Can we try "resolutions" fields ?

glennsl: "expose the tips"?

I agree it would be better to sue resolutions though. Fixed in https://github.com/revery-ui/revery/pull/887/commits/e89373a3393f56f16501e37d8c238f9896223972. This should be patched upstream relatively soon anyway.

Et7f3: The tips is in the issue I have linked.

glennsl: Honestly I wouldn't apply such a hack to my own system, let alone recommend it to others. I think recommending this resolution if people have problems with their own projects should be fine though?

PR #891, opened 2020-06-04
feat(skia): Linear gradients

bryphe:
Simple 2-stop gradient:
image

3-stop gradient:
image

image

### DIFF
--- a/src/reason-skia/wrapped/types/SkiaWrappedTypes.re
+++ b/src/reason-skia/wrapped/types/SkiaWrappedTypes.re
@@ -165,6 +165,37 @@ module M = (T: TYPE) => {
     let t = typedef(t, "sk_point_t");
   };
 
+  module Shader = {
+    
+    type tileMode =
+    | Clamp
+    | Repeat
+    | Mirror;
+
+    let tileMode =
+      skiaCEnum(
+        "sk_shader_tilemode_t",
+        [
+          (
+            Clamp,
+            "CLAMP_SK_SHADER_TILEMODE"
+          ),
+          (
+            Repeat,
+            "REPEAT_SK_SHADER_TILEMODE"
+          ),
+          (
+            Mirror,
+            "MIRROR_SK_SHADER_TILEMODE"
+          )
+        ]
+      );
+    
+    type t;
+    let t: typ(structure(t)) = structure("sk_shader_t");
+    let t = typedef(t, "sk_shader_t");
+  };
+
   module Vector = Point;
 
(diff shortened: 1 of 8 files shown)


src/reason-skia/wrapped/types/SkiaWrappedTypes.re
```diff
@@ -165,6 +165,27 @@ module M = (T: TYPE) => {
...
+
+    let tileMode =
+      skiaCEnum(
+        "sk_shader_tilemode_t",
+        [
+          (Clamp, "CLAMP_SK_SHADER_TILEMODE"),
+          (Repeat, "REPEAT_SK_SHADER_TILEMODE"),
+          (Mirror, "MIRROR_SK_SHADER_TILEMODE"),
+        ],
+      );
```
glennsl, line 182: I've converted some of these enums to polymorphic variants in the SVG branch, just because it was easier than figuring out how to access these types from user code. And I think polymorphic variants are a more natural fit for variants that are just used as flag/enum arguments anyway. Up for debate of course, but I think it's a good idea.

```suggestion
    type tileMode = [
      | `clamp
      | `repeat
      | `mirror
    ];

    let tileMode: T.typ(tileMode) =
      skiaCEnum(
        "sk_shader_tilemode_t",
        [
          (`clamp, "CLAMP_SK_SHADER_TILEMODE"),
          (`repeat, "REPEAT_SK_SHADER_TILEMODE"),
          (`mirror, "MIRROR_SK_SHADER_TILEMODE"),
        ],
      );
```

Et7f3: > access these types from user code

Why pattern matching isn't enough ?

glennsl: The problem I had was constructing a value, not pattern matching, but either way with nominal types you have to reference the module the type is defined in, in some way. And because `reason-skia` is organized in an extremely convoluted way with a poorly defined public interface, I had a hard time figuring out how to get to that module (`SkiaWrapped.Path\ in my case). I'm not even sure it is directly accessible outside the project, except through inference.

So the easiest way to get around that, for me, was to just not use nominal types, to avoid needing to reference said module altogether.

bryphe: Makes sense to me - I saw some issues getting to the SkiaWrappedTypes as well - this seems like a reasonable solution for now.

PR #892, opened 2020-06-04
Use Clickable component in ClickableText

SirFurness:
This stops weird things like right-clicking on a link opening that link.

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -34,6 +34,8 @@ let%component make =
                 ~onKeyUp=?,
                 ~onTextEdit=?,
                 ~onTextInput=?,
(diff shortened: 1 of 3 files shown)


src/UI_Components/Clickable.re
```diff
@@ -34,6 +34,8 @@ let%component make =
                 ~onKeyUp=?,
                 ~onTextEdit=?,
                 ~onTextInput=?,
+                ~onMouseEnter as onMouseEnterUserCallback=?,
+                ~onMouseLeave as onMouseLeaveUserCallback=_evt => (),
```
glennsl, line 38: You don't strictly need to rename these. It might even be beneficial to use shadowing to prevent `onMouseLeaveUserCallback` from being called directly. no big deal though.

Issue #878, opened 2020-06-02
Scale factor calculation on Linux


taylon:
Hi!

I noticed an issue with the scaling factor calculation in Linux that causes a very high scale factor to be set, which in turn makes Oni2 as an example to render like this:

<img src="https://user-images.githubusercontent.com/79991/83483541-bd1b1480-a457-11ea-9b78-4176b8bdd465.png" width="250"/>

Looking at https://github.com/revery-ui/revery/blob/9cf3daa0fa2f54e2b4005d0257e3519738dce37d/src/Core/Window.re#L94-L97 I can see that the the DPI value returned by SDL2 is wrong:

```
[DEBUG]   +20ms Revery.Core.Window : hdpi: 130048.000000 / vdpi: 36576.000000 / ddpi: 95525.617188
[TRACE]    +0ms Revery.Core.Window : _getScaleFactor - Linux - inferring from DPI: 910.000000
```

I'm using a very unusual monitor resolution of 5120x1440 which might be an edge case that causes some weird bug, or it might be X returning it wrong for whatever reason.

Either way, what do you think would be a good mitigation for this?

We could say that if the resulting scaleFactor is higher than a given threshold then assume that the calculation is wrong and set it to 1.0?

zbaylin:
Part of me thinks this is the result of an error that SDL is giving us but we aren't catching. If we look here:
https://github.com/revery-ui/reason-sdl2/blob/fd489ba88457503c899fa06c655601aedb645db8/src/sdl2_wrapper.cpp#L415-L428

We can see that we dont check for an error like is documented here: https://wiki.libsdl.org/SDL_GetDisplayDPI

I can't imagine why it would return values like the ones you posted if it were operating properly. I'll investigate this tonight, sorry you're having this issue!

taylon:
No problem, actually let me investigate that one :)

zbaylin:
@taylon haha be my guest! Let me know what you come up with!

taylon:
So I did some investigation and it is pretty complicated :laughing: 

The values returned by SDL2 are indeed wrong, and that is because the value reported by xrandr itself is wrong (more precisely here [SDL_x11modes.c](https://github.com/revery-ui/esy-sdl2/blob/728d61c15764d50b12b7885642c18e9c80585d21/src/video/x11/SDL_x11modes.c#L427)).

It seems to be a pretty well known issue though, the following links have more details:
https://github.com/glfw/glfw/issues/1019
[...]

PR #896, opened 2020-06-05
fix(rendering/#878): use 1.0 as default scale factor for Linux

taylon:
See #878 for more details :)

### DIFF
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -76,13 +76,19 @@ module WindowMetrics: {
           );
           scale;
 
-        // On Linux, there's a few other things to try:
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for the detailed investigation in #878 , @taylon - the conclusion and proposed fix makes sense to me 👍  Really appreciate you diving in and fixing this!

bryphe:
There was an issue with the Windows build, so I'm kicking that off again to validate... but once that's good, I'll bring this in

bryphe:
CI is green 🎉  

Bringing this in now - thanks again for your investigation and fix here, @taylon !

PR #897, opened 2020-06-05
shorten reason-native-crash-utils commit hash

glennsl:
Apparently `esy` isn't able to unify the full hash and short hash of dependencies, and since it's a pain in the ass to reproduce the full hash in every package that needs it, I opted for just shortening this instead.


Et7f3:
We should opt for long hash because with long hash github allow us to fetch directly that commit. With short hash you have to clone master then checkout

glennsl:
Noted, but IMO having to reproduce the full hash everywhere is a lot more tedious than having to wait a couple extra seconds for a full clone. Unless it's causing other problems.

Referencing commits directly ought to be a lot rarer than it is though, which would help ease it either way.

PR #899, opened 2020-06-05
fix accidentally converting keydown to keyup

despairblue:
This introduced an accidental change converting KeyDown events into KeyUp events. That lead to Inputs not calling their `onKeyDown` handler anymore.

https://github.com/revery-ui/revery/commit/e5f363a1ac1d2b3a91f30ce936efa18478fedb1a#diff-dca70cf54f7954dcef8dd3e027ccb53eR21

This fixes it.


Et7f3:
Thanks that's work perfectly with #828 and it shouldn't break any target. I merge :)

PR #900, opened 2020-06-05
feat(reason-skia): bind to sk_fontmgr_t

zbaylin:
I'm very new to `ctypes`, so please @glennsl or @bryphe let me know if I made any errors (they seem to be pretty easy to make 😳). 

This just binds to Skia's `sk_fontmgr_t`, which provides (I think) everything we use from `reason-font-manager` (along with other classes/types).

I also added the `MatchFamilyStyle` function which should get us closer to emulating what we get from `reason-font-manager`


zbaylin:
Oops, I realized that MatchFamilyStyle can return `nullptr/NULL` according to here: https://api.skia.org/classSkFontMgr.html#afd99f7b624eb7f41bf889036a4e841a8

src/reason-skia/wrapped/bindings/SkiaWrappedBindings.re
```diff
@@ -61,6 +61,23 @@ module M = (F: FOREIGN) => {
...
+    let makeDefault =
+      foreign("sk_fontmgr_create_default", void @-> returning(t));
+
+    let matchFamilyStyle =
+      foreign(
+        "sk_fontmgr_match_family_style",
+        t
+        @-> string
+        @-> FontStyle.t
+        @-> returning(ptr_opt(SkiaTypes.Typeface.t)),
```
bryphe, line 77: Thanks for switching this to model the `NULL` case 👍

bryphe:
Looks good to me too! Thanks for wiring this up, Zach.

Might be worth adding an example usage to `SkiaCli.re` (can be run with `esy '@examples' x SkiaCli`) - this gets run on CI, and helps ensure that the API doesn't cause any crashes, at least

zbaylin:
Awesome, thanks Bryan! The only thing I realized I forgot is that the typeface returned from the Match call must be unref'd so I'm adding that now!

PR #901, opened 2020-06-06
refactor(font): revamp FontFamily and Text APIs

glennsl:
The objective here is primarily to remove ambiguity, and secondarily to reduce verbosity.


src/Font/FontFamily.rei
```diff
@@ -2,22 +2,15 @@ open FontManager;
...
+  ((~weight: FontWeight.t, ~italic: bool, ~mono: bool) => string) => t;
+let fromFile: string => t;
+let system: string => t;
+
+let resolve:
+  (~italic: bool=?, ~mono: bool=?, FontWeight.t, t) =>
+  result(FontCache.t, string);
 
-let toPath: (t, FontWeight.t, bool, bool) => string;
+let toPath: (~italic: bool=?, ~mono: bool=?, FontWeight.t, t) => string;
```
glennsl, line 16: The important part

src/Draw/Text.rei
```diff
@@ -0,0 +1,39 @@
...
+    ~italic: bool=?,
+    ~mono: bool=?,
+    ~fontFamily: Family.t,
+    ~fontSize: float,
+    ~fontWeight: Weight.t,
+    string
+  ) =>
+  dimensions;
+
+let indexNearestOffset: (~measure: string => int, string, int) => int;
```
glennsl, line 39: Also this

glennsl:
> At first I wasnt sure if we should change the names of italicized and monospaced but I think it's better just because it's shorter.

"italic" is also the more common form (by an order of magnitude according to google) and more consistent with properties like "bold" (vs "bolded"). "monospaced" is trickier though, because it doesn't have an adjective form as far as I'm aware, and is soooo long, but can also be ambiguous outside the context of a font, so I've left it in the full form in some places.

PR #902, opened 2020-06-07
feat(reason-skia): bind to `sk_typeface_open_stream` and `sk_stream_t`

zbaylin:
This does a couple main things:
- Creates a `Stream` module
  - has `getLength` & `hasLength` functions
- Adds a `Data.ofStream` function (`sk_data_new_from_stream`)
- Adds a `Typeface.openStream` function (`sk_typeface_open_stream`)
- Writes the found Arial typeface to a file, `Arial.ttf` in the `SkiaFontManagerCli` example

This adds a small C error that Im not sure if there's a mechanism to fix within ctypes. The stream returned by `sk_typeface_open_stream` is an `sk_stream_asset_t` ([`SkStreamAsset`](https://api.skia.org/classSkStreamAsset.html)), which, in the C++/OOP world, is a subclass of `sk_stream_t` ([`SkStream`](https://api.skia.org/classSkStream.html)). However, since there's no notion of inheritance in C the compiler gives an incompatible pointer types warning. We have other semi-similar warnings but those are about type qualifiers like `const`. @bryphe let me know if you think this is something worth trying to fix and, if so, if you know any way to do so.

### DIFF
--- a/src/reason-skia/Skia.rei
+++ b/src/reason-skia/Skia.rei
@@ -29,11 +29,29 @@ module FontStyle: {
 module Hinting: {type t = SkiaWrapped.Hinting.t;};
 module TextEncoding: {type t = SkiaWrapped.TextEncoding.t;};
 
+module Stream: {
+  type t;
+
+  let hasLength: t => bool;
+  let getLength: t => int;
+};
+
+module Data: {
+  type t;
+
+  let makeFromFileName: string => option(t);
+
+  let makeString: t => string;
+
+  let ofStream: (Stream.t, int) => t;
+};
+
 module Typeface: {
   type t;
 
   let makeFromName: (string, FontStyle.t) => option(t);
   let makeFromFile: (string, int) => option(t);
+  let openStream: t => Stream.t;
 };
 
 module FontManager: {
@@ -364,14 +382,6 @@ module Path: {
     ) =>
     unit;
   let close: t => unit;
-};
-
-module Data: {
-  type t;
-
-  let makeFromFileName: string => option(t);
-
-  let makeString: t => string;
 };
 
 module ImageInfo: {
--- a/src/reason-skia/Skia.rei
+++ b/src/reason-skia/Skia.rei
@@ -43,7 +43,7 @@ module Data: {
 
   let makeString: t => string;
 
-  let ofStream: (Stream.t, int) => t;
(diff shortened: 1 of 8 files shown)


src/reason-skia/Skia.rei
```diff
@@ -29,11 +29,29 @@ module FontStyle: {
...
+};
+
+module Data: {
+  type t;
+
+  let makeFromFileName: string => option(t);
+
+  let makeString: t => string;
+
+  let ofStream: (Stream.t, int) => t;
```
glennsl, line 46: This naming is really inconsistent. I understand that at least some of this is inherited, but it really sticks out here.

`makeFrom...` is redundant as `from...` communicates the same concept , but so does `of...` which is also used here. Both are common conventions, but we should choose on and stick with it.

`makeString` on the other hand is not a constructor, despite being name `make...`. It should be `toString` I would think.

src/reason-skia/Skia.rei
```diff
@@ -29,11 +29,29 @@ module FontStyle: {
...
+};
+
 module Typeface: {
   type t;
 
   let makeFromName: (string, FontStyle.t) => option(t);
   let makeFromFile: (string, int) => option(t);
+  let openStream: t => Stream.t;
```
glennsl, line 54: Same issue here with the `makeFrom...`s.

`openStream` is also a bit confusingly named, as I'd expect an `open` to have a corresponding `close` operation. This seems more like a `getStream` operation, or just `stream` for a more declarative and functional API style.

src/reason-skia/Skia.rei
zbaylin: Ah you're right. I was definitely inconsistent there. The `makeFrom`s are from their corresponding Skia function names. I guess I just thought it sounded more OCaml-like to call it `ofStream` 😆. Regardless, I changed it to `makeFromStream` for consistency sake.

src/reason-skia/Skia.rei
zbaylin: I changed this one to `toStream`. I'm totally open to changing all the names of the `makeFrom`s (there's a lot), but I'm not sure if that's in the scope of this PR. If you think it is, let me know!

glennsl: I agree, as long as this PR is consistent it would be better to follow up with a refactor in a later PR.

PR #905, opened 2020-06-08
feat(components): add `underline` prop to Text component

zbaylin:
This is another fairly simple addition. We get the underline info from Skia via FontMetrics and draw a rectangle from it.

Here's what the text looks like all underlined (before I added the property):
Screen Shot 2020-06-08 at 8 53 54 AM

I also added it to the components that use `Text`, like `ClickableText` and `Link`. I made `Link`s underlined by default and `ClickableText`s not.

### DIFF
--- a/src/Font/FontMetrics.re
+++ b/src/Font/FontMetrics.re
@@ -3,7 +3,7 @@ type t = {
   lineHeight: float,
   ascent: float,
   descent: float,
-  underlinePos: float,
+  underlinePosition: float,
   underlineThickness: float,
 };
 
@@ -12,21 +12,21 @@ let empty = (size: float) => {
   lineHeight: size,
   ascent: 0.,
   descent: 0.,
-  underlinePos: 0.,
+  underlinePosition: 0.,
   underlineThickness: 0.,
 };
 
 let ofSkia = (size: float, lineHeight: float, metrics: Skia.FontMetrics.t) => {
   let ascent = Skia.FontMetrics.getAscent(metrics);
   let descent = Skia.FontMetrics.getDescent(metrics);
-  let underlinePos = Skia.FontMetrics.getUnderlinePosition(metrics);
+  let underlinePosition = Skia.FontMetrics.getUnderlinePosition(metrics);
   let underlineThickness = Skia.FontMetrics.getUnderlineThickness(metrics);
   {
     height: size,
     lineHeight,
     ascent,
     descent,
-    underlinePos,
+    underlinePosition,
     underlineThickness,
   };
 };
--- a/src/Font/FontMetrics.re
+++ b/src/Font/FontMetrics.re
@@ -3,6 +3,8 @@ type t = {
   lineHeight: float,
   ascent: float,
   descent: float,
+  underlinePos: float,
+  underlineThickness: float,
 };
 
 let empty = (size: float) => {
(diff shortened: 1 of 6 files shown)


src/UI/TextNode.re
```diff
@@ -180,22 +195,29 @@ class textNode (text: string) = {
...
       | {width: textWidth, _} as style
           when textWidth == Layout.Encoding.cssUndefined =>
         _this#handleTextWrapping(width, style)
       | {textOverflow: Ellipsis | UserDefined(_), _} =>
         _this#textOverflow(float_of_int(width))
       | style => _this#handleTextWrapping(width, style)
-      }
-    );
+      };
+    _textWidth = dims.width;
```
bryphe, line 219: Is this related to the underline? Or does this fix an unrelated bug?

src/UI_Primitives/Text.re
```diff
@@ -20,6 +20,7 @@ let%nativeComponent make =
                       ~italic=false,
                       ~monospaced=false,
                       ~fontSize=14.,
+                      ~underlined=false,
```
bryphe, line 23: Nice! 🎉

src/UI/TextNode.re
bryphe: Ah, I think I understand - this is saving the width so we know how to draw the underline

How does the underline worked in the wrapped case? Is it possible for the wrapped lines to be different lengths?

src/Font/FontMetrics.re
```diff
@@ -3,17 +3,30 @@ type t = {
...
+  underlinePos: float,
+  underlineThickness: float,
 };
 
 let empty = (size: float) => {
   height: size,
   lineHeight: size,
   ascent: 0.,
   descent: 0.,
+  underlinePos: 0.,
```
glennsl, line 15: Nitpicky, but I would prefer this to be spelled out, as it doesn't seem like this is going to be used all that frequently.

src/UI/TextNode.re
```diff
@@ -180,22 +195,29 @@ class textNode (text: string) = {
...
+    _underlined = underlined;
+  };
+  pub measure = (width, _height): LayoutTypes.dimensions => {
     _isMeasured = true;
     /**
          If the width value is set to cssUndefined i.e. the user did not
          set a width then do not attempt to use textOverflow
        */
-    (
+    let dims =
```
glennsl, line 210: Also this:

```suggestion
    let dimensions =
```

I find type annotations to be much less useful and necessary when variables are properly named.

src/UI/TextNode.re
```diff
@@ -180,22 +195,29 @@ class textNode (text: string) = {
...
           when textWidth == Layout.Encoding.cssUndefined =>
         _this#handleTextWrapping(width, style)
       | {textOverflow: Ellipsis | UserDefined(_), _} =>
         _this#textOverflow(float_of_int(width))
       | style => _this#handleTextWrapping(width, style)
-      }
-    );
+      };
+    _textWidth = dims.width;
+    dims;
```
glennsl, line 220:
```suggestion
    dimensions;
```

src/UI/TextNode.re
glennsl, line 219:
```suggestion
    _textWidth = dimensions.width;
```

zbaylin: Yeah it definitely is, thanks for pointing that out. I'll run the text measurement on each line

zbaylin: I fixed this in the latest commits!

src/Font/FontMetrics.re
zbaylin: Sounds good. I changed it in the latest commits (along with all the other naming suggestions)

PR #908, opened 2020-06-09
Ignore double click by default

SirFurness:
A `Clickable` without an `onDoubleClick` will not capture double clicks allowing for normal spam clicking to work.

### DIFF
--- a/src/UI_Components/Clickable.re
+++ b/src/UI_Components/Clickable.re
@@ -24,7 +24,7 @@ let%component make =
                 ~style=[],
                 ~onClick=() => (),
                 ~onRightClick=() => (),
-                ~onDoubleClick=() => (),
+                ~onDoubleClick=?,
                 ~onAnyClick=_event => (),
                 ~componentRef=?,
                 ~onBlur=?,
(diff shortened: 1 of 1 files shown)


src/UI_Components/Clickable.re
```diff
@@ -49,6 +49,17 @@ let%component make =
   let resetMouseDownTimes = () =>
     mouseDownTimes := Constants.initialMouseDownTimes;
 
+  let isUsingDoubleClick = () =>
+    switch (onDoubleClick) {
+    | Some(_) => true
+    | None => false
+    };
```
glennsl, line 56: Can be simplified to:

```suggestion
  let isUsingDoubleClick = onDoubleClick != None;
```

src/UI_Components/Clickable.re
```diff
@@ -49,6 +49,17 @@ let%component make =
...
+  let onDoubleClick = () =>
+    switch (onDoubleClick) {
+    | Some(fn) => fn()
+    | None => ()
+    };
```
glennsl, line 61: Can be simplified to:
```suggestion
  let onDoubleClick = Option.value(onDoubleClick, ~default=() => ());
```

src/UI_Components/Clickable.re
```diff
@@ -76,7 +87,7 @@ let%component make =
 
       switch (mouseEvt.button) {
       | MouseButton.BUTTON_LEFT =>
-        if (isDoubleClick()) {
+        if (isUsingDoubleClick() && isDoubleClick()) {
```
glennsl, line 90: Neither of these really need to be functions I don't think.

PR #910, opened 2020-06-11
Add onSomethingDown callbacks

SirFurness:
Adds callbacks for the different clicks on mouse down.

This is useful for certain things like changing the position of the cursor in an `Input` component on mouse down.


src/UI_Components/Clickable.rei
```diff
@@ -20,9 +20,13 @@ let make:
     ~key: Brisk_reconciler.Key.t=?,
     ~style: list(Revery_UI.Style.viewStyleProps)=?,
     ~onClick: unit => unit=?,
+    ~onDown: unit => unit=?,
     ~onRightClick: unit => unit=?,
+    ~onRightDown: unit => unit=?,
     ~onDoubleClick: unit => unit=?,
+    ~onDoubleDown: unit => unit=?,
     ~onAnyClick: NodeEvents.mouseButtonEventParams => unit=?,
+    ~onAnyDown: NodeEvents.mouseButtonEventParams => unit=?,
```
glennsl, line 29: This should be `onMouseDown`, `onRightMouseDown` etc. to disambiguate it from for example `onKeyDown`.

SirFurness:
You're probably right that it's not a common need. I was mainly thinking in terms of future improvements to the `Input` component. It needs the cursor to change on left click mouse down and it will eventually need a word to be highlighted on double click mouse down.

I don't think any component should use `onMouseDown` from a `View` because it is called for any type of mouse button. This leads to things like the `Slider` component moving with a right-click drag. My current thinking is just that anything that cares about mouse clicks should use a `Clickable` so that things like `onMouseDown` work only with left clicks and code checking for what mouse button was used does not need to be duplicated. It could also make the intent more clear (e.g. that `Slider` is meant to be clicked).

I guess it depends on how the `Clickable` component is meant to be used. Maybe there should be a different component?

glennsl:
I'm honestly not really sure what `Clickable` is supposed to be either, as the events it exposes could easily have been made available directly on `View`. 

> I don't think any component should use onMouseDown from a View because it is called for any type of mouse button. This leads to things like the Slider component moving with a right-click drag.

The mouse button is passed in the events args. That's how you're able to implement this. And this is analogous to how `onMouseDown` works on the Web. For something as rarely needed as this I personally think that's fine.

Left-only click, right click and double click is common enough that I think it warrants separate events, and they fit well with `Clickable`. But the rest I'm not so sure about design-wise. It would be nice to have a holistic idea and some overall guiding principles for where we're going with this I think.

SirFurness:
Sounds good! A document with some guiding design principles would be quite nice, especially with regards to big things like #489.

PR #911, opened 2020-06-12
fix(render): fix text getting cut off from improper int <-> float conversion

zbaylin:
Previously we were casting the float returned from the text wrap call directly to an int. This was causing remeasures to improperly wrap because we shaved off the fractional space required for the text.

### DIFF
--- a/src/UI/TextNode.re
+++ b/src/UI/TextNode.re
@@ -9,6 +9,10 @@ open Revery_Font;
 open Style;
 open ViewNode;
 
+open {
+       let int_of_float_ceil = f => int_of_float(f +. 1.);
+     };
+
 class textNode (text: string) = {
   as _this;
   val mutable text = text;
@@ -180,8 +184,8 @@ class textNode (text: string) = {
          );
 
     {
-      width: int_of_float(Float.ceil(width)),
-      height: int_of_float(Float.ceil(lineHeightPx)),
+      width: int_of_float_ceil(width),
+      height: int_of_float_ceil(lineHeightPx),
     };
   };
   pub setText = t =>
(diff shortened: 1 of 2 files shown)


src/UI/TextNode.re
```diff
@@ -179,7 +179,10 @@ class textNode (text: string) = {
            _fontWeight,
          );
 
-    {width: int_of_float(width), height: int_of_float(lineHeightPx)};
+    {
+      width: int_of_float(Float.ceil(width)),
+      height: int_of_float(Float.ceil(lineHeightPx)),
+    };
```
bryphe, line 185: Thanks for catching this, @zbaylin !

src/UI/TextNode.re
```diff
@@ -272,9 +290,11 @@ class textNode (text: string) = {
     };
     let maxWidthLine = List.fold_left(pickWiderLine, 0., _lines);
     {
-      width: int_of_float(maxWidthLine),
+      width: int_of_float(Float.ceil(maxWidthLine)),
       height:
-        int_of_float(float_of_int(List.length(_lines)) *. lineHeightPx),
+        int_of_float(
+          Float.ceil(float_of_int(List.length(_lines)) *. lineHeightPx),
+        ),
```
glennsl, line 297: Protip: A quick and convenient way of doing `ceil` when you're already converting it to an int is to just add `1.`:

```suggestion
        int_of_float((float_of_int(List.length(_lines)) *. lineHeightPx) + 1.),
```

And for rounding just add `0.5` instead.

zbaylin: Oh thanks for that! I'll make a convenience function for it at the top

PR #917, opened 2020-06-22
fix(markdown): change syntax highlight API to take an option(string) for language

zbaylin:
This should help resolve the case in onivim/oni2#1947. This makes sure the passed in highlighter always gets called, even if there is no language.

### DIFF
--- a/src/UI_Components/Markdown.re
+++ b/src/UI_Components/Markdown.re
@@ -449,7 +449,7 @@ and generateCodeBlock =
        )
        |> React.listToElement;
 
-     | _ => <View />
+     | None => <View />
      }}
   </View>;
 };
--- a/src/UI_Components/Markdown.re
+++ b/src/UI_Components/Markdown.re
@@ -44,6 +44,7 @@ type style = {
   h6: list(Style.textStyleProps),
   fontFamily: Family.t,
   codeFontFamily: Family.t,
+  codeBlockFontSize: float,
   baseFontSize: float,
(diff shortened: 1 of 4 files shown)


src/UI_Components/Markdown.re
```diff
@@ -408,19 +409,7 @@ and generateCodeBlock =
...
-         monospaced=true
-         fontSize={styles.codeBlockFontSize}
-         style=Style.[
-           textWrap(TextWrapping.WrapIgnoreWhitespace),
-           ...{styles.paragraph},
-         ]
-       />
-
-     | (Some(label), Some(code)) =>
+     | (label, Some(code)) =>
```
glennsl, line 412: There's no reason to match on `label` here, as far as I can see, as it's just rebound to the same name.

src/UI_Components/Markdown.re
```diff
@@ -449,7 +449,7 @@ and generateCodeBlock =
        )
        |> React.listToElement;
 
-     | (_, _) => <View />
+     | _ => <View />
```
glennsl, line 452: It's also good practice to avoid the wildcard pattern and enumerate all possible patterns unless it's very tedious, as that enables exhaustiveness checking which helps catch bugs! 

It can also be very useful to drive development when adding a new constructor to a variant like `msg` for example. If you put dune in watch mode you can then have the compiler tell you step-by-step where you need to go to make changes.

```suggestion
     | None => <View />
```

PR #918, opened 2020-06-23
better handling of platforms without a font-manager

EduardoRFS:
## What?

* switch config/discover.ml to config/discover.re
* use feature detection on config/discover.re
* support platforms without / broken font-config
* add new findFontOpt API

## Why? 

- Mostly Android and iOS. When you're cross compiling, `uname` doesn't work, so we needed another system detection mechanism, but overall feature detection is a better idea. This new system will also enable it to work on weird platforms.

- Android doesn't have a FontManager like `font-config` so we needed a dummy.

- The `findFontOpt` API is needed because there is already some cases where `findFont` can return NULL, and now the dummy also does that, so we needed to handle it, previously it would just crash the entire app. This can also be used to implement a fallback mechanism.


src/reason-font-manager/FontManager.cpp
```diff
@@ -1,46 +1,55 @@
...
+  caml_release_runtime_system();
+
+  FontDescriptor *query = new FontDescriptor(NULL, NULL, fontFamily, NULL, weightToCheck, widthToCheck, isItalic, isMono);
+  FontDescriptor *font = findFont(query);
+
+  delete query;
+
+  if (font == NULL) {
+    caml_raise_not_found();
+    return Val_unit;
```
bryphe, line 36: Does this need to be `CAMLreturn(Val_unit)`?

src/reason-font-manager/config/discover.re
```diff
@@ -0,0 +1,71 @@
...
+  test: Configurator.t => bool,
+};
+
+let root = Sys.getenv("cur__root");
+
+let ccopt = s => ["-ccopt", s];
+let cclib = s => ["-cclib", s];
+
+let check_headers = (headers, conf) => {
+  let code = String.concat("\n", headers);
```
bryphe, line 16: Neat!

src/reason-font-manager/config/discover.re
```diff
@@ -0,0 +1,71 @@
...
+  ...feature,
+  flags: [] @ cclib("-lfontconfig") @ cclib("-lstdc++"),
+  cxx_flags: ["-fPIC", "-lstdc++"],
+  test: check_headers(["#include <fontconfig/fontconfig.h>"]),
+};
+let windows_direct_write = {
+  ...feature,
+  flags: [] @ cclib("-ldwrite") @ cclib("-lstdc++"),
+  cxx_flags: ["-fno-exceptions", "-fno-rtti", "-lstdc++"],
+  test: check_headers(["#include <dwrite.h>", "#include <dwrite_1.h>"]),
```
bryphe, line 45: Thanks for moving us out of the `uname` world and into the feature-detection world - this is a great example to follow and more robust.

src/reason-font-manager/FontManager.re
```diff
@@ -2,9 +2,33 @@ module FontDescriptor = FontDescriptor;
 module FontWeight = FontWeight;
 module FontWidth = FontWidth;
 
-external _findFont: (string, int, int, bool, bool) => FontDescriptor.raw =
+external findFont': (string, int, int, bool, bool) => FontDescriptor.raw =
   "fm_findFont";
 
+let findFontOpt =
```
bryphe, line 8: We've been using (or moving) to the reverse convention:
- a `findFont: (...) => option(FontDescriptor.raw)`
- a `findFontExn: (...) => FontDescriptor.raw`

Assuming the 'default' case is to use the type system to express a font being available

I'm not sure if we documented this at all in Revery... we have it in our Onivim style guide, though: https://onivim.github.io/docs/for-developers/style-guide#exceptions-vs-result

src/reason-font-manager/FontManager.cpp
bryphe: Also, do we need to `caml_acquire_runtime_system` prior to `caml_raise_not_found()`?

EduardoRFS: Noice catch, It doesn't need the `CAMLreturn`, but it's good idea to have.

Yeah should acquire the runtime system, interesting because it was working

src/reason-font-manager/config/discover.re
EduardoRFS: Slowly as the requirements are more explicit we can build our own feature detection mechanism too

src/reason-font-manager/FontManager.re
EduardoRFS: Oh I like that, was trying to avoid a breaking change, but that's a lot better

bryphe:
I tested this on Windows, and everything is working (both the example app, and the `FontQuery` CLI tool)

PR #919, opened 2020-06-23
sdl2 escapes keycodes to support 32bits

EduardoRFS:
## Why?

OCaml on 32bits systems, have an int size of 31 bits, but SDL keycodes are actually 32bits integers, and some of them actually uses 31 bits of the integer, like `let right = 1073741903`.

One platform that suffers of this is problem is the Raspberry Pi, which by default is 32bits.

## Solution

There is no SDL keycode negative, so to fix that without a breaking change we can use the negative part of our 31 bits int to fit any SDL keycode.

The only problem is that if you try to serialize big keycodes they will be different based on the platform, that is not ideal, but seems like a really small edge case and changing everything to `int32` is probably not worth it.

### DIFF
--- a/src/reason-sdl2/sdl2.re
+++ b/src/reason-sdl2/sdl2.re
@@ -301,6 +301,9 @@ module Scancode = {
 
 module Keycode = {
   type t = int;
+  // as some keycodes are bigger than 2^30 but still less than 2^31
+  // so by using the negative numbers we can fit them in a normal int
+  let int = Int32.to_int;
 
   external getName: t => string = "resdl_SDL_GetKeyName";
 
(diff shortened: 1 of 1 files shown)


src/reason-sdl2/sdl2.re
```diff
@@ -313,87 +316,87 @@ module Keycode = {
...
-  let x = 120;
-  let y = 121;
-  let z = 122;
-
-  let delete = 127;
-
-  let right = 1073741903;
-  let left = 1073741904;
+  let unknown = int(0l);
+  let backspace = int(8l);
```
Et7f3, line 320: I don't remember the prefix but their is a prefix to cast at compile time here it is at runtime

EduardoRFS: Hmmm couldn't found anything related to that, but this happens at compile time anyways, even in classic mode and even in bytecode mode. Also this isn't a safe transformation, technically it's not a type cast as you loose one bit

bryphe:
Looks good to me! Thanks for fixing this, @EduardoRFS . Exciting to have it work on Raspberry Pi 😍

PR #920, opened 2020-06-24
sdl2 better feature and system detection

EduardoRFS:
## Why?

By removing the use of `uname` we can cross compile from Linux to Mac / Window and from Mac to Linux / Windows. It's a neat tooling for development as compiling from a powerful Linux computer is a lot faster than compiling on Windows(especially for me).

Also used the `SDL_config.h` to detect features available, this should also fix **wayland** and it's half of the problem regarding iOS and Android.

## Why not full feature detection?

Feature detection for this specific case is quite hard, I plan to have it, but in the futuru, also there is some options with are inherently system based. So this is a middle ground for now but may change in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,13 +22,12 @@
         "scope": "global"
       }
     },
-    "build": [
-      "dune build -p font-manager,harfbuzz,skia,Revery -j4"
-    ],
+    "build": "dune build -p font-manager,harfbuzz,skia,sdl2,Revery",
     "install": [
       "esy-installer font-manager.install",
       "esy-installer harfbuzz.install",
       "esy-installer skia.install",
+      "esy-installer sdl2.install",
       "esy-installer Revery.install",
       "bash -c \"#{os == 'windows' ? 'cp /usr/x86_64-w64-mingw32/sys-root/mingw/bin/*.dll \\'$cur__bin\\'': ':'}\"",
       "bash -c \"cp #{esy-skia.bin}/skia.dll \\'$cur__bin\\' #{os == 'windows' ? '' : '2>/dev/null || true'}\"",
@@ -72,4 +71,3 @@
     "@opam/ocaml-lsp-server": "ocaml/ocaml-lsp:ocaml-lsp-server.opam#38bd51a15c98b4f6ff841e5c914a8cdacee15ea6"
   }
 }
-
--- a/sdl2.opam
+++ b/sdl2.opam
@@ -1,0 +1,7 @@
+opam-version: "1.2"
+version: "dev"
+maintainer: "bryphe@outlook.com"
+author: ["Bryan Phelps"]
+build: [
+
+]
--- a/src/reason-sdl2/config/discover.re
+++ b/src/reason-sdl2/config/discover.re
@@ -1,26 +1,40 @@
+module Configurator = Configurator.V1;
+
 type os =
-  | Windows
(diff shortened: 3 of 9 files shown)


src/reason-sdl2/sdl2_wrapper.cpp
```diff
@@ -1287,19 +1299,15 @@ CAMLprim value resdl_SDL_CreateWindow(value vName, value vX, value vY,
...
-  SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-  SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
 #else
-  // There's no guarantee that Linux 3.0 is available on Linux.
-  // ie, on my CentOS 6 box, with latest Intel drivers - only 2.1 is supported.
+  SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+#endif
+
   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
   SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
```
EduardoRFS, line 1309: This is just the minimal version, in our case it will be the minimum of `OpenGL 2.0` or `OpenGLES 2.0` but the newest compatible version will always be the chosen one.

src/reason-sdl2/config/discover.re
```diff
@@ -1,26 +1,40 @@
+module Configurator = Configurator.V1;
+
 type os =
-  | Windows
-  | Mac
   | Linux
-  | Unknown;
```
EduardoRFS, line 5: Removed `Unknown` as being sincerely we're not going to handle this case and it was just adding noise

package.json
```diff
@@ -22,9 +22,7 @@
         "scope": "global"
       }
     },
-    "build": [
-      "dune build -p font-manager,harfbuzz,skia,Revery -j4"
-    ],
+    "build": "dune build -p font-manager,harfbuzz,skia,sdl2,Revery",
```
EduardoRFS, line 25: Removed the -j4, as it's a rate limit, if it was desired to limit it, maybe we can discuss another limit? -j4 seems quite low for 2020

sdl2.opam
```diff
@@ -0,0 +1,7 @@
+opam-version: "1.2"
+version: "dev"
+maintainer: "bryphe@outlook.com"
+author: ["Bryan Phelps"]
+build: [
+
+]
```
bryphe, line 7: Thanks for creating a package!

src/reason-sdl2/config/discover.re
bryphe: Sounds good 👍

src/reason-sdl2/sdl2_wrapper.cpp
```diff
@@ -208,24 +209,35 @@ CAMLprim value resdl_SDL_GetNativeWindow(value vWin) {
 
   void *pNativeWindow = NULL;
   switch (wmInfo.subsystem) {
-#ifdef WIN32
+#ifdef SDL_VIDEO_DRIVER_WINDOWS
```
bryphe, line 212: Thanks for cleaning these up!

PR #922, opened 2020-06-25
refactor(examples): move examples to ReveryExamples package

zbaylin:
This moves the examples from the Revery package to the ReveryExamples package.

This has two advantages:
- Revery no longer builds the examples as part of the library. It doesn't really make sense for a consumer to build all the examples when they wont use them. Unless I'm missing something, this seems like a positive.
- The examples can use the Revery library as an external consumer. This will help to get separate bytecode and native builds working for the examples.

### DIFF
--- a/examples.json
+++ b/examples.json
@@ -1,6 +1,7 @@
 {
   "source": "./package.json",
   "scripts": {
+      "run": "esy x Examples"
   },
   "override": {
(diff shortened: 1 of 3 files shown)


zbaylin:
Ah @Et7f3 I think you're right. I'll change that now.

PR #925, opened 2020-06-27
feat(font-manager): fallback when Font is not found

EduardoRFS:
## Why?

After #918 instead of the applications crashing when a font is not found it throws a `Font_not_found` , but there is some platforms like Android who will never have a font because it's using a `dummy` font-manager, so we need a way to handle that.

## Solution

This is not a great API, but shouldn't be a problem considering that this kinda of choice should be handled on an application level. We can mark it as unstable or deprecated.

Still proposed it, because it's a nice to have even if it's not great and we can use it to develop a proper implementation to replace it later.

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -363,4 +363,17 @@ let init = app => {
   ();
 };
 
+Revery.Font.Discovery.callback_font_not_found :=
+  (
+    (~weight, ~width, ~italic, ~mono, family) => {
+      path: Environment.getAssetPath("Roboto-Bold.ttf"),
+      postscriptName: "Roboto-Bold",
+      family: "Roboto",
+      weight: Normal,
+      width: Normal,
+      italic: false,
+      monospace: false,
+    }
+  );
+
 App.start(init);
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -363,17 +363,17 @@ let init = app => {
   ();
 };
 
-Revery.Font.Discovery.callback_font_not_found :=
-  (
-    (~weight, ~width, ~italic, ~mono, family) => {
-      path: Environment.getAssetPath("Roboto-Bold.ttf"),
-      postscriptName: "Roboto-Bold",
-      family: "Roboto",
-      weight: Normal,
-      width: Normal,
-      italic: false,
-      monospace: false,
-    }
-  );
+Revery.Font.Discovery.setFallbackResolver(
+  (~weight as _, ~width as _, ~mono as _, ~italic as _, _) =>
+  {
+    path: Environment.getAssetPath("Roboto-Bold.ttf"),
+    postscriptName: "Roboto-Bold",
+    family: "Roboto",
+    weight: Normal,
+    width: Normal,
+    italic: false,
+    monospace: false,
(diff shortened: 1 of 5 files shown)


examples/Examples.re
```diff
@@ -363,4 +363,17 @@ let init = app => {
   ();
 };
 
+Revery.Font.Discovery.callback_font_not_found :=
```
zbaylin, line 366: I think, even if we use a ref on the backend, this should be a function. Something like `onFontNotFound` or something like that. Also, we generally camel case everything, just for reference.

src/Font/Discovery.re
```diff
@@ -4,6 +4,11 @@
 */
 open FontManager;
 
+let callback_font_not_found =
+  ref((~weight, ~width, ~italic, ~mono, family) =>
+    raise(FontManager.Font_not_found)
+  );
+
```
zbaylin, line 11: Yeah this is fine I think, I would just add a separate function to set it rather than making it visible globally. @glennsl is a much bigger expert on this stuff than I am so maybe he can weigh in 👍. Also the camel case I mentioned before.

src/Font/Discovery.re
```diff
@@ -12,6 +17,19 @@ let find =
       ~italic=false,
       family,
     ) =>
-  FontManager.findFontExn(~weight, ~width, ~mono, ~italic, ~family, ());
+  FontManager.findFont(~weight, ~width, ~mono, ~italic, ~family, ());
+
+let findExn =
```
zbaylin, line 22: Awesome 😃

examples/Examples.re
zbaylin: `onFountNotFound` is probably not descriptive enough, maybe like `setFallbackFont`

zbaylin:
This is good to have, thanks @EduardoRFS! It might change a bit next week, though, when @bryphe and I work on font fallback as a general feature. But for now, even if it doesn't exist for super long, I think it's worth having.

Thanks again!

examples/Examples.re
glennsl: `setFallbackResolver` maybe?

src/Font/Discovery.re
glennsl: Agree on all counts. Completely exposed global refs makes my timbers shiver, so while this isn't that much of a restriction, anything is better than nothing!

examples/Examples.re
EduardoRFS: Agreed on all points, I was trying to make this API feels off place intentionally, but if you guys are okay with that will make the requested changes in some hours.

src/Font/Discovery.re
```diff
@@ -4,6 +4,11 @@
 */
 open FontManager;
 
+let callbackFontNotFound =
```
zbaylin, line 7: One more small piece of feedback :) -- there should probably be a `.rei` file here to hide the ref from being modified

EduardoRFS: This module is already private, Revery_Font.re hides it, you can try to get it using the autocomplete, it's not going to work

PR #926, opened 2020-06-27
Build update skia

Et7f3:
I have rebased revery-ui/esy-skia on top of mono/skia

Since their is a 10k diff we might copy master in new branch (to not break user that haven't updated because commit are kept)
and I force push the rebased branch in place of master

Question: in drawString method we have a textEncoding and we use always GlyphId (in one test we use Utf8) should we expose textEncoding arguments ?

todo: remove Deprecated module ? (I don't see any mention of a deprecation so it should be fine)


Et7f3:
Also SkiaSdl return a exit code of 2 and don't do anything.

PR #927, opened 2020-06-29
feat(harfbuzz): initial Skia integration

zbaylin:
This does a couple of things:
- Deprecates `hb_new_face` by renaming it to `hb_face_from_path`
  - A deprecation warning is added in case anyone else uses this API
- Refactor the C codepath of creating a new Harfbuzz face to minimize duplication
- Adds a `hb_face_from_skia` which takes a Skia typeface and returns a face
- Internally adds a `hb_face_from_data` which takes a string (blob) and a length and returns a typeface
- Moves the HarfbuzzCLI example to use the new Skia-based API

### DIFF
--- a/src/reason-harfbuzz/Harfbuzz.re
+++ b/src/reason-harfbuzz/Harfbuzz.re
@@ -37,7 +37,7 @@ let hb_face_from_skia = sk_typeface => {
   let bytes = Skia.Data.makeString(data);
 
   switch (Internal.hb_face_from_bytes(bytes, String.length(bytes))) {
-  | Error(msg) as e => e
+  | Error(_) as e => e
   | Ok(face) =>
     let ret = {face: face};
 
--- a/src/reason-harfbuzz/Harfbuzz.re
+++ b/src/reason-harfbuzz/Harfbuzz.re
@@ -5,16 +5,39 @@ type hb_shape = {
 
 module Internal = {
   type face;
-  external hb_new_face: string => result(face, string) = "rehb_new_face";
+  external hb_face_from_path: string => result(face, string) =
+    "rehb_face_from_path";
+  external hb_face_from_bytes: (string, int) => result(face, string) =
+    "rehb_face_from_bytes";
   external hb_destroy_face: face => unit = "rehb_destroy_face";
   external hb_shape: (face, string) => array(hb_shape) = "rehb_shape";
 };
 
 type hb_face = {face: Internal.face};
 
-let hb_new_face = str => {
-  switch (Internal.hb_new_face(str)) {
+let hb_face_from_path = str => {
+  switch (Internal.hb_face_from_path(str)) {
   | Error(msg) => Error(msg)
+  | Ok(face) =>
+    let ret = {face: face};
+
+    let finalise = ({face}) => Internal.hb_destroy_face(face);
+
+    Gc.finalise(finalise, ret);
+    Ok(ret);
+  };
+};
+
+let hb_new_face = str => hb_face_from_path(str);
+
+let hb_face_from_skia = sk_typeface => {
+  let stream = Skia.Typeface.toStream(sk_typeface);
+  let length = Skia.Stream.getLength(stream);
+  let data = Skia.Data.makeFromStream(stream, length);
+  let bytes = Skia.Data.makeString(data);
+
+  switch (Internal.hb_face_from_bytes(bytes, String.length(bytes))) {
+  | Error(msg) as e => e
   | Ok(face) =>
     let ret = {face: face};
 
--- a/src/reason-harfbuzz/Harfbuzz.re
+++ b/src/reason-harfbuzz/Harfbuzz.re
@@ -7,7 +7,7 @@ module Internal = {
   type face;
   external hb_face_from_path: string => result(face, string) =
     "rehb_face_from_path";
-  external hb_face_from_bytes: (string, int) => result(face, string) =
+  external hb_face_from_data: (string, int) => result(face, string) =
     "rehb_face_from_bytes";
   external hb_destroy_face: face => unit = "rehb_destroy_face";
   external hb_shape: (face, string) => array(hb_shape) = "rehb_shape";
@@ -36,7 +36,7 @@ let hb_face_from_skia = sk_typeface => {
   let data = Skia.Data.makeFromStream(stream, length);
   let bytes = Skia.Data.makeString(data);
 
-  switch (Internal.hb_face_from_bytes(bytes, String.length(bytes))) {
+  switch (Internal.hb_face_from_data(bytes, String.length(bytes))) {
   | Error(_) as e => e
   | Ok(face) =>
     let ret = {face: face};
--- a/src/reason-harfbuzz/Harfbuzz.rei
+++ b/src/reason-harfbuzz/Harfbuzz.rei
@@ -5,5 +5,10 @@ type hb_shape = {
   cluster: int,
 };
 
+let hb_face_from_path: string => result(hb_face, string);
+let hb_face_from_skia: Skia.Typeface.t => result(hb_face, string);
+
+[@ocaml.deprecated "Deprecated in favor of hb_face_from_path"]
 let hb_new_face: string => result(hb_face, string);
+
 let hb_shape: (hb_face, string) => array(hb_shape);
--- a/src/reason-harfbuzz/harfbuzz.cpp
+++ b/src/reason-harfbuzz/harfbuzz.cpp
@@ -73,7 +73,7 @@ CAMLprim value rehb_face_from_path(value vString) {
   FILE *file = fopen(szFont, "rb");
 
   if (!file) {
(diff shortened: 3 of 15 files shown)


src/reason-harfbuzz/Harfbuzz.re
```diff
@@ -25,4 +28,24 @@ let hb_new_face = str => {
...
 
+let hb_new_face = str => hb_face_from_path(str);
+
+let hb_face_from_skia = sk_typeface => {
+  let stream = Skia.Typeface.toStream(sk_typeface);
+  let length = Skia.Stream.getLength(stream);
+  let data = Skia.Data.makeFromStream(stream, length);
+  let bytes = Skia.Data.makeString(data);
+
+  switch (Internal.hb_face_from_data(bytes, String.length(bytes))) {
```
zbaylin, line 39: At first glance this may seem like it wont always work because there could be a `\0` anywhere, but according to the OCaml FFI docs I reference, OCaml strings can handle `\0`'s in them:

https://web.archive.org/web/20190503112320/https://www.linux-nantes.org/~fmonnier/OCaml/ocaml-wrapping-c.html#ref_rawdata

This seems to be true because the CLI test works and it seems unlikely that there would be 0 `NULL` bytes in the blob

zbaylin:
I'm having a hard time debugging that memory leak. I cant duplicate it on my local machine with macOS. If I run it in Instruments.app, I get no leaks:
image

I wonder if is Linux-specific or maybe another fontconfig problem?

zbaylin:
I'm fairly confident that the leak is from fontconfig, and the reason it's not suppressed is because ASAN isn't recording the callstack deep enough. If I test it on my local machine, I get the same leaks, except mine are labeled as being from `fontconfig`. Not sure why it's different (possibly differing versions of `gcc` or ASAN or something), but the leaks look exactly the same:

<pre><font color="#CC0000"><b>==80510==ERROR: LeakSanitizer: detected memory leaks</b></font>

<font color="#3465A4"><b>Direct leak of 1536 byte(s) in 2 object(s) allocated from:</b></font>
    #0 0x7f811bc31ffe in __interceptor_realloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dffe)
    #1 0x7f811bafe3a0  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x213a0)

<font color="#3465A4"><b>Direct leak of 256 byte(s) in 1 object(s) allocated from:</b></font>
    #0 0x7f811bc31bc8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dbc8)
    #1 0x7f811bafe2f4  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x212f4)

<font color="#3465A4"><b>Indirect leak of 1536 byte(s) in 48 object(s) allocated from:</b></font>
    #0 0x7f811bc31dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7f811bafde0f  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x20e0f)

<font color="#3465A4"><b>Indirect leak of 544 byte(s) in 17 object(s) allocated from:</b></font>
    #0 0x7f811bc31dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7f811bafe93c  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x2193c)

<font color="#3465A4"><b>Indirect leak of 468 byte(s) in 26 object(s) allocated from:</b></font>
    #0 0x7f811bbba3dd in strdup (/lib/x86_64-linux-gnu/libasan.so.5+0x963dd)
    #1 0x7f811bafdb64 in FcValueSave (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x20b64)

<font color="#3465A4"><b>Indirect leak of 384 byte(s) in 12 object(s) allocated from:</b></font>
    #0 0x7f811bc31dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7f811bafdcae  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x20cae)

<font color="#3465A4"><b>Indirect leak of 96 byte(s) in 2 object(s) allocated from:</b></font>
    #0 0x7f811bc31bc8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dbc8)
    #1 0x7f811baf7e11 in FcLangSetCreate (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x1ae11)

SUMMARY: AddressSanitizer: 4820 byte(s) leaked in 108 allocation(s).
</pre>

Not sure what the best solution is for this, but maybe I just temporarily disable ASAN until we can find a way to suppress __all__ fontconfig leaks?

src/reason-harfbuzz/harfbuzz.cpp
```diff
@@ -76,14 +64,46 @@ CAMLprim value rehb_destroy_face(value vFont) {
...
+CAMLprim value rehb_face_from_path(value vString) {
   CAMLparam1(vString);
   CAMLlocal1(ret);
 
   const char *szFont = String_val(vString);
+  
+  FILE *file = fopen(szFont, "rb");
+
+  if (!file) {
+    return NULL;
```
bryphe, line 76: Just noticed this - this should be `CAMLreturn(Val_unit)`. Could cause a GC crash. 

Not a bug you introduced, this was around before, just got noticed because it moved around 😄  Would you mind fixing it as part of this PR, though?

src/reason-harfbuzz/harfbuzz.cpp
```diff
@@ -9,6 +9,8 @@
 #include <hb-ot.h>
 #include <hb.h>
 
+#include "ctypes_cstubs_internals.h"
```
bryphe, line 12: Is this needed?

src/reason-harfbuzz/Harfbuzz.rei
```diff
@@ -5,5 +5,10 @@ type hb_shape = {
   cluster: int,
 };
 
+let hb_face_from_path: string => result(hb_face, string);
+let hb_face_from_skia: Skia.Typeface.t => result(hb_face, string);
+
+[@ocaml.deprecated "Deprecated in favor of hb_face_from_path"]
```
bryphe, line 11: Thanks for adding this 👍

src/reason-harfbuzz/harfbuzz.cpp
zbaylin: Ah not anymore! Thanks!

src/reason-harfbuzz/harfbuzz.cpp
zbaylin: Yep sounds good!

zbaylin: Since the return type for the function should be a `result` I instead changed it to `CAMLreturn(Val_error("File not found"))` 👍

src/reason-harfbuzz/harfbuzz.cpp
```diff
@@ -76,14 +63,46 @@ CAMLprim value rehb_destroy_face(value vFont) {
...
+CAMLprim value rehb_face_from_path(value vString) {
   CAMLparam1(vString);
   CAMLlocal1(ret);
 
   const char *szFont = String_val(vString);
+  
+  FILE *file = fopen(szFont, "rb");
+
+  if (!file) {
+    CAMLreturn(Val_error("File does not exist"));
```
bryphe, line 75: Awesome, thanks for fixing this!

src/reason-harfbuzz/Harfbuzz.re
```diff
@@ -25,4 +28,24 @@ let hb_new_face = str => {
...
+  | Error(_) as e => e
+  | Ok(face) =>
+    let ret = {face: face};
+
+    let finalise = ({face}) => Internal.hb_destroy_face(face);
+
+    Gc.finalise(finalise, ret);
+    Ok(ret);
+  };
+};
```
glennsl, line 49: Sorry I didn't see this before, but this tight coupling between `harfbuzz` and `skia` is really unfortunate. It also seems unnecessary, as it could just as well have been implemented as separate `toData`/`fromData` functions in `skia` and `harfbuzz` respectively. Unless I'm missing something here I'd like to do just that in order to remove this coupling.

zbaylin: That's fine, the reason I did it like that was because I think I had plans to reduce the amount of copying the typeface back and forth in memory. I was concerned that since fonts can be several MB that it might be a severe perf impact, but it didnt turn out to be that way. I can make a PR to uncouple them if you want!

glennsl: Yeah that would be great. Thanks!

For v2 I'm trying to make these packages self-contained precisely to make it harder to do stuff like this, as well as prevent packages from using each others internal modules directly, which is also done in several places.

PR #928, opened 2020-06-30
fix(sdl2): add missing stubs

lessp:


src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
 function resdl_SDL_GL_SwapWindow() {
     // no op
 }
+
+// Provides: resdl_SDL_GetNumVideoDisplays
+function resdl_SDL_GetNumVideoDisplays() {
+    return 1;
```
lessp, line 618: Not sure if we can get the the number of displays via the web api. Are we OK with simply returning 1 here?

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+    if (value) {
+        return value.toString();
+    } else {
+        return ""
+    }
+}
+
+// Provides: resdl_SDL_GetWindowPixelFormat
+function resdl_SDL_GetWindowPixelFormat(_win) {
+    return "SDL_PIXELFORMAT_ARGB8888";
```
lessp, line 632: Not sure what we'd like to return here? Some sort of `Unknown` tag perhaps?

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+}
+
+// Provides: resdl_SDL_GetPixelFormatName
+function resdl_SDL_GetPixelFormatName(value) {
+    return "resdl_SDL_GetPixelFormatName: Not implemented";
```
lessp, line 637: Same here.

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+}
+
+
+// Provides: resdl_SDL_GetLinkedVersion
+function resdl_SDL_GetLinkedVersion() {
+    return 1;
```
lessp, line 643: Same here.

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+}
+
+// Provides: resdl_SDL_GetScancodeFromName
+function resdl_SDL_GetScancodeFromName() {
+    return "scancodeFromName";
```
lessp, line 648: Both this and `GetKeyFromName` could probably be retrieved from the web api, I'd think!

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+}
+
+// Provides: resdl_SDL_GetKeyFromName
+function resdl_SDL_GetKeyFromName() {
+    return "keyFromname";
+}
+
+// Provides: resdl_SDL_GetCompiledVersion
+function resdl_SDL_GetCompiledVersion() {
+    return "resdl_SDL_GetCompiledVersion: Not implemented";
```
lessp, line 658: 🤷

src/reason-sdl2/sdl2_stubs.js
lessp: https://wiki.libsdl.org/SDL_Scancode

src/reason-sdl2/sdl2_stubs.js
```diff
@@ -612,3 +612,60 @@ function resdl_SDL_WaitTimeoutEvent() {
...
+}
+
+// Provides: resdl_SDL_GetScancodeFromName
+function resdl_SDL_GetScancodeFromName() {
+    return "scancodeFromName";
+}
+
+// Provides: resdl_SDL_GetKeyFromName
+function resdl_SDL_GetKeyFromName() {
+    return "keyFromname";
```
lessp, line 653: https://wiki.libsdl.org/SDL_GetKeyFromName
https://wiki.libsdl.org/SDL_Scancode

src/reason-sdl2/sdl2_stubs.js
glennsl: I wouldn't worry too much about getting this exactly right as I think the dummy stubs approach is fundamentally broken. There ought to be separate platform APIs for desktop, JS and mobile. More on that soon, but for now I'd suggest just going with whatever works short-term.

PR #929, opened 2020-06-30
chore(ci): Upgrade Ubuntu to 18.04

zbaylin:


zbaylin:
@bryphe I think you might need to manually delete the build artifacts from 16.04. It looks like they're causing symbol/linking errors on the 18.04

.ci/restore-build-cache.yml
```diff
@@ -11,7 +11,7 @@ steps:
         branchName: 'refs/heads/master'
         buildVersionToDownload: 'latestFromBranch'
         downloadType: 'single'
-        artifactName: 'cache-$(Agent.OS)-install'
+        artifactName: 'cache-$(Agent.OS)_$(Agent.MachineName)-install'
```
bryphe, line 14: Ah, unfortunately this won't work, I don't think - the machine name is different every run, so we'd lose our cache.

But what we could do is temporarily disable the cache in `azure-pipelines.yml`, and then re-enable it on `master`.

zbaylin: Sounds good, I dont have the ability to push directly to master, so let me know when you want me to merge this so we can get the cache back!

PR #931, opened 2020-06-30
Fix sliders in examples

SirFurness:
This makes the sliders in the examples actually slideable.

### DIFF
--- a/examples/SpringExample.re
+++ b/examples/SpringExample.re
@@ -65,9 +65,13 @@ module Example = {
       ];
   };
   let%component make = () => {
-    let%hook (stiffness, setStiffness) = Hooks.state(160.);
-    let%hook (damping, setDamping) = Hooks.state(10.);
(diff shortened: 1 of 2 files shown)


examples/SpringExample.re
```diff
@@ -65,9 +65,12 @@ module Example = {
       ];
   };
   let%component make = () => {
-    let%hook (stiffness, setStiffness) = Hooks.state(160.);
-    let%hook (damping, setDamping) = Hooks.state(10.);
-    let%hook (targetPosition, setTargetPosition) = Hooks.state(256.0);
+    let%hook (stiffness, setStiffness) =
+      Hooks.reducer(~initialState=160., (value, _) => value);
```
bryphe, line 69: I wonder if we should just revamp our `Hooks.state` to essentially emulate this (remove the boilerplate of the `(value, _) => value`), in order to make it more reliable.

PR #932, opened 2020-06-30
feat(font): Add OpenType font features

zbaylin:
This will integrate OpenType's font features with Harfbuzz.

Sourced from here:
https://harfbuzz.github.io/shaping-opentype-features.html


src/reason-harfbuzz/harfbuzz.cpp
```diff
@@ -104,9 +104,20 @@ static value createShapeTuple(unsigned int codepoint, unsigned int cluster) {
...
 
-CAMLprim value rehb_shape(value vFace, value vString) {
-  CAMLparam2(vFace, vString);
-  CAMLlocal1(ret);
+CAMLprim value rehb_shape(value vFace, value vString, value vFeatures) {
+  CAMLparam3(vFace, vString, vFeatures);
+  CAMLlocal2(ret, feat);
+
+  int featuresLen = Wosize_val(vFeatures);
+  hb_feature_t *features = (hb_feature_t *)malloc(featuresLen * sizeof(hb_feature_t));
```
bryphe, line 112: Does this leak? It seems like we'd need to `free` it. Might be worth adding to `HarfbuzzCli` so we can ensure it passes ASAN.

src/Font/Feature.rei
```diff
@@ -0,0 +1,247 @@
+type t;
+
+type position = [ | `Start | `End | `Position(int)];
+
+type feature =
```
bryphe, line 5: Wow! Thanks for creating this variant covering the OpenType features

src/Font/Feature.re
```diff
@@ -0,0 +1,489 @@
...
+  | AlternateVerticalHalfMetrics
+  | VowelJamoForms
+  | VerticalKanaAlternates
+  | VerticalKerning
+  | ProportionalAlternateVerticalMetrics
+  | VerticalAlternatesAndRotation
+  | VerticalAlternatesForRotation
+  | SlashedZero;
+
+let getTag = feature =>
```
bryphe, line 245: Nice! I guess we may also want the reverse for Onivim, such that the font features can be set, like:
"editor.fontLigatures": "'ss02', 'ss9'"

Or they could just be passed directly, maybe?

src/Font/Feature.rei
```diff
@@ -0,0 +1,247 @@
...
+  | VowelJamoForms
+  | VerticalKanaAlternates
+  | VerticalKerning
+  | ProportionalAlternateVerticalMetrics
+  | VerticalAlternatesAndRotation
+  | VerticalAlternatesForRotation
+  | SlashedZero;
+
+let make:
+  (~feature: feature, ~value: int, ~start: position, ~end': position) => t;
```
bryphe, line 247: Is the `end'` because `end` is a keyword? We've been using `stop` instead.

src/Font/FontCache.re
```diff
@@ -115,16 +122,16 @@ let getMetrics: (t, float) => FontMetrics.t =
...
 
-let shape: (t, string) => ShapeResult.t =
-  ({hbFace, shapeCache, _}, str) => {
-    switch (ShapeResultLruHash.find(str, shapeCache)) {
+let shape: (~features: list(Feature.t)=?, t, string) => ShapeResult.t =
+  (~features=[], {hbFace, shapeCache, _}, str) => {
+    switch (ShapeResultLruHash.find((str, features), shapeCache)) {
     | Some(v) =>
-      ShapeResultLruHash.promote(str, shapeCache);
+      ShapeResultLruHash.promote((str, features), shapeCache);
```
bryphe, line 129: An alternate API design could be to 'bake-in' the features with the font type itself (ie, side-by-side with the ({skiaFace, harfbuzzFace, shapingFeatures, }) etc) - then the `shape` would have this  implicitly as part of the font:

```
	let shape: (t, string) => ShapeResult.t =
  ({hbFace, shapeCache, features _}, str) => {	 
...
});
```

It might mean that the `shape` operation in the cached case is more performant, since the equality and hash would be simpler than working with the nested type. Something to think about if this shows up in our profiling.

src/Font/Feature.rei
zbaylin: Oh cool! Thats much better than `end'`. Thanks!

src/Font/FontCache.re
zbaylin: Oh that's an interesting idea! Something I'll investigate!

src/Font/Feature.re
glennsl: Unless these are going to be pattern matched, it would be much more efficient to represent them as strings directly. Something like:

```ml
module Feature: {
  type t;

  let accessAllAlternates: t;
  ...
} = {
  type t = string;

  let access AllAlternates = "aalt";
  ...
};

glennsl: That would also make it more open to extension, for example with a `let custom: string => t;`

src/reason-harfbuzz/harfbuzz.cpp
bryphe: Looks like the latest runs are showing a leak - this is probably it.

src/Font/FontCache.re
zbaylin: Looking into this a little further I'm realizing that this would decrease our cache hits when loading the same font with different features. Since each font with distinct features would take up it's own cache entry, there would be a lot of duplication since we would be creating the Harfbuzz face from the Skia face multiple times. Performance is definitely a concern for having the tuple with the features and the string -- I'll try to write up some benchmarks to compare the two.

src/reason-harfbuzz/harfbuzz.cpp
zbaylin: Thanks for pointing that out! Fixed in a2cf61e!

src/Font/FontCache.re
zbaylin: Looks like the current benchmarks already test this for us:

With the tuple:
https://dev.azure.com/revery-ui/revery/_build/results?buildId=6499&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=2b629420-0742-5d3a-52a2-2c62bb0704f0&l=35

Master (without tuple): https://dev.azure.com/revery-ui/revery/_build/results?buildId=6500&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=2b629420-0742-5d3a-52a2-2c62bb0704f0&l=35

Funnily enough the master benchmark is actually slightly slower -- probably just random noise, but it also shows how the tuple doesn't slow cache down very much, if at all.

bryphe: Perfect! Thanks for investigating this. Since there is no perf impact, I'm fine with how it is 👍

zbaylin:
Thanks for the approval, @bryphe! For posterity sake, anyone who is interested in playing around with the features should check out https://wakamaifondue.com/. It shows all the features any font file you have has, with examples through CSS as well. Helped me figure out how to write tests/the CLI for this, which I'm finishing up now!

PR #934, opened 2020-07-01
fix(harfbuzz/#933): Invalid glyph counts w/ Ligatures

zbaylin:
@bryphe caught that we were incorrectly calculating the length of the returned buffer. I fixed the call order and the glyph_count.

The string "fi" now correctly only returns one glyph in Roboto Mono, and the other strings work as they did before.

### DIFF
--- a/src/reason-harfbuzz/harfbuzz.cpp
+++ b/src/reason-harfbuzz/harfbuzz.cpp
@@ -115,13 +115,14 @@ CAMLprim value rehb_shape(value vFace, value vString) {
   hb_buffer_add_utf8(hb_buffer, String_val(vString), -1, 0, -1);
   hb_buffer_guess_segment_properties(hb_buffer);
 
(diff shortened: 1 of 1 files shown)


src/reason-harfbuzz/harfbuzz.cpp
```diff
@@ -115,13 +115,14 @@ CAMLprim value rehb_shape(value vFace, value vString) {
   hb_buffer_add_utf8(hb_buffer, String_val(vString), -1, 0, -1);
   hb_buffer_guess_segment_properties(hb_buffer);
 
-  unsigned int len = hb_buffer_get_length(hb_buffer);
-  hb_glyph_info_t *info = hb_buffer_get_glyph_infos(hb_buffer, NULL);
+  hb_shape(hb_font, hb_buffer, NULL, 0);
```
bryphe, line 118: Thanks for fixing this up!

PR #937, opened 2020-07-02
refactor(font): deprecate reason-font-manager in favor of Skia

### LINKED ISSUE #894: Font families: multi-variant font files are ignored
This is a bug with the new font families implementation in #879. If a font contains more than one variant (as is supported in newer versions of OpenType and TrueType), we only render the normal variant [1].

It's not really a bug with my implementation, but rather an oversight in the code that's existed for a while and only really became evident with this. 

(Mostly) @bryphe and I did some investigation and found that, in `reason-skia`, we hardcode 0 as the index for all fonts, as seen here:
https://github.com/revery-ui/revery/blob/a525e7f07d1417791c6ac2b574ed7fb9e68ceaeb/src/Font/FontCache.re#L68

The documentation on that function is sort of iffy, but I think it refers to a `VariationIndex` in OpenType fonts, at least:
https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#device-and-variationindex-tables

I don't think `reason-font-manager` gives us the index, so we might have to either patch it in to do that, or ditch it entirely.

[1]: This is only kind of true: it's whatever variant is at index 0, since we hardcode that.

zbaylin:
This PR moves to a purely Skia-based font-discovery and management API.

This fixes #894 and will help when we add font-fallback. @EduardoRFS also noted on Discord that it will help on mobile platforms that dont have OS-based font discovery primitives baked in.


zbaylin:
Looks like we don't get any performance loss with this PR, which is always good :)

This PR:
```
+----------------------------------------------------------------+-------------+---------------------+
|  Draw: drawText                                                |  10000      |  0.0862729549408    |
+----------------------------------------------------------------+-------------+---------------------+
```


Master:
```
+----------------------------------------------------------------+-------------+---------------------+
|  Draw: drawText                                                |  10000      |  0.0877161026001    |
+----------------------------------------------------------------+-------------+---------------------+
```

src/Font/FontCache.re
```diff
@@ -8,6 +8,12 @@ module StringHashable = {
   let hash = Hashtbl.hash;
 };
 
+module SkiaTypefaceHashable = {
+  type t = option(Skia.Typeface.t);
+  let equal = (===);
```
zbaylin, line 13: I made a `Typeface.equals` function to replace this, but it turned out to be markedly slower than this. There is a `UniqueID` method in the C++ Skia lib, but it doesnt look like there are C bindings for it. Not sure if we can make a custom C++ integration file to get that functionality, but this seems to work "good enough" for now. Let me know if you have any suggestions!

test/TestRunner.re
```diff
@@ -10,5 +10,6 @@ if (Sys.win32) {
 Revery_Core_Test.TestFramework.cli();
 Revery_Math_Test.TestFramework.cli();
 Revery_UI_Test.TestFramework.cli();
+Revery_Font_Test.TestFramework.cli();
```
bryphe, line 13: Thanks for adding a test suite!

bench/lib/DrawBench.re
```diff
@@ -25,7 +25,10 @@ module Data = {
 };
 
 let drawText = canvasContext => {
-  switch (Revery.Font.load("TestFont.ttf")) {
+  let maybeSkia =
```
bryphe, line 28: Appreciate you checking the benchmark results to ensure there isn't a regression - text rendering is definitely on the hot path 👍

src/Font/FontCache.re
bryphe: It sounds like the ideal would be to add the `UniqueID` method - it should be relatively straightforward to add here: https://github.com/revery-ui/esy-skia/blob/c8bded8894f6624a0c1b0b1b2736630754bef62b/src/c/sk_typeface.cpp#L77

Although the referential equality (`===`) I _think_ should work OK with the skia pointer.

src/Font/FontFamily.re
```diff
@@ -34,28 +25,82 @@ module FontFamilyCache = Lru.M.Make(FontFamilyHashable, FontDescriptorWeight);
...
   | Linux => system("Liberation Sans")
+  | Mac => system("System Font")
   | _ => system("Arial")
   };
 
-let resolve = (~italic=false, ~mono=false, weight, solver) =>
-  solver(~italic, ~mono, weight) |> FontCache.load;
+let defaultMono =
+  switch (Revery_Core.Environment.os) {
+  | Mac => system("Menlo")
```
bryphe, line 91: Thanks for looking at default fonts for each platform 👍

test/Font/FamilyTest.re
```diff
@@ -0,0 +1,39 @@
...
+  (maybeSkia, maybeName);
+};
+
+describe("Family", ({test, _}) => {
+  test("fromFile", ({expect, _}) => {
+    let family = Family.fromFile("Inconsolata.otf");
+    let (maybeSkia, maybeName) = getInfoFromSkia(family);
+
+    expect.option(maybeSkia).toBeSome();
+    expect.option(maybeName).toBe(Some("Inconsolata"));
```
bryphe, line 19: Cool! Nice to have this scaffolding in place.

src/Font/FontFamily.re
```diff
@@ -34,28 +25,82 @@ module FontFamilyCache = Lru.M.Make(FontFamilyHashable, FontDescriptorWeight);
...
+  | Linux => system("Liberation Serif")
+  | _ => system("Times New Roman")
+  };
+
+let resolve = (~italic=false, weight, solver) =>
+  solver(~italic, weight) |> FontCache.load;
 
-let toPath = (~italic=false, ~mono=false, weight, solver) =>
-  solver(~italic, ~mono, weight);
+let toMaybeSkia = (~italic=false, weight, solver) => solver(~italic, weight);
```
bryphe, line 106: Minor: We haven't really been including `Maybe` in function names when they return an `option` (just for variable names)

bryphe:
With this change - it seems like we don't really need `reason-font-manager` here anymore. Were you planning on removing it in a separate change?

src/Font/FontFamily.re
zbaylin: Ah okay, I can change that. Thanks!

PR #938, opened 2020-07-03
feat(native): add Locale module

zbaylin:
This is to get the current user (and maybe eventually system) locale information.


src/Native/locale_cocoa.c
```diff
@@ -0,0 +1,10 @@
+#ifdef __APPLE__
+#include <stdio.h>
+
+#import <Cocoa/Cocoa.h>
+
+char *revery_getUserLocale_cocoa() {
+    NSLocale *nsLocale = [NSLocale currentLocale];
+    return (char *)[[nsLocale localeIdentifier] UTF8String];
```
bryphe, line 8: Thanks for researching this! Just curious - does `setlocale` not work on OSX as expected?

src/Native/locale.c
```diff
@@ -0,0 +1,29 @@
...
+
+CAMLprim value revery_getUserLocale() {
+    CAMLparam0();
+    char *ret;
+#ifdef __APPLE__
+    ret = revery_getUserLocale_cocoa();
+#else
+    ret = setlocale(LC_CTYPE, NULL);
+#endif
+    CAMLreturn(caml_copy_string(ret));
```
bryphe, line 28: Just to be safe - it might be worth wrapping this in a value, like:
```
CAMLparam0();
CAMLlocal1(localeValue);
...
localeValue = caml_copy_string(ret);
CAMLreturn(localeValue)
```

It _might_ be OK to not use `CAMLlocal` in this case - but I never like to take changes with the FFI!

src/Core/App.re
```diff
@@ -255,6 +255,10 @@ let start = init => {
 
   Revery_Native.initApp();
 
+  AppLog.infof(m =>
+    m("Operating in locale : %s", Revery_Native.Locale.getUser())
```
bryphe, line 259: Good idea to add a log for this

src/Native/locale_cocoa.c
zbaylin: Yeah unfortunately it always returns the default POSIX locale `C`. On windows it doesn't work correctly as well from what I can tell.

src/Native/locale.c
zbaylin: Sounds good, I added this in 1f0efe3!

src/Native/locale_win32.c
```diff
@@ -0,0 +1,18 @@
+#ifdef _WIN32
+
+#include "win32_target.h"
+
+#include <windows.h>
+#include <stdlib.h>
+
+char *revery_getUserLocale_win32() {
+    WCHAR locale[LOCALE_NAME_MAX_LENGTH];
+    char *localeStr = malloc(LOCALE_NAME_MAX_LENGTH * sizeof(char));
```
bryphe, line 10: Hmm, where does this get freed?

Unfortunately we don't have ASAN run on Windows - so it might not catch a leak here.

src/Core/Environment.rei
```diff
@@ -54,3 +54,10 @@ type os =
   | Unknown;
 
 let os: os;
+
+/**
+[getUserLocale] returns the current user locale. Note that on some platforms
+(including macOS) the locale can change during runtime
+*/
+let getUserLocale: unit => string;
```
bryphe, line 62: Would you mind adding a simple test case for this? Just something that calls and verifies the string is not empty - basically just to ensure it doesn't crash.

src/Native/locale_win32.c
zbaylin: Good point, I originally thought the garbage collector would free it but forgot that we copy the string. I'll fix that now.

src/Core/Environment.rei
zbaylin: Yeah sure! I'll do that now.

src/Native/locale.c
```diff
@@ -23,12 +23,15 @@ CAMLprim value revery_getUserLocale() {
...
+    camlRet = caml_copy_string(ret);
 #elif WIN32
     ret = revery_getUserLocale_win32();
+    camlRet = caml_copy_string(ret);
 #else
     setlocale(LC_CTYPE, "");
     ret = setlocale(LC_CTYPE, NULL);
-#endif
     camlRet = caml_copy_string(ret);
+    free(ret);
```
bryphe, line 34: This doesn't look like it's hitting the Windows path - just linux.

I guess we should check:
1) Does `ret` need to be freed for Apple?
2) Does `ret` need to be freed for Windows? I think the answer is yes
3) Does `ret` need to be freed for Linux?

If the answer is to all of those, we could `free(ret)` after the `#ifdef`/`#endif` block. Otherwise, if we just need to free some of them - we'd need to include it in the platform condition

zbaylin: Yeah that was my bad. I fixed it in eedb70b. On Linux and macOS we dont need to free (in fact we segfault if we do, predictably). I added a `shouldFree` variable that I think makes it a bit clearer. Let me know if you think this is better!

bryphe: Looks good to me! Thanks @zbaylin

PR #939, opened 2020-07-03
feat(reason-skia): bind to FontManager matchFamilyCharacterStyle

zbaylin:
This does some non-traditional Caml operations like shifting/or-ing bits. I think I got it all right, but if not, please let me know! I'll add some examples to SkiaFontManagerCli for it.


src/reason-skia/Skia.rei
```diff
@@ -66,6 +66,8 @@ module FontManager: {
 
   let makeDefault: unit => t;
   let matchFamilyStyle: (t, string, FontStyle.t) => option(Typeface.t);
+  let matchFamilyStyleCharacter:
```
bryphe, line 69: Per our discussion - the last argument should be a `Uchar.t`: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Uchar.html - that helps protect us, via the type system, from all sorts of invalid strings (and puts the burden on the consumer to get the `Uchar.t` - but often we'll have it anyway)

bryphe: (Will save us that bit-shifting work too!)

zbaylin: Yep sounds good! Once Zed gets merged in that will simplify this a lot. At least the pipeline from Revery to Skia.

examples/skia-font-manager-cli/SkiaFontManagerCli.re
```diff
@@ -95,6 +95,22 @@ let draw = canvas => {
...
+  let emoji = "😃";
+  let char = Zed_utf8.unsafe_extract(emoji, 0);
+  let maybeTypeface =
+    FontManager.matchFamilyStyleCharacter(
+      fontManager,
+      "Arial",
+      style,
+      ["en_US"],
+      char,
+    );
```
bryphe, line 108: Awesome!

Does this work on all platforms? If so - we might want to add a test case for this, too!

src/reason-skia/Skia.rei
bryphe: Perfect!

examples/skia-font-manager-cli/SkiaFontManagerCli.re
zbaylin: I'm actually not sure! We'll see when CI is finished, I guess!

zbaylin: Actually it looks like it does! The hard part about adding tests for it is Linux -- CI gives DejaVu Sans, but Im not sure what other distros will return. What do you think is the best course of action?

PR #941, opened 2020-07-05
chore(monorepo): bundle in Zed

zbaylin:
This bundles [ocaml-community/zed](https://github.com/ocaml-community/zed)'s Utf8 module. Oni already bundles this in, and we will very much benefit from some of the helper functions when implementing font-fallback. Once this is included here we can use this version in Oni


bryphe:
Code change looks good, thanks for bringing this in, @zbaylin !

Can you please add `zed_utf8` to our ThirdPartyLicenses.txt for revery, like we do here: https://github.com/onivim/oni2/blob/baa315dff6c9ac0384adcd50d9bea23d8517ae93/ThirdPartyLicenses.txt#L18 ?

zbaylin:
> Prior to merging, can you please add `zed_utf8` to our ThirdPartyLicenses.txt for revery, like we do here: https://github.com/onivim/oni2/blob/baa315dff6c9ac0384adcd50d9bea23d8517ae93/ThirdPartyLicenses.txt#L18 ?

Sure thing!

PR #943, opened 2020-07-07
feat: Android

EduardoRFS:
There is quite some duplicated code happening, related to system detection, probably in the near future I would like to change how we do that.

## Changes

- Get the OS during runtime using the value from `config.h`
- Add Android specific flags to reason-sdl2 and reason-skia
- Add special handle for Android scale factor

## How to test

Currently esy doesn't provide the needed tools, so we need some hacks, they're provided at [reason-mobile](https://github.com/EduardoRFS/reason-mobile)

### DIFF
--- a/bench.json
+++ b/bench.json
@@ -4,7 +4,7 @@
       "run": "esy '@bench' x ReveryBench"
   },
   "override": {
-      "build": ["dune build -p font-manager,harfbuzz,skia,sdl2,Revery,ReveryBench -j4"],
+      "build": ["dune build -p harfbuzz,skia,sdl2,Revery,ReveryBench -j4"],
       "install": [
           "esy-installer ReveryBench.install"
       ]
--- a/src/Core/Environment.re
+++ b/src/Core/Environment.re
@@ -27,7 +27,7 @@ let os = {
   webGL
     ? Browser
     : (
-      switch (Revery_Native.Environemnt.get_os()) {
+      switch (Revery_Native.Environment.get_os()) {
       | `Android => Android
       | `Linux => Linux
       | `Mac => Mac
--- a/src/Core/Environment.re
+++ b/src/Core/Environment.re
@@ -16,9 +16,10 @@ let sleep = (t: Time.t) =>
 external yield: unit => unit = "caml_thread_yield";
 
 type os =
-  | Windows
-  | Mac
+  | Android
   | Linux
+  | Mac
+  | Windows
   | Browser
   | Unknown;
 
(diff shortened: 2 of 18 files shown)


src/Core/Environment.re
```diff
@@ -16,27 +16,23 @@ let sleep = (t: Time.t) =>
...
-      | _ =>
-        let ic = Unix.open_process_in("uname");
-        let uname = input_line(ic);
-        let _ = close_in(ic);
-        switch (uname) {
-        | "Darwin" => Mac
-        | "Linux" => Linux
-        | _ => Unknown
-        };
+      switch (Revery_Native.Environemnt.get_os()) {
```
lessp, line 30:
```suggestion
      switch (Revery_Native.Environment.get_os()) {
```

src/Native/Revery_Native.re
```diff
@@ -1,4 +1,5 @@
 module Dialog = Dialog;
+module Environemnt = Environment;
```
lessp, line 2:
```suggestion
module Environment = Environment;
```

bryphe:
> There is quite some duplicated code happening, related to system detection, probably in the near future I would like to change how we do that.

Sounds good! Getting the end-to-end compilation on Android working is a tremendous step forward.

bench.json
```diff
@@ -4,7 +4,7 @@
       "run": "esy '@bench' x ReveryBench"
   },
   "override": {
-      "build": ["dune build -p font-manager,harfbuzz,skia,sdl2,Revery,ReveryBench -j4"],
+      "build": ["dune build -p harfbuzz,skia,sdl2,Revery,ReveryBench -j4"],
```
bryphe, line 7: Ah, I just ran into this too trying to bring over latest Revery to Onivim - thanks for fixing!

src/reason-sdl2/config/discover.re
```diff
@@ -9,7 +10,11 @@ let detect_system_header = {|
   #if __APPLE__
     #define PLATFORM_NAME "mac"
   #elif __linux__
-    #define PLATFORM_NAME "linux"
+    #if __ANDROID__
+      #define PLATFORM_NAME "android"
+    #else
+      #define PLATFORM_NAME "linux"
+    #endif
```
bryphe, line 17: Nice! 🍾  🥂

EduardoRFS:
@bryphe Currently it's not properly documented, especially because the project for integration is not published yet, so you will need to deep dive on how to integrate SDL on Android

But the building process is similar to this, except that currently it's not possible to build revery, because of some weird details, but revery-workshop is possible to build.

https://gist.github.com/EduardoRFS/1686a1d9e913e1d4613b8dae86c7a72f#file-revery-ios-sh-L20

PR #944, opened 2020-07-07
feat(font): add fallback

zbaylin:
This leverages Skia's `matchFamilyStyleCharacter` and Harfbuzz to create fallback for fonts.


src/Draw/CanvasContext.re
```diff
@@ -217,33 +217,3 @@ let clipPath =
     (v: t, ~clipOp: clipOp=Intersect, ~antiAlias=false, path: Skia.Path.t) => {
   Canvas.clipPath(v.canvas, path, clipOp, antiAlias);
 };
-
-module Deprecated = {
```
zbaylin, line 221: This was marked as deprecated and didn't really fit into the new API. I can re-add it if it's necessary but removing it caused no build errors so I assume it's not required.

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+                 [ShapeResult.{hbFace, skiaFace, glyphId, cluster}, ...acc],
+               [],
+             );
+        FallbackLruHash.add(substr, v, fallbackCache);
+        FallbackLruHash.trim(fallbackCache);
+        v;
+      };
+    };
+
+    /* This is a pretty complicated switch block but it is required in order
```
zbaylin, line 229: This one's pretty complex. I tried making it as simple as possible, but if anyone finds any extraneous cases please let me know. I added comments on each pattern/case to indicate the scenario in which it would be used.

src/UI/TextNode.re
```diff
@@ -76,18 +72,29 @@ class textNode (text: string) = {
...
             ascentPx *. (-1.0) +. lineHeightPx *. float_of_int(lineIndex);
 
-          let glyphString =
+          let glyphStrings =
             line
             |> Revery_Font.shape(font)
-            |> Revery_Font.ShapeResult.getGlyphString;
+            |> Revery_Font.ShapeResult.getGlyphStrings;
+
+          let offset = ref(0.);
```
zbaylin, line 80: Also maybe not the best API (probably something semi-recursive is better?), but the principle is required. Now that we can have multiple sections of text per line, we need to keep track of where to place each.

examples/Hello.re
```diff
@@ -127,7 +127,7 @@ let render = () =>
     <Row>
       <AnimatedText delay=Time.zero text="Welcome" />
       <AnimatedText delay={Time.ms(500)} text="to" />
-      <AnimatedText delay={Time.seconds(1)} text="Revery" />
+      <AnimatedText delay={Time.seconds(1)} text="Revery 😃" />
```
bryphe, line 130: Awesome!

examples/skia-font-manager-cli/SkiaFontManagerCli.re
```diff
@@ -96,8 +96,8 @@ let draw = canvas => {
     );
   };
 
-  let emoji = "😃";
-  let char = Zed_utf8.unsafe_extract(emoji, 0);
+  let emoji = "😃👋";
```
bryphe, line 99: Can we toss in a japanese character here somewhere? 

How about `鬼` or `おに`- I found it's Japanese for [`Oni`](https://en.wikipedia.org/wiki/Oni) 😄

zbaylin: Haha sure! Will do!

src/Draw/CanvasContext.re
bryphe: It's used in Onivim 2, so just a heads up that we'll need to convert it there potentially when we merge - but I'm OK removing it 👍

src/Font/FontCache.re
```diff
@@ -21,6 +21,12 @@ module SkiaTypefaceHashable = {
     };
 };
 
+module UcharHashable = {
+  type t = Uchar.t;
+  let equal = Uchar.equal;
+  let hash = Hashtbl.hash;
```
bryphe, line 27: I wonder if it'd be OK / faster to just use `Uchar.to_int`?

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
 
 let getSkiaTypeface: t => Skia.Typeface.t = font => font.skiaFace;
 
+let unresolvedGlyphID = 0;
+
+let matchCharacter = (fallbackCharacterCache, uchar, skiaFace) =>
```
bryphe, line 165: Thanks for adding caching around this!

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+    FallbackCharacterLruHash.promote(uchar, fallbackCharacterCache);
+    maybeTf;
+  | None =>
+    let familyName = skiaFace |> Skia.Typeface.getFamilyName;
+    let maybeTf =
+      Skia.FontManager.matchFamilyStyleCharacter(
+        FontManager.fontManager,
+        familyName,
+        skiaFace |> Skia.Typeface.getFontStyle,
+        [Environment.userLocale],
```
bryphe, line 177: If the user locale changes, it might invalidate the cache

src/Font/FontCache.re
zbaylin: That was my first instinct, I actually cant remember why I didnt do it. I'll change it now!

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+    FallbackCharacterLruHash.add(uchar, maybeTf, fallbackCharacterCache);
+    FallbackCharacterLruHash.trim(fallbackCharacterCache);
+    maybeTf;
+  };
+
+let shaper = (hbFace, skiaFace, fallbackCache, fallbackCharacterCache, str) => {
+  let fallback = (len, (list, last, i), Harfbuzz.{glyphId, cluster}) => {
+    let shapeNode =
+      if (glyphId == unresolvedGlyphID) {
+        let uchar = Zed_utf8.unsafe_extract(str, cluster);
```
bryphe, line 196: I think `unsafe_extract` can throw - so we may want to wrap in an exception handler, so that we can fall-back to an unshaped result or empty glyphs in case there is something busted in the string itself.

zbaylin: I think so too -- sounds good!

src/Font/FontCache.re
zbaylin: Is that always undesirable? I'm not too familiar with how the locale effects the results of the function call, but I can envision times where that's a desired result. For now, the userLocale is constant, so it shouldnt be a huge issue but I can look into this more!

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+        let newTypeface =
+          matchCharacter(fallbackCharacterCache, uchar, skiaFace);
+        let result = load(newTypeface);
+        switch (result) {
+        | Ok({hbFace, skiaFace, _}) =>
+          Error((hbFace, skiaFace, cluster, uchar))
+        | Error(_) => Ok(ShapeResult.{hbFace, skiaFace, glyphId, cluster})
```
bryphe, line 203: I think this mapping of `Ok` -> `Error` and vice-versa is really confusing - we should come up with a clearer representation. At the very least, it'd be better to have something more domain specific - like the `Unshaped`/`Hole`/`Shaped` from before.

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+      )
+    );
+    FallbackCharacterLruHash.add(uchar, maybeTf, fallbackCharacterCache);
+    FallbackCharacterLruHash.trim(fallbackCharacterCache);
+    maybeTf;
+  };
+
+let shaper = (hbFace, skiaFace, fallbackCache, fallbackCharacterCache, str) => {
+  let fallback = (len, (list, last, i), Harfbuzz.{glyphId, cluster}) => {
+    let shapeNode =
```
bryphe, line 194: Can this be extracted to a function outside of `shaper`?

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+        Uchar.to_int(uchar),
+        familyName,
+      )
+    );
+    FallbackCharacterLruHash.add(uchar, maybeTf, fallbackCharacterCache);
+    FallbackCharacterLruHash.trim(fallbackCharacterCache);
+    maybeTf;
+  };
+
+let shaper = (hbFace, skiaFace, fallbackCache, fallbackCharacterCache, str) => {
```
bryphe, line 192: It seems like `shaper` could just take a `t` and be destructured like `({hbFace, skiaFace, ..., _}, str)`

src/Font/FontCache.re
zbaylin: Sounds good to me. I can make a type for it that's more concrete/obvious as to what the purpose is.

src/Font/FontCache.re
bryphe, line 192: The name `shaper` is a noun - usually we've been using verbs for functions (ie, `shape` instead of `shaper`)

src/Font/FontCache.re
zbaylin: Yep! Will do.

src/Font/FontCache.re
zbaylin: Ah okay, makes sense -- I used `shaper` because of the other `shape` function that calls it but I can come up with a better name!

src/Font/FontCache.re
bryphe: The user locale can affect the shaping result, potentially: https://raphlinus.github.io/rust/skribo/text/2019/02/27/text-layout-kickoff.html

image
(Same unicode, different glyphs due to the locale!)

src/Font/FontCache.re
glennsl: Maybe you already know this, but I'll mention it just in case.

`unsafe`  typically doesn't mean it can throw, though it can do that too, it means it makes assumptions which if violated can result in undefined behaviour. That can mean it returns garbage data or that it crashes hard because it's trying to access an invalid memory address, for example. So wrapping it in an exception handler might also be necessary, but it's not sufficient. You have to be absolutely sure that that assumptions hold as well.

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
 
 let getSkiaTypeface: t => Skia.Typeface.t = font => font.skiaFace;
 
+let unresolvedGlyphID = 0;
+
+let matchCharacter = (fallbackCharacterCache, uchar, skiaFace) =>
+  switch (FallbackCharacterLruHash.find(uchar, fallbackCharacterCache)) {
+  | Some(maybeTf) =>
```
glennsl, line 167: Variable names should generally not be abbreviated, unless it's painfully obvious what the abbreviation means. In this case it's not. There's not a single mention of "typeface" so far in this function.

src/Font/FontCache.re
```diff
@@ -47,14 +53,29 @@ module ShapeResultWeighted = {
...
+module SkiaTypefaceWeighted = {
+  type t = option(Skia.Typeface.t);
+  let weight = _ => 1;
+};
+
 module MetricsLruHash = Lru.M.Make(FloatHashable, MetricsWeighted);
 module ShapeResultLruHash = Lru.M.Make(StringHashable, ShapeResultWeighted);
+module FallbackLruHash = Lru.M.Make(StringHashable, FallbackWeighted);
+module FallbackCharacterLruHash =
+  Lru.M.Make(UcharHashable, SkiaTypefaceWeighted);
```
glennsl, line 70: Up to this point, the LRU names have referred to the value type, but these suddenly start referring to the key and their role (I think). This seems very inconsistent and confusing.

More generally I'm wondering why these are named `LruHash`. That seems like leaking technical details. Could these not just be named something like `MetricsCache` or `TypefaceLookupByChar`, describing their role instead of the technical decisions? Usually you want to isolate technical decisions so you have the freedom to change it in the future without having to refactor the entire codebase, or alternatively have it become a confusing mess.

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+      };
+
+    let shapeFallback = (startIndex, endIndex, skiaFace, hbFace) => {
+      let substr = String.sub(str, startIndex, endIndex - startIndex + 1);
+      switch (FallbackLruHash.find(substr, fallbackCache)) {
+      | Some(v) =>
+        FallbackLruHash.promote(substr, fallbackCache);
+        v;
+      | None =>
+        let v =
```
glennsl, line 216: `v` is an almost completely meaningless name. It should not ever be used, except for something like a completely generic collection where the value can literally be anything, and where you can easily see both where it's used and where it's defined at the same time, practically a one-liner. This is far from meeting either criteria, and so makes the code unnecessarily hard to read for seemingly no other reason than laziness.

I know there's plenty of this going around the codebase already, but we really ought to do better.

src/Font/FontCache.re
```diff
@@ -128,15 +160,181 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+        v;
+      };
+    };
+
+    /* This is a pretty complicated switch block but it is required in order
+       to make this construction as efficient as possible. Each case is
+       described in a comment above itself. */
+    let (newList, newLast) =
+      switch (last, shapeNode) {
+      // There is no hole before this, and we don't have a hole now.
```
glennsl, line 234: I have no idea what `last`, `shapeNode` or a "hole" even is here. I think it would be worthwhile spending a bit more time on this to try to make the code and data structures more descriptive, breaking it up into smaller function with more descriptive names and using dedicated variant types instead of `option` and `result` if possible. And to document the overall algorithm.

src/Font/FontManager.re
```diff
@@ -0,0 +1,5 @@
+/* FontManager.re
+   This module is so that multiple other modules can use the same font
+   manager without creating circular imports */
+
+let fontManager = Skia.FontManager.makeDefault();
```
glennsl, line 5: Minor nit, but this name is pretty redundant. How about calling it `instance` instead?

examples/skia-font-manager-cli/SkiaFontManagerCli.re
```diff
@@ -96,8 +96,8 @@ let draw = canvas => {
...
+  let char = Zed_utf8.unsafe_extract(emoji, String.length(emoji) / 2);
```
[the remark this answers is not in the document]
zbaylin: You're right -- looks like I had missed the `get` function: https://github.com/ocaml-community/zed/blob/710e90d3d8b9476f3936f98c7b700b01c3a194fe/src/zed_utf8.mli#L82

src/Font/FontCache.re
zbaylin: I agree that it's confusing, especially because the font cache is named `FontCache` rather than `FontLruHash` or something. I think changing them all to `[roll]Cache` would be a good idea, but let me know if you think otherwise.

src/Font/FontCache.re
zbaylin: Haha I accept full responsibility for the laziness. Will change that in my refactoring of the API today.

src/Font/FontCache.re
zbaylin: Fixed in 58bc2ee!

examples/skia-font-manager-cli/SkiaFontManagerCli.re
zbaylin: Added in f2a7796!

examples/skia-font-manager-cli/SkiaFontManagerCli.re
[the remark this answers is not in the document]
zbaylin: Changed in f2a7796 !

src/Font/FontCache.re
zbaylin: These were all changed in the refactoring/streamlining I did today. Hopefully you find those to be better!

src/Font/FontManager.re
zbaylin: Changed in 2bb87ae!

src/Font/FontCache.re
zbaylin: I updated all of them to the `Cache` naming scheme in 29dc689.

src/Font/FontCache.re
zbaylin: Fixed in 29dc689!

src/Font/ShapeResult.re
```diff
@@ -1,30 +1,49 @@
...
-  {shapes, glyphString};
+let size = ({glyphStrings, _}) =>
+  glyphStrings
+  |> List.fold_left((acc, (_, str)) => acc + String.length(str), 0);
+
+let ofHarfbuzz = nodes => {
+  let rec loop = (~nodes, ~str, ~maybeTypeface) =>
+    switch (nodes, maybeTypeface) {
+    | ([], Some(typeface)) => [(typeface, str)]
+    | ([], None) => failwith("Invalid")
```
bryphe, line 21: This `failwith` could cause a runtime crash, which we'd like to avoid. 

Can we design the API such that we can react and handle such failures - either by using a `result` type, or if there is a 'null' that makes sense (like an empty glyph that could rendered) - that would be OK!

src/Font/ShapeResult.re
```diff
@@ -1,30 +1,49 @@
...
+    | ([{skiaFace, glyphId, _}, ...rest], Some(skFace)) =>
+      let lowBit = glyphId land 255 |> Char.chr |> String.make(1);
+      let highBit =
+        (glyphId land 255 lsl 8) lsr 8 |> Char.chr |> String.make(1);
```
bryphe, line 25: This `lowBit`/`highBit` logic seems duplicated - can we consolidate that in a helper function?

bryphe: Something like - `glyphIdToString`.

src/Font/ShapeResult.re
```diff
@@ -1,30 +1,49 @@
...
+      if (skFace === skiaFace) {
+        let str = lowBit ++ highBit ++ str;
```
bryphe, line 27: I wonder how the perf of this compares to the previous strategy? String concatenation is pretty expensive. How does it fare for a long string?

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+  };
+};
+
+let rec generateShapes: (t, string) => list(ShapeResult.shapeNode) =
+  ({hbFace, skiaFace, _} as font, str) => {
+    let rec loop =
+            (~font: t, ~shapes: list(Harfbuzz.hb_shape), ~hole: Hole.t) => {
+      switch (shapes) {
+      | [] =>
+        Hole.resolve(
```
bryphe, line 231: This revised algorithm looks much clearer to me than the previous approach - thanks for factoring out the `Hole` module to consolidate the merging/resolution of the hole-runs

src/Font/FontRenderer.re
```diff
@@ -9,16 +9,23 @@ let measure = {
...
-    Skia.Paint.setTypeface(paint, skiaFace);
     Skia.Paint.setTextSize(paint, size);
-    let width = Skia.Paint.measureText(paint, glyphString, None);
+
+    let width =
+      glyphStrings
+      |> List.fold_left(
+           (acc, (skiaFace, str)) => {
+             Skia.Paint.setTypeface(paint, skiaFace);
+             acc +. Skia.Paint.measureText(paint, str, None);
```
bryphe, line 25: Cool! We probably measure for a text run with mixed fonts 🎉

src/Font/ShapeResult.re
zbaylin: Good question. The difference is the recursion makes it harder to know the length up front, so constructing it this way is easier.

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+          loop(~font, ~shapes=tail, ~hole=newHole);
+        } else {
+          Hole.resolve(
+            ~string=str,
+            ~font,
+            ~generateShapes,
+            ~endAt=cluster,
+            hole,
+          )
+          @ [
```
glennsl, line 250: list concatenation can get pretty expensive. For `n @ m`, it's O(n), and since `generateShapes` is recusisve it seems like this might approach O(n^2). I think you can avoid this by passing around an accumulator argument.

src/Font/ShapeResult.re
glennsl: I wonder if accumulating a list of strings and using `String.concat` at the end would be faster. It would avoid allocating a bunch of intermediate strings at least, but of course you'd have to allocate all the cons cells for the list.

If you know the number of strings up front, `String.concat` tends to be faster. If you don't know, using `Buffer` usually is, but that's of course very imperative.

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+      let uchar = Zed_utf8.unsafe_extract(str, startCluster);
+      let maybeFallbackFont =
+        matchCharacter(font.fallbackCharacterCache, uchar, font.skiaFace)
+        |> load;
+
+      switch (maybeFallbackFont) {
+      | Error(_) => []
+      | Ok(fallbackFont) =>
+        let substring = String.sub(str, startCluster, endAt - startCluster);
+        generateShapes(fallbackFont, substring) |> List.rev;
```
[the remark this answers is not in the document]
zbaylin: I think from an API point of view keeping them like this makes a little bit more intuitive sense, if I understand your question. Here the Hole is responsible for solving itself and any holes that might occur _inside of it_. Basically what you're doing is taking a chunk out of the main string and saying "We don't know what to do with this substring, we'll have it process itself.". As I type this I'm realizing it sounds a bit OOP, but I still think that keeping them separate makes a little more sense. Happy to reconsider though!

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+    FallbackCharacterCache.trim(fallbackCharacterCache);
+    maybeTypeface;
+  };
+
+module Hole = {
+  type t =
+    | Empty
+    | StartsAt(int);
+
+  let empty = Empty;
```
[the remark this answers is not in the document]
zbaylin: Yeah originally the type was `Empty | Range(start, finish)` but I realized that the `finish` part was extraneous. I could change it to just be an option, which would make it `maybeHole` and instead of the hole being empty, the hole "wouldn't exist", so to speak. Let me know what you think!

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+          )
+          @ [
+            ShapeResult.{hbFace, skiaFace, glyphId, cluster},
+            ...loop(~font, ~shapes=tail, ~hole=Hole.empty),
+          ];
+        }
+      };
+    };
+
+    let shapes = Harfbuzz.hb_shape(hbFace, str) |> Array.to_list;
```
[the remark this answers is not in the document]
zbaylin: Very true, I'll change that now.

src/Font/FontCache.re
zbaylin: Sounds good! I'm having trouble wrapping my head around how that accumulator would work. Would you mind giving me a small nudge in the right direction? Sorry!

src/Font/FontCache.re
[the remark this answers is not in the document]
zbaylin: Fixed in f4939ca!

src/Font/ShapeResult.re
zbaylin: I changed it to use `String.concat` in 82570a9. Thanks @glennsl!

src/Font/ShapeResult.re
zbaylin: Added in f6dd958! Thanks!

src/Font/ShapeResult.re
zbaylin: Since this would only ever be the result of an improper construction, there would be no glyphs to render. I changed it to just give an empty list, rather than a runtime crash. Thanks for reminding me to change it!

src/Font/FontCache.re
zbaylin: No problem! Thanks for the inspiration!

src/Font/FontCache.re
zbaylin: Thanks @glennsl! I switched to using `get` which has all the checks built in. I also wrapped it in an exception handler to return the NULL character (`'\0'`) on error.

src/Font/FontCache.re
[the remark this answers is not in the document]
glennsl: I see what you mean, but this still seems unnecessarily coupled and complex to me. It's not a big deal, but it does make ti more difficult to add an accumulator, for example, as we'll see below. I might be missing something though.

src/Font/FontCache.re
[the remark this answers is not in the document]
glennsl: Using an `option` instead will allow you to use `Option.map` and such, which is nice. And even if the hole itself is now just an `int`, you can make it private so that you can still only create and manipulate it with the functions in `Hole`. Sounds like a good idea to me, but I don't have the complete picture here.

src/Font/FontCache.re
glennsl: This is made a bit more complicated because you have mutually recursive functions involving higher-order functions, since `generateShapes` is called inside `Hole.resolve` you have to pass the accumulator through both of those, and `loop` instead of having it contained in `generateShapes`, but the principle is the same regardless. When calling one of the mutually recursive functions, instead of adding to the returned list, you add to an accumulator that you pass to it, which is returned when one of them decides there's nothing more to do. This also has the added benefits of making the functions tail recursive and, in this case, reversing the list as you go instead of having to do so afterwards.

A simple example:

```reason
// Without accumulator
let rec f =
  fun
    | [] => []
    | [head, ...tail] => [head * 2, ...g(tail)]

and g =
  fun
    | [] => []
    | [head, ...tail] => [head / 2, ...f(tail)];
    
f([1, 2, 3]); // [2, 1, 6]

// With accumulator
let rec f = acc =>
  fun
    | [] => acc
    | [head, ...tail] => g([head * 2, ...acc], tail)

and g = acc =>
  fun
    | [] => acc
    | [head, ...tail] => f([head / 2, ...acc], tail);
    
f([], [1, 2, 3]); // [6, 1, 2]
```

Does that help?

src/Font/FontCache.re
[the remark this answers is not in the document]
zbaylin: Just changed -- thanks @glennsl!

src/Font/FontCache.re
zbaylin: Awesome, thanks for your help. I switched to an accumulator-based approach in f8435c7!

src/Font/FontCache.re
```diff
@@ -120,25 +150,125 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+        Uchar.to_int(uchar),
+        familyName,
+      )
+    );
+    FallbackCharacterCache.add(uchar, maybeTypeface, fallbackCharacterCache);
+    FallbackCharacterCache.trim(fallbackCharacterCache);
+    maybeTypeface;
+  };
+
+module Hole = {
```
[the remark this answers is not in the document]
zbaylin: Yeah I agree it's a little confusing. I just pushed 48e88aa, which adds a bunch of clarifying comments. Hope those clear it all up!

src/UI/TextNode.re
```diff
@@ -77,18 +73,29 @@ class textNode (text: string) = {
...
+               CanvasContext.drawText(
+                 ~paint=_textPaint,
+                 ~x=offset^,
+                 ~y=baselineY,
+                 ~text=str,
+                 canvas,
+               );
+
+               offset :=
+                 offset^ +. Skia.Paint.measureText(_textPaint, str, None);
```
bryphe, line 96: I suspect this could be expensive - we might also have to duplicate this logic in `Canvas.drawText`. It might be helpful to have a `drawShapedtext` that does this - we'll probably have to duplicate this logic in the editor surface on Onivim, at the very least

bryphe: Not a blocker for now - but we should how this works when we merge into Onivim, and see if it shows up on the profiler.

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+        matchCharacter(font.fallbackCharacterCache, uchar, font.skiaFace)
+        |> load;
+
+      switch (maybeFallbackFont) {
+      | Error(_) =>
+        // Just because we can't find a font for this character doesn't mean
+        // the rest of the hole can't be resolved. Here we insert the "unknown"
+        // glyph and try to resolve the rest of the string.
+        let substring =
+          String.sub(str, startCluster + 1, endAt - startCluster);
```
glennsl, line 242: This seems very inefficient. Although it only applies to holes, it's still `O(n^2)` and doing `n` allocations. It also seems unnecessary to have the hole resolve itself using a complicated mutual recursion scheme instead of resolving each character individually as the unknown glyphs are encountered. Am I missing something?

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+      ) => {
+    switch (maybeHole) {
+    | None => accumulator
+    | Some(startCluster) =>
+      Log.debugf(m =>
+        m("Resolving hole: startCluster : %d, str: %s", startCluster, str)
+      );
+      let uchar =
+        try(Zed_utf8.get(str, startCluster)) {
+        | _ => '\000' |> Uchar.of_char
```
glennsl, line 230: This should be a constant.

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+      };
+
+    let shapes = Harfbuzz.hb_shape(~features, hbFace, str);
+    loop(~font, ~shapes, ~index=0, ~accumulator, ~maybeHole=None);
+  }
+
+and shape: (~features: list(Feature.t)=?, t, string) => ShapeResult.t =
+  (~features=[], {shapeCache, _} as font, str) => {
+    switch (ShapeResultCache.find((str, features), shapeCache)) {
     | Some(v) =>
```
glennsl, line 338: I recognize this is old code, but it takes two seconds to clean up, and it would have taken two seconds to do it right to begin with. There's no good reason to be this lazy...

```suggestion
    | Some(result) =>
```

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
-      ShapeResultLruHash.promote((str, features), shapeCache);
+      ShapeResultCache.promote((str, features), shapeCache);
       v;
```
glennsl, line 340:
```suggestion
      result;
```

src/Font/ShapeResult.re
```diff
@@ -1,30 +1,56 @@
...
-
-  let i = ref(0);
-
-  while (i^ < len) {
-    let idx = i^;
-    let {glyphId, _}: Harfbuzz.hb_shape = shapes[idx];
+type t = {
+  shapes: list(shapeNode),
+  glyphStrings: list((Skia.Typeface.t, string)),
+};
```
glennsl, line 11: Also technically outside the scope of this PR, but I find `ShapeResult` to be a very vague concept. What's the point of this type? Why do we need to keep `shapes` around? I can't see that it's used anywhere.

src/Font/FontCache.re
glennsl: Oh, I think I see, it's to reduce the number of calls to `Harfbuzz.hb_shape`. That's not very obvious from the code. Does it really have that much overhead?

I'm having a go at refactoring this myself, to either get a better sense of the subtleties, or to end up with an example to show you what I mean.

zbaylin: The problem with resolving each character individually is often times you will get code points that modify next or previous code points. For instance, the US flag emoji is the two characters [U][S] where [x] symbolizes "regional indicator x" (Norway is predictably [N][O]!). The regional indicators are visible on their own (they look like 🇦), but they're meant to be rendered with each other in the event that the font supports it.

The other example is the skin tone modifiers. 👨🏻 is the combination of both 👨 and 🏻.

There are other cases other than emojis, like Hebrew and Arabic. I'm more familiar with the Hebrew case, where vowels are positioned around different parts of the previous character, if they're rendered at all. Furthermore there are certain characters that make multiple sounds based on certain modifiers:

ש - "s" sound
שׁ - "sh" sound (whatever the default font on my Mac is for Github renders this horribly, but if you zoom in you can see a dot over the right-most part of the letter).

These two characters are the same in the alphabet, yet make different sounds and can have different representations -- very confusing.

Not to go too much into Hebrew (it's really the only example of a non-Latin script I know), but some letters are rendered different if they're at the end of a word, i.e. מ and ם -- both these letters are the letter "mem" but the second one is a "mem sofit" and only is ever used at the end of a word. From what I can tell, Harfbuzz often figures all of this out contextually, which goes to show you how much overhead that `hb_shape` call can involve.

If you look at my first implementation (latest commit is 9f6488a before I changed it), I was actually doing the resolution on a per-character basis. However that involved all the pitfalls described above.

I should also note that if you try to test some of the examples I provided (i.e. a flag) you might get some improper clustering -- this is because I pulled out the Harfbuzz upgrade into a separate PR. If you use our version, `@revery/esy-harfbuzz 2.6.8000` the clustering for the glyphs should be fixed.

This was a lot longer than I anticipated, but hopefully it gives some context as to some of the design decisions 😄

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
+  // The end is sent to the `resolve` function
+  type t = option(int);
+
+  // If we have a hole, don't change the start. However,
+  // if we don't have a hole, we need to start it here.
+  let extend = (~cluster, hole) =>
+    switch (hole) {
+    | None => Some(cluster)
+    | Some(_) => hole
+    };
```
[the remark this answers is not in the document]
zbaylin: I know there's a lot of muddying since `Hole` is fully visible to the rest of the file, but I was operating as if `Hole.t` was an abstract type originally (that's what the redundant `let empty = Empty;` was for). What do you think a better name for this would be? I think `make` doesn't fit well either since you're not always creating anything new. I know you suggested a much larger scale refactoring, so maybe this should just be part of that.

src/Font/FontCache.re
bryphe: > There's no good reason to be this lazy..

Watch the tone here, please. I expect communication here to be respectful. This doesn't add anything to the technical conversation.

src/Font/Discovery.re
```diff
@@ -2,11 +2,10 @@
     Discovery.re
 
 */
+open FontManager;
 open Skia;
```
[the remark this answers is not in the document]
zbaylin: Good point! Fixed in d43682c!

src/Font/FontCache.re
```diff
@@ -48,15 +54,30 @@ module ShapeResultWeighted = {
...
   skiaFace: Skia.Typeface.t,
-  metricsCache: MetricsLruHash.t,
-  shapeCache: ShapeResultLruHash.t,
+  metricsCache: MetricsCache.t,
+  shapeCache: ShapeResultCache.t,
+  fallbackCache: FallbackCache.t,
+  fallbackCharacterCache: FallbackCharacterCache.t,
 };
 
 type _t = t;
```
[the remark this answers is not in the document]
zbaylin: Totally agree -- I fixed it in b46a55b!

src/Font/FontCache.re
```diff
@@ -122,25 +152,198 @@ let getMetrics: (t, float) => FontMetrics.t =
...
       ret;
     };
   };
 
 let getSkiaTypeface: t => Skia.Typeface.t = font => font.skiaFace;
 
-let shape: (~features: list(Feature.t)=?, t, string) => ShapeResult.t =
-  (~features=[], {hbFace, shapeCache, _}, str) => {
-    switch (ShapeResultLruHash.find((str, features), shapeCache)) {
+let unresolvedGlyphID = 0;
```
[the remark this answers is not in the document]
zbaylin: Added in 8490511! Thanks!

src/Font/FontCache.re
zbaylin: Also fixed in 8490511 👍

src/Font/FontCache.re
bryphe: That's a good point about `String.sub` allocating - the harfbuzz APIs actually let us specify offsets:
```
  hb_buffer = hb_buffer_create();
  hb_buffer_add_utf8(hb_buffer, String_val(vString), -1, 0, -1);
  hb_buffer_guess_segment_properties(hb_buffer);
```
https://github.com/revery-ui/revery/blob/1808801a74263429f02b8f0abfb3275a15852c7d/packages/reason-harfbuzz/src/harfbuzz.cpp#L147

The last two parameters are `item_offset` and `item_length`: https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-add-utf8 - so we could potentially re-use the same string, but just give it different offsets/length for the substring. Would likely be more efficient

We could change our harfbuzz shape binding to take the offset and length parameters

src/Font/ShapeResult.re
bryphe: For shaping - there are really two outputs - the list of glyphs the text is shaped to, as well as the position offsets for each glyph. The glyphs probably aren't necessary, since they just get transformed to `glyphStrings'. 

The positions would make sense to keep around... but we aren't using them today. So probably makes sense to remove `shapes` for now, and we incorporate `positions` eventually.

src/Font/FontCache.re
```diff
@@ -48,15 +54,30 @@ module ShapeResultWeighted = {
...
-module ShapeResultLruHash =
+module FallbackWeighted = {
+  type t = list(ShapeResult.shapeNode);
+  let weight = _ => 1;
+};
+
+module SkiaTypefaceWeighted = {
+  type t = option(Skia.Typeface.t);
+  let weight = _ => 1;
+};
```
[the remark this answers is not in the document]
zbaylin: Certainly! Are you suggesting a change in the LRU library or something more local here?

src/Font/FontCache.re
glennsl: Thanks for the detailed explanation @zbaylin! That's very helpful and makes a lot of sense! Let me think about this for a bit to come up with something concrete.

src/Font/FontCache.re
glennsl: And I expect code to be written with respect for its audience, yet here we are... Anyway, point taken, but I hope the irony isn't lost.

src/Font/FontCache.re
zbaylin: No problem @glennsl! I've learned a ton about text rendering the past couple weeks, something I want to maybe formalize into a blog post or something when this is all finished up 👍

src/Font/FontCache.re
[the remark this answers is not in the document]
glennsl: It seems generally useful, and that it would be a valuable addition to the library, but I'm not entirely sure.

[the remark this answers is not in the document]
zbaylin: Sounds good, I can try to patch it in to the library. I know right now we use @bryphe's fork of `lru` -- maybe we can move that fork to this org? Either way works.

src/Font/ShapeResult.re
zbaylin: Thanks @bryphe and @glennsl! Removed the shapes field in be9ed81!

src/Font/FontCache.re
zbaylin: Thanks @glennsl -- fixed in 9c5e227!

src/Font/FontCache.re
zbaylin: Thanks @glennsl! Fixed in 9c5e227!

zbaylin:
@glennsl looks like you were able to maintain the clustering with Harfbuzz 2.6.8!

HB 2.6.8
image

HB 1.9.0 (current)
image

zbaylin:
Looks like this is (finally) done! This is by far one of the biggest PRs I've written -- I couldn't have done it without the help of @glennsl and @bryphe! Thanks guys!

bryphe:
Congrats on getting this in, @zbaylin ! Huge step forward for Revery 🎉

PR #945, opened 2020-07-08
feat(native): support iOS

EduardoRFS:
Again, more system detection code, but this time is done.

## Changes 

- Add iOS specific flags to revery, reason-sdl2 and reason-skia
- proper detection of framebufferId
- the application lives inside an SDL_main

This last change is required as our application needs to live inside the [UIApplicationMain](https://github.com/SDL-mirror/SDL/blob/2cf2cfd5b0b0e634bba32295d7ba7e64b298aea3/src/video/uikit/SDL_uikitappdelegate.m#L70), this is essentially a no-op on other platforms.

## How to test

Same as #943 will publish a guide and a CI today

### DIFF
--- a/src/Draw/CanvasContext.re
+++ b/src/Draw/CanvasContext.re
@@ -44,9 +44,11 @@ let create = (window: Revery_Core.Window.t) => {
     None;
   | Some(glContext) =>
     Log.debug("Skia context created successfully.");
+    let framebufferId = Sdl2.Gl.getFramebufferBinding();
+    Log.debug(Printf.sprintf("Framebuffer binding %d.", framebufferId));
     let framebufferInfo =
       Gr.Gl.FramebufferInfo.make(
-        Unsigned.UInt.of_int(0),
+        Unsigned.UInt.of_int(framebufferId),
         Unsigned.UInt.of_int(0x8058),
       );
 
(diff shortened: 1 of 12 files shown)


src/reason-sdl2/sdl2_wrapper.cpp
```diff
@@ -1539,6 +1562,33 @@ CAMLprim value resdl_SDL_Init() {
...
+    char **argv = (char**)calloc(sizeof(char*), argc + 1);
+    for (int i = 0; i < argc; i++) {
+      argv[i] = (char*)String_val(Field(ml_argv, i));
+    }
+
+    resdl_uikit_main_callback = closure;
+    SDL_UIKitRunApp(argc, argv, [](int argc, char *argv[]){
+      caml_callback(resdl_uikit_main_callback, Val_unit);
+      return 0;
+    });
```
bryphe, line 1584: Cool! Nice find

src/Draw/CanvasContext.re
```diff
@@ -44,9 +44,11 @@ let create = (window: Revery_Core.Window.t) => {
     None;
   | Some(glContext) =>
     Log.debug("Skia context created successfully.");
+    let framebufferId = Sdl2.Gl.getFramebufferBinding();
```
bryphe, line 47: Nice catch!

PR #948, opened 2020-07-11
fix(text wrap): make UTF-8 aware

zbaylin:
Now that we have @OhadRau's revery-text-wrap in the monorepo, we need to make it UTF-8/Unicode aware so that we dont split on arbitrary byte boundaries. This was a big pain point when I was writing font-fallback, and this PR should fix it 👍

### DIFF
--- a/examples/text-wrap-cli/ReveryTextWrapCli.re
+++ b/examples/text-wrap-cli/ReveryTextWrapCli.re
@@ -1,42 +1,38 @@
 open Revery_TextWrap.Wrap;
 
 /* Get the width of a character */
-let width_of_char =
-  fun
-  | '1'
-  | 'i'
-  | 'l'
-  | 'I' => 0.5
-  | 'w'
-  | 'W'
-  | 'm'
-  | 'M' => 1.2
-  | '\t' => 4.0
-  | '\n' => 0.0
-  | _ => 1.0;
+let width_of_token = str => {
+  String.length(str) |> float;
+};
+
+Printexc.record_backtrace(true);
 
 let () = {
   print_endline("0:");
   Tokenize.split_tokens("0") |> Queue.iter(print_endline);
   print_endline("------------------");
-  wrap(~width_of_char, ~max_width=100.0, ~debug=true, "0")
+  wrap(~width_of_token, ~max_width=100.0, ~debug=true, "0")
   |> List.iter(print_endline);
   let wrapInput =
     "Here's another example of text where wrapping might be more difficult. "
     ++ "This string is very, very long and consists of words of varying lengths. "
     ++ "By utilizing some extremely long words, we can hopefully trigger some of "
     ++ "the more obscure edge-cases that word-wrapping can result in, such as "
-    ++ "placing a hyphen in the middle of a word on top of another hyphen.";
+    ++ "placing a hyphen in the middle of a word on top of another hyphen."
+    ++ "Let's also throw some UTF8 in here, who doesn't like a good emoji or 10?"
+    ++ "😀 🤔 🤭 🤫 🤥 😶 😐 😑 😬"
+    ++ "Also let's try some CJK!"
+    ++ "日本語の場合はランダムに生成された文章以外に、 著作権が切れた小説などが利用されることもある。";
   print_endline("==================");
-  wrap(~width_of_char, ~max_width=5.0, ~hyphenate=true, wrapInput)
+  wrap(~width_of_token, ~debug=true, ~max_width=5.0, ~hyphenate=true, wrapInput)
   |> List.iter(print_endline);
   print_endline("==================");
-  wrap(~width_of_char, ~max_width=10.0, ~hyphenate=true, wrapInput)
+  wrap(~width_of_token, ~max_width=10.0, ~hyphenate=true, wrapInput)
   |> List.iter(print_endline);
   print_endline("==================");
-  wrap(~width_of_char, ~max_width=15.0, wrapInput)
+  wrap(~width_of_token, ~max_width=15.0, wrapInput)
   |> List.iter(print_endline);
   print_endline("==================");
-  wrap(~width_of_char, ~max_width=40.0, wrapInput)
+  wrap(~width_of_token, ~max_width=40.0, wrapInput)
   |> List.iter(print_endline);
 };
--- a/examples/text-wrap-cli/ReveryTextWrapCli.re
+++ b/examples/text-wrap-cli/ReveryTextWrapCli.re
@@ -24,7 +24,13 @@ let () = {
     ++ "Also let's try some CJK!"
     ++ "日本語の場合はランダムに生成された文章以外に、 著作権が切れた小説などが利用されることもある。";
   print_endline("==================");
-  wrap(~width_of_token, ~debug=true, ~max_width=5.0, ~hyphenate=true, wrapInput)
+  wrap(
+    ~width_of_token,
+    ~debug=true,
+    ~max_width=5.0,
+    ~hyphenate=true,
+    wrapInput,
+  )
   |> List.iter(print_endline);
   print_endline("==================");
   wrap(~width_of_token, ~max_width=10.0, ~hyphenate=true, wrapInput)
--- a/examples/text-wrap-cli/ReveryTextWrapCli.re
+++ b/examples/text-wrap-cli/ReveryTextWrapCli.re
@@ -24,13 +24,7 @@ let () = {
     ++ "Also let's try some CJK!"
     ++ "日本語の場合はランダムに生成された文章以外に、 著作権が切れた小説などが利用されることもある。";
   print_endline("==================");
-  wrap(
-    ~width_of_token,
-    ~debug=true,
-    ~max_width=5.0,
-    ~hyphenate=true,
-    wrapInput,
-  )
+  wrap(~width_of_token, ~max_width=5.0, ~hyphenate=true, wrapInput)
   |> List.iter(print_endline);
   print_endline("==================");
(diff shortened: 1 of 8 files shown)


zbaylin:
I should also note I made the measuring much faster because before we were iterating over each character in a string and measuring it, rather than measuring the whole string itself. This is much better from a caching point of view since we arent measuring the same string twice but in different ways. For measuring characters, I just create a singleton using Zed, which should be constant time.

examples/text-wrap-cli/ReveryTextWrapCli.re
```diff
@@ -1,42 +1,38 @@
...
-  | 'w'
-  | 'W'
-  | 'm'
-  | 'M' => 1.2
-  | '\t' => 4.0
-  | '\n' => 0.0
-  | _ => 1.0;
+let width_of_token = str => {
+  String.length(str) |> float;
+};
```
zbaylin, line 6: Unfortunately changing the measuring algorithm makes writing a "fake" function for this a little harder, so right now Im just making each character (byte) the same size.

OhadRau: (Disclaimer: haven't looked at the code yet) how does this affect the actual implementation. I.e. are we still thinking about font width when we perform the actual wrapping?

zbaylin: @ohadrau Of course 😄. If you look in `TextNode.re`, you can see that we use Skia's text measurement API for it.

src/revery-text-wrap/wrap.re
```diff
@@ -95,78 +98,75 @@ let wrap_queue =
...
-                let char_width = width_of_char(token.[i]);
-                if (debug) {
-                  Printf.printf(
-                    "--Index: %d, Char: %c, Char_width: %f\n",
-                    i,
-                    token.[i],
-                    char_width,
-                  );
-                };
+              Zed_utf8.rev_explode(token)
```
OhadRau, line 101: Putting on my performance hat again: this is probably not _terrible_ but we are incurring quite a bit more overhead than the previous code again. Some reasons for that:

* Before we were iterating directly on the string, which means everything is a direct memory access. This is great because we basically guarantee cache hits and we never have to dereference extra addresses. Now this has been replaced with a list, which means each char = 1 more layer of indirection. Beyond just the overhead of a list, we now have to deal with the pitfalls of linked lists (i.e. more cache/TLB misses).
* Creating this exploded list leads to a bunch of tiny allocations for each cell in the list. One of the goals I had when originally writing this module was trying to avoid a lot of the downsides associated with functional code (i.e. lots of unnecessary allocations, lots of unnecessary abstractions like lambdas, etc.) since this code is probably along some very hot paths in a project like Oni. Luckily, this problem really lends itself to imperative style code so theoretically tools like buffers and queues and arrays should help us out.
* In the next line, we're now _reversing_ the string. I'm pretty confused by that, since there's just a normal `Zed_utf8.explode` function isn't there? But anyways, linked list reversal is going to cause us to iterate over the entire list & reallocate everything again :(

Don't take this as an issue with your code because I think it might just be a problem in the Zed_utf8 API. But if there's a better way of iterating over a string I would really go for that. It might be possible to do something akin to Java iterators (i.e. `next()` and `hasNext()` in a while loop could work, or just an `iteri` function should be fine).

src/revery-text-wrap/wrap.re
```diff
@@ -38,10 +40,11 @@ let wrap_queue =
        /* Store the width of this portion */
        let width = ref(0.0);
        /* Tokenize the line by whitespace and for each token: */
-       split_tokens(line)
+       let x = split_tokens(line);
```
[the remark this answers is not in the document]
zbaylin: Oops yeah that was when I was debugging something and wanted to intercept the value between pipes. I can change that in the morning.

src/revery-text-wrap/wrap.re
zbaylin: Oops yeah that's another hold out from my debugging - I need to be better about marking those as I go along.

Your point about exploding is valid -- I dont know how much work it would be to get some of the zed functions over to Buffers because the hard stuff relies on pattern matching, which isn't supported with buffers. It might end up being the case that writing those functions would cause more eventual overhead than just keeping the explosion as is. I totally get why you wanted to be imperative with this, and that's one of my gripes with functional programming too. However, if you look in zed I think they do a pretty good job of optimizing in a functional way, namely through only ever prepending at the head of a list. I would use for loops and all the other magical things a more C-style approach would bring, but that would require a lot of knowledge of the string that we don't really have -- primarily the bounds for each UTF-8 codepoint.

Honestly for now I think that decreasing the measuring overhead outweighs anything I introduced here (although I'd love to be proven wrong). Text measurement is a lot more expensive considering that fonts can be multiple kilobytes to hundreds of megabytes - keeping those in a hardware cache is a much bigger ask than a linked list of a couple strings. 

That brings me to my final point which is that I don't anticipate any string rendered being long enough such that it would make any meaningful difference. The explosion is occurring on tokens, not the entire string. 

I kind of rambled in some points here but I think you know me well enough to know what I'm trying to say. Let me know if you think I'm wrong about anything!

OhadRau: Agree about the possibility about having to rewrite Zed stuff. Obviously doesn't seem optimal and if it comes to that it's probably more of a long-term thing. I _really_ hope there is some better support for UTF-8 buffers somewhere for OCaml but it's really not the end of the world I guess.

In terms of the list stuff, yes & no. I think you're getting a little caught up on asymptotic time vs. real run time, which is something I've been super guilty of in the past. The problem here isn't really the fact that we're iterating over a list, but rather the fact that we're creating unnecessary indirection in memory. Linked-list iteration is great because it's asymptotically equivalent to array iteration & it's a very common access pattern. Unfortunately, modern processors aren't great at dealing with linked lists as compared to arrays. If you haven't watched [this talk from Jane Street](https://youtu.be/RV-4Xddk0Yc?t=26) yet I really recommend it -- the first couple mins talk about why memory accesses suck.

The biggest issue at play is the time it takes us to access memory. Arrays are much more cache friendly and so they're great for this kind of stuff. As we move to linked lists, we are splitting up related segments of memory into random locations all over the heap. In practice, the cache doesn't know that these are interlinked and that means that a much larger % of the time is spent chasing down pointers and looking up addresses to figure out where the physical memory is even located. To illustrate this, I ran a quick benchmark:

image

As you can see above, iterating over an array in a loop is ~1/2 the cost of recursively iterating over a list (without tail recursion). Iterating over the array _iteratively_ is a relatively smaller cost, but still takes 47% longer to complete (this is more or less equivalent to tail recursion). In either case, this alone is a major cost to consider and something we should avoid when we don't really need to rely on a linked list.

To make matters worse, linked lists mean extra allocation. This is bad in any language since we're making O(n) allocations and allocations are generally a ton of heap accesses + some complex logic, but it's especially bad in OCaml because we then need to consider the cost of garbage collection. The more allocations we have, the more loose pointers the runtime has to chase down each time the GC gets triggered. Then, each allocation still results in a call to `free()`.

Finally, we're simply doing too much computation here -- the data structure we started with is already able to do more or less what we need. Think about it, the iteration is already implemented to back the `explode` function. I'm not sure if it's exposed as part of the module, but since we're working with our own fork of Zed_utf8 to my understanding it should be relatively simple to just modify the code to expose something here.

The key here is that even though we're only working on small strings each time, these add up to probably 10x or more the cost of what we were doing before, without even accounting for the actual UTF8 part.

In general, these are great abstractions but there are some times where we need to turn off our FP brain and just tune into the way the computer wants you to write code. IMO this is one of those places, but again profiling might answer this better. It's definitely very possible that text measurement is far worse than any of this code, but at the very least I feel like we need to ditch reversing the list because that is just unnecessary. I would imagine that font measurement has some pretty clever data structures and logic, but I don't really have enough familiarity to know for sure. Do we happen to have any profiling set up that can easily be run? If not I can run some micro benchmarks with `Core_bench` in the morning and try to get an idea. Definitely don't mean to block this PR on something inconsequential, but I would feel better verifying the impact before merging.

zbaylin: You're right, last night I think I was arguing past your points and not really believing what I was saying -- the problem with doing code review at night 😄.

 `explode` is extraneous here because I was just using it as a mechanism to iterate and didn't use the list itself for anything. Creating a recursive function that does the same thing as explode sans the list creation should be better. 

In terms of linked lists in general, I share all your concerns. I actually hadn't seen that JS video (thanks!), but I often reference this video by Bjarne Stroustrup which covers a lot of the same points: https://youtu.be/YQs6IC-vgmo.

In terms of benchmarking, we use Rely. All of our benchmarks can be found in the `./bench` folder.

src/revery-text-wrap/wrap.re
```diff
@@ -95,78 +98,75 @@ let wrap_queue =
...
+                     if (width^ > 0.0) {
+                       /* If we're part-way through the string already */
+                       if (i > 0) {
+                         if (debug) {
+                           print_endline("--Clear+hyphenate");
+                         };
+                         /* We need to swap the last char of the buffer with a -, because the
+                            hyphen will be the last character that fits into this width. */
+                         let buffer_str = Buffer.contents(buffer);
+                         let last_uchar = last_uchar_of_string(buffer_str);
```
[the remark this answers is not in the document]
zbaylin: Totally agree with you on the last-char mechanism. I felt bad writing it, but as of now there really is no good way to do what is needed. In terms of the function call, I don't really think that is that impactful, but I can annotate it to be inlined if you think that's appropriate.

OhadRau:
Awesome! Changes look great

PR #951, opened 2020-07-12
chore(revery-text-wrap): cleanup

glennsl:


examples/text-wrap-cli/ReveryTextWrapCli.re
```diff
@@ -8,9 +8,6 @@ let width_of_token = str => {
 Printexc.record_backtrace(true);
 
 let () = {
-  print_endline("0:");
-  Tokenize.split_tokens("0") |> Queue.iter(print_endline);
-  print_endline("------------------");
```
glennsl, line 13: This seems like an internal that shouldn't be exposed, so I removed it. Feel free to correct me if I'm wrong though.

glennsl:
Thanks! This is really more of an interface cleanup, but sure why not!

packages/revery-text-wrap/src/Revery_TextWrap.re
```diff
@@ -109,32 +107,25 @@ let wrap_queue =
...
-                print_endline("Decision: ignore");
-              };
-              ();
-              /* If it's gonna be too long with the whitespace, don't add it & just prepare to wrap,
-                 so that we don't get the whitespace at the start of the next line. */
+              Log.trace(
+                "Decision: ignore",
+                /* If it's gonna be too long with the whitespace, don't add it & just prepare to wrap,
+                   so that we don't get the whitespace at the start of the next line. */
+              );
```
glennsl, line 114: refmt... :man_shrugging:

PR #954, opened 2020-07-13
chore(deps): upgrade esy-harfbuzz to 2.6.8

zbaylin:
This moves our dependency on esy-harfbuzz to our own version at 2.6.8.


zbaylin:
@bryphe those health checks we talked about -- did you want to bake those in here or in Oni? Let me know if here and I can make them part of this PR!

bryphe:
@zbaylin - implementing them in Oni is fine!

Here's where we'd add it:
https://github.com/onivim/oni2/blob/65cef61bf8973d2f808f8a3dc8bc6a1eefd2f4b4/src/bin_editor/HealthCheck.re#L67

(Similar to the `libuv` check).

We'd just want to verify we get a valid version for both the 'compiled' harfbuzz version and the 'linked' harfbuzz version - so we might need to expose those as APIs on the reason side, is the only necessary addition.

zbaylin:
@bryphe Sounds good -- I added those version functions in the form of `hb_version_string_compiled` and `hb_version_string_runtime`

PR #956, opened 2020-07-14
feat(harfbuzz): add start and stop API to shape

zbaylin:
Rather than creating substrings to shape, we can just pass in the start and stop into Harfbuzz 👍

### DIFF
--- a/packages/reason-harfbuzz/src/Harfbuzz.re
+++ b/packages/reason-harfbuzz/src/Harfbuzz.re
@@ -64,9 +64,14 @@ let hb_shape = (~features=[], ~start=`Start, ~stop=`End, {face}, str) => {
        )
     |> Array.of_list;
   let startPosition = positionToInt(start);
-  let stopPosition = positionToInt(stop);
+  let length =
+    switch (stop) {
+    | `Position(n) => n - startPosition
+    | `Start => 0
+    | `End => (-1)
+    };
 
-  Internal.hb_shape(face, str, arr, startPosition, stopPosition);
+  Internal.hb_shape(face, str, arr, startPosition, length);
 };
 let hb_new_face = str => hb_face_from_path(str);
 
--- a/packages/reason-harfbuzz/src/Harfbuzz.re
+++ b/packages/reason-harfbuzz/src/Harfbuzz.re
@@ -16,7 +16,8 @@ module Internal = {
   external hb_face_from_data: (string, int) => result(face, string) =
     "rehb_face_from_bytes";
   external hb_destroy_face: face => unit = "rehb_destroy_face";
-  external hb_shape: (face, string, array(feature)) => array(hb_shape) =
+  external hb_shape:
+    (face, string, array(feature), int, int) => array(hb_shape) =
     "rehb_shape";
 };
 
@@ -50,7 +51,7 @@ let hb_face_from_path = str => {
   };
 };
 
-let hb_shape = (~features=[], {face}, str) => {
+let hb_shape = (~features=[], ~start=`Start, ~stop=`End, {face}, str) => {
   let arr =
     features
     |> List.map(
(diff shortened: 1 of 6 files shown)


packages/reason-harfbuzz/src/Harfbuzz.re
```diff
@@ -62,7 +63,15 @@ let hb_shape = (~features=[], {face}, str) => {
...
        )
     |> Array.of_list;
-  Internal.hb_shape(face, str, arr);
+  let startPosition = positionToInt(start);
+  let length =
+    switch (stop) {
+    | `Position(n) => n - startPosition
+    | `Start => 0
+    | `End => (-1)
+    };
```
glennsl, line 72: What happens if `n` is smaller than `startPosition` so that `length` ends up more negative than `-1`? Does it still just treat it as `` `End``?

packages/reason-harfbuzz/src/Harfbuzz.rei
```diff
@@ -20,4 +20,11 @@ let hb_face_from_skia: Skia.Typeface.t => result(hb_face, string);
...
 let hb_shape:
-  (~features: list(feature)=?, hb_face, string) => array(hb_shape);
+  (
+    ~features: list(feature)=?,
+    ~start: position=?,
+    ~stop: position=?,
+    hb_face,
+    string
+  ) =>
+  array(hb_shape);
```
glennsl, line 30: I like that the API extension is non-breaking and still really nice!

packages/reason-harfbuzz/src/Harfbuzz.re
zbaylin: I think eventually the int is made unsigned so it would just be `INT_MAX - n` -- it ends on `NULL` termination anyway, and I doubt we will be shaping any string anywhere near the 32/64 bit integer limit haha

packages/reason-harfbuzz/src/harfbuzz.cpp
```diff
@@ -143,9 +146,10 @@ CAMLprim value rehb_shape(value vFace, value vString, value vFeatures) {
 
   hb_font_t *hb_font = (hb_font_t *)vFace;
 
+
   hb_buffer_t *hb_buffer;
   hb_buffer = hb_buffer_create();
-  hb_buffer_add_utf8(hb_buffer, String_val(vString), -1, 0, -1);
+  hb_buffer_add_utf8(hb_buffer, String_val(vString), -1, start, len);
```
bryphe, line 152: It would be great if we had a test case to cover the new functionality - ie, 'abc' with a `start` of `1` and `len` of 1 should just have a `b` glyph. A unicode case would be helpful as well - to verify the start/length behavior is bytes rather than index.

packages/reason-harfbuzz/src/Harfbuzz.re
bryphe: This would be good to have covered by a test case, too, just to be sure!

packages/reason-harfbuzz/src/harfbuzz.cpp
zbaylin: Sounds good -- I can add tests for it to make sure it behaves as expected!

zbaylin: Just added tests for both -- the UTF-8 example uses bytes as you would expect!

packages/reason-harfbuzz/test/ShapingTest.re
```diff
@@ -0,0 +1,32 @@
...
+
+  test("substring", ({expect, _}) => {
+    let expectedResult = [|{glyphId: 70, cluster: 1}|];
+    let shapes =
+      hb_shape(font, "abc", ~start=`Position(1), ~stop=`Position(2));
+
+    expect.equal(expectedResult, shapes);
+  });
+
+  test("substring UTF-8", ({expect, _}) => {
```
bryphe, line 25: Excellent! Thanks for adding these, @zbaylin

PR #957, opened 2020-07-14
reafactor(feat/font/fallback): self-contained generateShapes

glennsl:
This doesn't even work properly right now, and could use some cleanup, but it shows the basic ideas I wanted to get a cross. The hole resolver has been simplified and is fully contained inside `generateShapes`, I've also tried to prepare it for using the bounded harfbuzz API. Having the bounds be bytes-based and the loop index be glyph-based is a bit confusing though. And it does need to be threaded through a lot of functions.

I'll keep working on this a bit to try to figure out why it's not working properly, and see if I can find a better way to pass the bounds around. I think it's good enough to run by you for a preliminary review though, or to show it just as an example if you want to go in a different direction.

### DIFF
--- a/src/Font/FontCache.re
+++ b/src/Font/FontCache.re
@@ -243,21 +243,21 @@ let generateShapes:
 
     and loopShapes =
         (
-          ~stop,
+          ~stopCluster,
           ~acc,
           ~holeStart=?,
           ~index,
           {hbFace, skiaFace, _} as font,
           shapes,
         ) => {
-      let resolvePossibleHole = () =>
+      let resolvePossibleHole = (~stop) =>
         switch (holeStart) {
         | Some(start) => resolveHole(~acc, ~start, ~stop)
         | None => acc
         };
 
       if (index == Array.length(shapes)) {
-        resolvePossibleHole();
+        resolvePossibleHole(~stop=stopCluster);
       } else {
         let Harfbuzz.{glyphId, cluster} = shapes[index];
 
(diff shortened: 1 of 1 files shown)


zbaylin:
Thanks @glennsl! Your approach is kind of similar to an alternative I had in mind where `Hole.resolve` returns a "resolution" which is a substring (or bounds) and a maybe typeface. What specifically isn't working, if I may ask?

glennsl:
Right, that would probably be easier now with the accumulator.

> What specifically isn't working, if I may ask?

It's not showing the emoji in the title of the examples app. Not even the unknown char square. So I think it's resolving it correctly but not adding it to the accumulator for some reason...

src/Font/FontCache.re
```diff
@@ -192,159 +192,99 @@ let matchCharacter = (fallbackCharacterCache, uchar, skiaFace) =>
...
+              ...acc,
+            ],
+            ~start=start + 1,
+            ~stop,
+          )
+        };
+      }
+
+    and loop_shapes = (~stop, ~acc, ~holeStart=?, ~index, shapes) => {
+      let resolvePossibleHole = (~stop) =>
```
bryphe, line 240: Why snake_case for `loop_shapes`?

glennsl: Good question. This started off as kind of a recursive state machine, and I like to use the underscore to distinguish the state as kind of a short-hand for `loopWhenState`. This no longer makes sense though, so you're absolutely right. Already changed.

glennsl:
Found the bug. I was getting the right glyph but using the font from `generateShapes` instead of the fallback font when creating the `ShapeResult`.

glennsl:
I think this is OK now, although I'm not entirely sure since there doesn't seem to be any tests outside the Hello example and without the updated esy-harfbuzz I'm not sure what works and not. I tried to play with the emoji in the Hello example, but that made Onivim freeze my computer 🙃.

Feel free to merge or tweak this however you want now.

zbaylin:
Awesome, thanks @glennsl -- it looks great. I'm a little nervous about it freezing your computer, but Im confused about you saying Oni caused it -- was that a typo or are you resolving Oni to this branch?

glennsl:
Oh, sorry, no it has nothing to do with this PR. I was running Oni normally and just found it a bit ironic that it happened while working on or near something that might eventually fix it. Just a bit of encouragement :)

zbaylin:
Oh haha now I get it -- sounds good. I'll merge this into my branch and then when #956 is merged I'll use that API

Issue #470, opened 2019-04-25
Performance: Show FPS counter


bryphe:
We should make it easy to add an FPS counter / track FPS - on both the native and JSOO side.

Related to #230 , in the sense of having a performance metric be front-and-center (FPS counter gives a view of overall performance, while a tracking square can help identify stutter/jitter).

Thanks @sgrove for the suggestion!

A couple of things we'll have to consider:

- What's the right place for this? Should the rendering pipeline track FPS, and expose a boolean option, or should we implement an `<FPSCounter />` in user-space?
- In the case where there is no animations active / no active re-renders - should we just show `FPS: -`? Alternatively, we could show a metric like 'lastRenderTime' which is the time in ms spent to render the last frame.

Grinshpon:
I decided to work on this and my progress is on my [add-fps-counter](https://github.com/Grinshpon/revery/tree/add-fps-counter) branch.

Keeping track of fps is simple, I just needed to add a counter to the window type in `Core/Window` that stores the tick (from SDL's `getTick`) after each render cycle, and the time it took to render the last frame, using that.

Then I gave the `Window` module two new functions: `getFPS`, which returns the FPS as a float, and `getLastRenderTime`, which returns the last render time in milliseconds as an int. 

I'm gonna see about adding an `<FPSCounter />` and then make a PR with my changes.

Grinshpon:
Since my PR was merged this issue can be closed. Sorry I forgot in to write "resolves #..." in the message.

glennsl:
Thanks for the PR @Grinshpon!

PR #962, opened 2020-07-18
Add fps counter

Grinshpon:
Add an fps counter (issue #470) and a way to display it. To enable or disable it, call the function `Window.setFPSCounter`:

    Window.setFPSCounter(window,true/false);


OhadRau:
any chance you could land a frame render time in this? should just be 1 more line of text to display but I think it's super useful for debugging

Grinshpon:
It'll take a bit more effort to make it readable, since it's updated every render frame it'll look like a bunch of flickering digits.

examples/Examples.re
```diff
@@ -332,6 +332,7 @@ let init = app => {
       app,
       "Welcome to Revery!",
     );
+  //Window.setFPSCounter(window,true);
```
zbaylin, line 335: **Minor**: Can you remove this comment please? We don't usually have dead code commented out.

examples/Stopwatch.re
```diff
@@ -46,7 +46,7 @@ module Clock = {
      * We'll make sure to dispatch the 'Stop' action when unmounting,
      * so we don't have a runaway timer!
      */
-    let%hook () = Hooks.effect(OnMount, () => Some(() => dispatch(Stop)));
+    let%hook () = Hooks.effect(OnMount, () => Some(_ => dispatch(Stop)));
```
zbaylin, line 49: **Minor** Can you change this back to the unit? If the function signature changes in the future we probably want the compiler to give an error.

src/Core/Window.re
```diff
@@ -151,6 +151,41 @@ module WindowMetrics: {
...
+    tickAfterLastRender: Sdl2.Timekeeping.getTicks(),
+    lastRenderTime: 0,
+    fps: 0.0,
+    fpsTimerDispose: noop,
+  };
+  let getFPS = (c: t) => c.fps;
+  let getLastRenderTime = (c: t) => c.lastRenderTime;
+  /* let setLastRenderTime = (c: t, time) => {
+       c.lastRenderTime = time;
+     }; */
```
zbaylin, line 173: **Minor**: More dead code removal

src/Core/Window.re
```diff
@@ -151,6 +151,41 @@ module WindowMetrics: {
...
+  let setFPSCounter = (c: t, s) => {
```
zbaylin, line 174: Does this need to be mutable? I.e. can you just return a new record with the `...x` spread?

src/Core/Window.re
```diff
@@ -774,3 +819,20 @@ let setRenderCallback = (window: t, callback) => window.render = callback;
...
+};
+
+let getLastRenderTime = (w: t) => {
+  FPS.getLastRenderTime(w.fpsCounter);
+};
+
+let setFPSCounter = (w: t, s) => {
+  w.showFPSCounter = s;
+  FPS.setFPSCounter(w.fpsCounter, s);
+};
```
zbaylin, line 834: Same mutation question here -- not a huge deal, just wondering.

src/Core/Window.re
Grinshpon: Since it's updated so frequently (every frame) I thought it made more sense to make it mutable.

zbaylin: Ah good point, that makes sense. Sounds good to me!

Grinshpon:
Alright, I'm pushing a commit that resolves everything above, and I added a `--show-fps` flag in the example.

Grinshpon:
@glennsl you're right, taking the average works and looks much better. As for the second part, wouldn't that essentially just be me manually re-implementing a timer?

glennsl:
It would, but more declaratively. You don't have to manage a timer instance, and it's overall much simpler. A timer (in the render loop) is just a counter and a conditional asking "is it time yet?".

Grinshpon:
makes sense, I'll do that

examples/Examples.re
```diff
@@ -332,6 +334,9 @@ let init = app => {
       app,
       "Welcome to Revery!",
     );
+  if (showFPSCounter^) {
+    Window.setFPSCounter(window, true);
+  };
```
glennsl, line 339: THis is pretty much equivalent to
```suggestion
  Window.setFPSCounter(window, showFPSCounter^);
```
Unless `setFPSCounter` has side-effects (outside of mutating the flag).

src/Core/Window.re
```diff
@@ -151,6 +151,47 @@ module WindowMetrics: {
...
+
+    if (c.timer >= 1000) {
+      c.timer = 0;
+      if (c.fpsAccum != 0.0) {
+        c.fps = float_of_int(c.ticksSinceUpdate) /. c.fpsAccum;
+      };
+      c.ticksSinceUpdate = 0;
+      c.fpsAccum = 0.0;
+    };
+  };
```
glennsl, line 192: Good call on renaming this `update`. You should almost never use a single-character variable name though, since it carries almost no meaning. In this case I would just call it `state`.

Otherwise this is till very complicated. All you really need is a time period and the number of frames rendered within that period. The time period comes from the timer itself, which is conveniently set to 1 second. So now all you need is the frame count, which should just be the number of times `update` has been called since reset.

src/Core/Window.re
```diff
@@ -163,6 +204,8 @@ type t = {
   mutable metrics: WindowMetrics.t,
   mutable isRendering: bool,
   mutable requestedUnscaledSize: option(size),
+  mutable fpsCounter: FPS.t,
+  mutable showFPSCounter: bool,
```
glennsl, line 208: Do these need to be separate? And if `showFPSCounter` is `false`, is the information in `fpsCounter` still valid? Alternatively, do we need to run code to keep `fpsCounter` up to date if we're not using it? Could we perhaps use a variant to make sure we don't run unnecessary code, and at the same time don't keep invalid state?

src/Core/Window.re
```diff
@@ -774,3 +822,19 @@ let setRenderCallback = (window: t, callback) => window.render = callback;
...
+  FPS.getLastRenderTime(w.fpsCounter);
+};
+
+let setFPSCounter = (w: t, s) => {
+  w.showFPSCounter = s;
+};
+
+let shouldShowFPSCounter = (w: t) => {
+  w.showFPSCounter;
+};
```
glennsl, line 840: Do we really need all these function in addition to `update`? What if `update` just returned the FPS directly as an `option`? `None` meaning it's disabled.

Also, `setFPSCounter` doesn't actually set the FPS counter. It sets the show flag. But even if named correctly, boolean arguments are a bit of a smell as it's often not clear what they mean. I would prefer separate `enable` and `disable` functions instead.

src/Core/Window.re
Grinshpon: `showFPSCounter` is checked to see whether the counter should be rendered, setting it doesn't affect the information in `fpsCounter`

Grinshpon: I feel like fpsCounter should be left to run in the background, so that the user still has the ability to get to get that information if they don't want it displayed.

src/Core/Window.re
Grinshpon: I think it would get a little murky if `update` did everything and also returned the FPS. It's only supposed to be called once at the end of each render frame.

src/Core/Window.re
glennsl: What would they do with it if not display it?

src/Core/Window.re
Grinshpon: Haha the single-character variable is a terrible habit, sorry about that.

As for the contents, the frame count is stored in `ticksSinceUpdate`, which I'm realizing is terribly named, and I'll change it to `frameCount`. But I am pretty much doing what you described. The first stanza is keeping track of the time it took to render the last frame.

src/Core/Window.re
Grinshpon: They could log it, or choose to create a custom display.

examples/Examples.re
Grinshpon: `setFPSCounter` is gone and replaced by `enable...` and `disable...` as per your other comment.

src/Core/Window.re
glennsl: > The first stanza is keeping track of the time it took to render the last frame.

Why do you need that? And why do you need `fpsAccum`? When a second has passed, is `frameCount` not already the number of frames per second?

src/Core/Window.re
glennsl: Unless it's logged every second, it'd still just be inaccurate snapshots. If it's logged every second it's pretty much information overload. And in any case this isn't information that's particularly useful in a log I don't think. I'm also not a big believer in "just in case someone needs it" features, as these things tend to pile up into an unmaintainable mess over time. And it also often adds assumptions and invariants that are eventually forgotten, because it's not typically used in that way. Not that big of a deal, though, and I think this is a good way to do it if that's what you want.

src/Core/Window.re
glennsl: > It's only supposed to be called once at the end of each render frame.

I guess this mostly goes back to the previous comment and depends on what conclusion you draw from that, but I don't see why it would suddenly be called multiple times. It's still called `update` and has the same side-effects.

src/Core/Window.re
Grinshpon: Huh, you're right. I don't need fpsAccum at all.

And @OhadRau mentioned to me that frame render time can be very useful in debugging. I'm also considering taking his advice and displaying the average render time each second along with the fps.

Grinshpon:
Before this is merged, I'm considering also showing the average frame render time like @OhadRau suggested earlier. What should it be labelled as when displayed? I'm thinking `AFT: ...` for average frame time.

glennsl:
I think what @OhadRau was referring to is the time it took to render each frame, not the time between frames. And I think that would be very useful as well. That's actually a thing we have been getting from logging via `Performance.bench`, and it's an awfully noisy way of doing it. Replacing that would be great!

Grinshpon:
In that case, it might be worth creating a separate issue for that, instead of tacking it on to this pr.

src/Core/Window.re
Grinshpon: I simplified `update`, it should be looking better now.

src/Core/Window.re
Grinshpon: I think I would prefer to leave it this way. I agree that adding stuff for the sake of adding it isn't ideal and I certainly don't want to add any bloat to revery, but I still believe having an ongoing fps metric can be useful.

src/Core/Window.re
Grinshpon: I may have slightly misunderstood your comment initially, but, going back to the other conversation, I think I would prefer to have it like this.

Grinshpon:
Nice, all checks pass, I'd like to know if I've sufficiently satisfied all requests?

src/Core/Window.re
```diff
@@ -151,6 +151,42 @@ module WindowMetrics: {
   let markDirty = metrics => {...metrics, isDirty: true};
 };
 
+module FPS = {
+  let timerInterval = 1000;
+  type t = {
+    // tickAfterLastRender and lastRenderTime are in milliseconds
+    mutable tickAfterLastRender: int,
+    mutable lastRenderTime: int,
```
glennsl, line 159: I still don't see any reason to keep `lastRenderTime` in the state. As implemented it's a not very useful snapshot, and a more useful average can easily be derived from `fps`  anyway.

Grinshpon: I removed it in the latest commit.

bryphe:
Thanks for the contribution, @Grinshpon !

PR #963, opened 2020-07-20
fix(text): make overflow UTF-8 aware

zbaylin:
This is a pretty simple fix, since Zed already gives us a function to do exactly what we were doing before but with UTF-8. 

I will also add some tests for this, just to make sure it's robust!

### DIFF
--- a/src/Core/TextOverflow.re
+++ b/src/Core/TextOverflow.re
@@ -10,7 +10,7 @@ let removeLineBreaks = text => {
 };
 
 let rec handleOverflow = (~maxWidth, ~text, ~measure, ~character="…", ()) => {
-  let clippedText = String.length(text) - 1 |> Str.string_before(text);
+  let clippedText = Zed_utf8.rchop(text);
 
   let width = measure(clippedText ++ character);
   width >= maxWidth && String.length(clippedText) > 1
--- a/test/Core/TextTests.re
+++ b/test/Core/TextTests.re
@@ -1,0 +1,37 @@
+open Revery_Core;
+open TestFramework;
+
+let font =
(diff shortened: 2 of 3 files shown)


src/Core/TextOverflow.re
```diff
@@ -10,7 +10,7 @@ let removeLineBreaks = text => {
 };
 
 let rec handleOverflow = (~maxWidth, ~text, ~measure, ~character="…", ()) => {
-  let clippedText = String.length(text) - 1 |> Str.string_before(text);
+  let clippedText = Zed_utf8.rchop(text);
```
bryphe, line 13: Nice find! I didn't know about this method - I'm glad there was already an efficient way to handle this.

test/Core/TextTests.re
```diff
@@ -0,0 +1,37 @@
...
+      Revery_Font.measure(
+        ~smoothing=Revery_Font.Smoothing.default,
+        font,
+        12.0,
+        str,
+      ).
+        width;
+
+    let abcSize = measure("ABC");
+    let overflown =
```
bryphe, line 28: Thanks for adding a test case 👍

test/Core/TextTests.re
```diff
@@ -0,0 +1,37 @@
...
+  Revery_Font.Family.fromFile("Roboto-Regular.ttf")
+  |> Revery_Font.Family.toSkia(Revery_Font.Weight.Normal)
+  |> Revery_Font.load
+  |> (
+    result =>
+      switch (result) {
+      | Ok(font) => font
+      | Error(e) => failwith(e)
+      }
+  );
```
bryphe, line 14: Since this is a test case - would be fine to use `Result.get_ok` here (would simplify these lines)

PR #964, opened 2020-07-20
fix(harfbuzz): uncouple from Skia

zbaylin:
@glennsl mentioned in #927 that these two libraries should be separate, and I totally agree. This uncouples them completely, except for the examples/CI to get font discovery working. @glennsl, let me know if you want me to remove that as well.


glennsl:
> This uncouples them completely, except for the examples/CI to get font discovery working. @glennsl, let me know if you want me to remove that as well.

That would be great, but not essential. I'm moving the examples back into the respective packages as well, and it would be nice to be able to work on harfbuzz without having to compile the monstrosity that is skia.

zbaylin:
@glennsl sounds good, I can decouple them. It's not necessary for Harfbuzz to have font discovery.

PR #968, opened 2020-07-24
feat(window): expose a few more window-options

### LINKED ISSUE #506: Allow windows to be transparent
Since GLFW supports transparent framebuffers and also transparent windows including decorations for some operating systems it would be nice if this feature could be accessible through revery.

See [this](https://www.glfw.org/docs/latest/window_guide.html#GLFW_TRANSPARENT_FRAMEBUFFER_hint) GLFW window hint for the transparent framebuffer and further down on the same page is an explanation how to make the whole window transparent.

lessp:
This adds a `transparency`-option to Window.

EDIT: I guess though, that this puts a transparency on everything, not only the "background-window" as you'd expect. 🤔 

Would close #506

### DIFF
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -352,6 +352,12 @@ let setSize = (~width: int, ~height: int, win: t) => {
   Internal.setRawSize(win, adjWidth, adjHeight);
 };
 
+let setMinimumSize = (~width: int, ~height: int, win: t) => {
+  Log.tracef(m => m("setMinimumSize - calling with: %ux%u", width, height));
+
+  Sdl2.Window.setMinimumSize(win.sdlWindow, width, height);
+};
+
 let setZoom = (window, zoom) => {
   window.metrics = WindowMetrics.setZoom(max(zoom, 0.1), window.metrics);
 };
@@ -650,6 +656,14 @@ let create = (name: string, options: WindowCreateOptions.t) => {
   // until after we create the window, so after creation we have to check
   // to make sure we're at the correct size for scaling.
   setSize(~width, ~height, window);
+
+  // Set a minimum size for the window
+  setMinimumSize(
+    ~width=options.minimumWidth,
+    ~height=options.minimumHeight,
+    window,
+  );
+
   setVsync(window, options.vsync);
 
   if (!options.decorated) {
@@ -674,11 +688,6 @@ let create = (name: string, options: WindowCreateOptions.t) => {
   if (options.maximized) {
     Sdl2.Window.maximize(sdlWindow);
   };
-
-  // onivim/oni2#791
-  // Set a minimum size for the window
-  // TODO: Make configurable
-  Sdl2.Window.setMinimumSize(sdlWindow, 200, 100);
 
   Sdl2.Window.setTransparency(sdlWindow, options.transparency);
 
@@ -711,7 +720,9 @@ let setInputRect = (_w: t, x, y, width, height) => {
 
 let setTransparency = (w: t, transparency) => w.transparency = transparency;
 
-let setBackgroundColor = (w: t, color: Color.t) => w.backgroundColor = color;
+let setBackgroundColor = (w: t, color: Color.t) => {
+  w.backgroundColor = color;
+};
 
 let getBackgroundColor = window => window.backgroundColor;
 
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -243,6 +243,7 @@ module Internal = {
       }
     | Android
     | Browser
+    // NOTE: might work for IOS?
     | IOS
     | Linux
     | Unknown
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -241,8 +241,12 @@ module Internal = {
       | Hidden => Sdl2.Window.setMacTitlebarHidden(w)
       | System => ()
       }
-    | Windows
-    | Linux => ()
+    | Android
+    | Browser
+    | IOS
+    | Linux
+    | Unknown
+    | Windows => ()
     };
   };
 
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -241,7 +241,8 @@ module Internal = {
       | Hidden => Sdl2.Window.setMacTitlebarHidden(w)
       | System => ()
       }
-    | Windows | Linux => ()
+    | Windows
+    | Linux => ()
     };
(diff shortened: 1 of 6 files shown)


lessp:
I'm not sure what to do about the transparency-flag. I think it makes sense to expose it for full opacity, but perhaps we should name it something else as to reserve it for pure background-transparency?

src/Core/WindowCreateOptions.re
```diff
@@ -72,10 +84,13 @@ let create =
       ~maximized=false,
       ~decorated=true,
       ~titlebarStyle=WindowStyles.System,
+      ~transparency=1.0,
```
glennsl, line 87: This sets the opacity, not transparency, which is its inverse. Fully transparent means it's see-through, while fully opaque means it's solid.

src/Core/WindowCreateOptions.re
```diff
@@ -24,6 +24,10 @@ type t = {
    independent, but as of right now, Transparent only works on macOS.
```
glennsl, line 24:
```suggestion
   independent, but as of right now, [Transparent] and [Hidden] only works on macOS.
```

src/Core/WindowCreateOptions.re
```diff
@@ -44,6 +48,14 @@ type t = {
     [height] is the initial vertical size of the [Window], with display scaling applied.
   */
   height: int,
+  /**
+    [minimumWidth] is the minimal horizontal size of the [Window].
```
glennsl, line 52:
```suggestion
    [minimumWidth] is the minimum horizontal size of the [Window].
```

src/Core/WindowCreateOptions.re
```diff
@@ -44,6 +48,14 @@ type t = {
...
+    */
+  minimumWidth: int,
+  /**
+    [minimumWidth] is the minimal vertical size of the [Window].
```
glennsl, line 56:
```suggestion
    [minimumWidth] is the minimum vertical size of the [Window].
```

src/Core/Window.re
```diff
@@ -232,16 +233,24 @@ type t = {
...
+  };
 
   let resetTitlebarStyle = window =>
     // On restore, we need to set the titlebar transparent again on Mac
-    if (window.titlebarStyle == Transparent) {
-      setTitlebarTransparent(window.sdlWindow);
+    switch (window.titlebarStyle) {
+    | Transparent => setTitlebarStyle(window.sdlWindow, Transparent)
+    | Hidden => setTitlebarStyle(window.sdlWindow, Hidden)
+    | _ => ()
```
glennsl, line 253: Avoid using the wildcard pattern if possible, as you lose the benefit of exhaustiveness checking and letting the compiler work for you.

```suggestion
    | System => ()
```

src/Core/Window.re
```diff
@@ -232,16 +233,24 @@ type t = {
...
 module Internal = {
-  let setTitlebarTransparent = (w: Sdl2.Window.t) =>
+  let setTitlebarStyle = (w: Sdl2.Window.t, style: WindowStyles.titlebar) => {
     switch (Environment.os) {
-    | Mac => Sdl2.Window.setMacTitlebarTransparent(w)
+    | Mac =>
+      switch (style) {
+      | Transparent => Sdl2.Window.setMacTitlebarTransparent(w)
+      | Hidden => Sdl2.Window.setMacTitlebarHidden(w)
+      | _ => ()
```
glennsl, line 242:
```suggestion
      | System => ()
```

src/Core/Window.re
```diff
@@ -232,16 +233,24 @@ type t = {
...
+      }
     | _ => ()
```
glennsl, line 244:
```suggestion
    | Windows | Linux => ()
```

glennsl:
> I'm not sure what to do about the transparency-flag. I think it makes sense to expose it for full opacity, but perhaps we should name it something else as to reserve it for pure background-transparency?

Since it's associated with the window, I don't think it''s a problem. Background transparency would instead be associated with painting I think.

src/Core/WindowCreateOptions.re
lessp: Ah, that's true! Switch our naming to `opacity` and leave a comment perhaps?

lessp:
> > I'm not sure what to do about the transparency-flag. I think it makes sense to expose it for full opacity, but perhaps we should name it something else as to reserve it for pure background-transparency?
> 
> Since it's associated with the window, I don't think it''s a problem. Background transparency would instead be associated with painting I think.

Thanks for reviewing and the suggested changes @glennsl!

src/Core/WindowCreateOptions.re
glennsl: Yeah, that sounds good!

lessp: Updated, let me know if you think we're good to go!

src/Core/Window.re
```diff
@@ -698,7 +698,9 @@ let create = (name: string, options: WindowCreateOptions.t) => {
     Sdl2.Window.maximize(sdlWindow);
   };
 
-  Sdl2.Window.setTransparency(sdlWindow, options.transparency);
+  // Sdl2 has this named as Transparency, but it actually works as opacity
+  // where a value of 1.0 means it's fully opaque
+  Sdl2.Window.setTransparency(sdlWindow, options.opacity);
```
glennsl, line 709: SDL2 actually has it right. It's `reason-sdl` that gets it wrong. And since `reason-sdl` is in this repo we could fix it completely in this PR.

src/Core/WindowCreateOptions.re
```diff
@@ -21,9 +21,15 @@ type t = {
...
    [titlebarStyle] sets the appearance of the titlebar. Eventually this will be platform
-   independent, but as of right now, Transparent only works on macOS.
+   independent, but as of right now, [Transparent] and [Hidden] only works on macOS.
     */
   titlebarStyle: WindowStyles.titlebar,
+  /**
+   [opacity] sets the opacity of the window.
+   A value of 1.0 means the window is fully opaque,
+   and a value of 0.0 means the window is completely transparent.
+   */
```
glennsl, line 31: :100:

src/Core/Window.re
lessp: Must have missed this!

PR #970, opened 2020-07-27
fix(font): Fix hang in font-fallback code

bryphe:
__Issue:__ Certain unicode constructions could cause the font-fallback algorithm to hang, as described in https://github.com/onivim/oni2/issues/2178

__Defect:__ We'd run into an infinite loop in the `fallbackFor` code, because we were assuming a `character index` when we actually needed a `byte offset`. This is another case where it'd be great to have the type system help us differentiate between these!

__Fix:__ The value passed into `fallbackFor` is a byte index, not a character index - so use `Zed_utf8.extract`, which operates on byte offsets, instead of `get`, which operates on character indices. Add several test cases.

### DIFF
--- a/src/Font/FontCache.re
+++ b/src/Font/FontCache.re
@@ -207,11 +207,17 @@ let generateShapes:
       Log.debugf(m =>
         m("Resolving fallback for: %s at byte offset %d", str, byteOffset)
       );
-      let uchar =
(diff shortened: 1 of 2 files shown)


bryphe:
Looks like this fixes the issue with the test file that was hanging, previously:

image

PR #972, opened 2020-07-28
feat(native): add initial tray-support for macOS

lessp:


examples/Examples.re
```diff
@@ -368,6 +368,12 @@ let init = app => {
       Console.log(Printf.sprintf("Moved: %d x %d", x, y))
     );
 
+  Native.Tray.make(
```
bryphe, line 371: This is awesome, @lessp ! Thanks for imlpementing this.

examples/Examples.re
```diff
@@ -368,6 +368,12 @@ let init = app => {
...
+    ~title=`Image(Environment.getAssetPath("outrun-logo.png")),
```
bryphe, line 372: In terms of the API - is it possible to have _both_ a `title` and an `image`? It seems like that might be the case on MacOS, at least - so it might be worth having an API like:
```
Native.Try.make(
~imagePath:string=?,
~title:string=?,
unit
)
```

An optional `title` and `imagePath` argument, so we can specify one or both.

On Windows, there's just the icon - but the title could be used as the tooltip for it.

bryphe: On Electron - this is the API: https://www.electronjs.org/docs/api/tray

It always requires an `iconPath` (`new Tray(iconPath)`), but then lets the consumer set a title later.

src/Native/tray.c
```diff
@@ -0,0 +1,44 @@
...
+    if (Field(title_v, 0) == hash_variant("Text")) {
+        ret = revery_makeTrayHandleText_cocoa(title);
+    } else {
+        ret = revery_makeTrayHandleImage_cocoa(title);
+    };
+
+    UNUSED(title);
+#elif USE_WIN32
+    fprintf(stderr, "WARNING: %s is not implemented on this platform.", __func__);
+    ret = NULL;
```
bryphe, line 38: Printing out to `stderr` can potentially cause a crash on Windows if there isn't a console allocated, unfortunately - so we may not want to log this. Alternatively, we could return a `result` from the stub here, and handle the logging in the OCaml side - but might be overkill.

src/Native/tray.c
```diff
@@ -0,0 +1,44 @@
...
+#else
+    fprintf(stderr, "WARNING: %s is not implemented on this platform.", __func__);
+    ret = NULL;
```
bryphe, line 41: I believe `ret` may need to be `Val_unit` instead of `NULL`.

src/Native/tray_cocoa.c
```diff
@@ -0,0 +1,31 @@
+#include "config.h"
+#ifdef USE_COCOA
+#include <stdio.h>
+
+#import <Cocoa/Cocoa.h>
+
+void *revery_makeTrayHandleText_cocoa(const char *titleText) {
+    NSStatusItem *statusItem = [NSStatusBar.systemStatusBar statusItemWithLength:NSVariableStatusItemLength];
```
bryphe, line 8: Unfortunately I found out naked-pointer support will be going away to support multi-core - so eventually we'll have to convert this to ocaml-managed objects: https://discuss.ocaml.org/t/ann-a-dynamic-checker-for-detecting-naked-pointers/5805

We might want to start using `caml_alloc_custom` to wrap these points - like in: https://github.com/onivim/reason-tree-sitter/blob/df6a110b990b556e16511e3c9be2820621548009/src/bindings.c#L71 - the nice thing is that you can set a custom finalize function, like here: https://github.com/onivim/reason-tree-sitter/blob/df6a110b990b556e16511e3c9be2820621548009/src/bindings.c#L38 - so that we could 'free' these when the ocaml runtime is done with them.

src/Native/Tray.rei
```diff
@@ -0,0 +1,16 @@
+type t;
+
+type title = [ | `Text(string) | `Image(string)];
+
+/**
+ * make
+ *
+ * Takes a [title] of either `Text(string) or `Image(string).
+ * Where `Image(string) takes an absolute path to a file.
```
bryphe, line 9: Thanks for the documentation here!

examples/Examples.re
lessp: I updated the API to instead only take an optional `~imagePath` with a unit argument, we could extend this later to take a title if we'd like? At least this way leaves the alternatives open!

b7a2043

src/Native/tray.c
lessp: Thanks, removed these for now. 🙂 

b7a2043

src/Native/tray.c
lessp: Hmm, I couldn't compile when trying that. Might not be necessary with the refactor, though!

b7a2043

src/Native/tray_cocoa.c
lessp: Ah! 

So, I think I managed to refactor to use `caml_alloc` with `Abstract_tag` instead of using naked pointers, unless naked pointers are completely disallowed which would add some more re-factoring, which would be fine. 👍

lessp:
Not sure how we'd like to separate things going forward without naked pointers. I guess we could keep a similar structure but we'd need to import the ocaml-ffi-helpers in `foo_cocoa.h` as well?

Currently src/Native/tray_cocoa.c is unused and I dumped all of that into `tray.c`

cc @zbaylin

cdaringe:
I can't wait to try this! Is it on pause indefinitely or just delayed?

lessp:
> I can't wait to try this! Is it on pause indefinitely or just delayed?

hey @cdaringe, currently (unfortunately) I have extremely limited free-time for OSS, so if anyone wants to take this further feel free. 🙂

lessp:
Merged with master and moved the examples to a separate file, I believe what's left here is:

- [ ] Being able to remove an item from the tray
- [ ] Vaguely remember @zbaylin creating helpers for allocating stuff, might want to refactor to use those

lessp:
Hey Zach, I believe this covers the first minimal surface area, albeit for macOS only:

> The API surface is fairly large, so we should start by thinking of a minimal subset - perhaps addressing the following scenarios:
> 
> Create a tray icon
> Destroy a tray icon

Very possible that I might have missed something!

src/Native/tray.c
```diff
@@ -0,0 +1,88 @@
...
+
+    NSString *nsTitle =
+        [NSString stringWithCString:title encoding:NSUTF8StringEncoding];
+
+    statusItem.button.image = NULL;
+    statusItem.button.title = nsTitle;
+
+    result = revery_wrapPointer(statusItem);
+
+    CAMLreturn(result);
```
zbaylin, line 66: Is it possible to make this part of `tray_cocoa.c` rather than in here? It makes it a lot cleaner if we dispatch each OS's functionality to their own functions.

src/Native/tray.c
```diff
@@ -0,0 +1,88 @@
...
+    result = caml_alloc(sizeof(NULL), Abstract_tag);
+#endif
+    CAMLreturn(result);
+}
+
+CAMLprim value revery_setTrayTitle(value trayHandle_v, value title_v) {
+    CAMLparam2(trayHandle_v, title_v);
+    CAMLlocal1(result);
+#ifdef USE_COCOA
+    NSStatusItem* statusItem = (NSStatusItem*)(void*)Field(trayHandle_v, 0);
```
zbaylin, line 54: Can you use `revery_unwrapPointer` here (or in `tray_cocoa.c` if you move it)? That way if we ever change how we represent pointers we don't have to go around changing stuff here.

src/Native/tray.c
```diff
@@ -0,0 +1,88 @@
...
+    CAMLreturn(result);
+#elif USE_WIN32
+    result = caml_alloc(sizeof(NULL), Abstract_tag);
+#else
+    result = caml_alloc(sizeof(NULL), Abstract_tag);
+#endif
+    CAMLreturn(result);
+}
+
+CAMLprim value revery_setTrayTitle(value trayHandle_v, value title_v) {
```
zbaylin, line 50: Minor nit: `trayHandle_v` -> `vTrayHandle` (same for `title_v`)

src/Native/Tray.rei
```diff
@@ -0,0 +1,38 @@
+type t;
```
zbaylin, line 1: Awesome, thanks for the documentation!

src/Native/tray.c
```diff
@@ -0,0 +1,88 @@
...
+        const char *imagePath = String_val(Some_val(imagePath_v));
+
+        NSImage *nsImage = revery_makeImageFromAbsolutePath_cocoa(imagePath);
+
+        statusItem.button.image = nsImage;
+
+        UNUSED(imagePath);
+    }
+
+    result = revery_wrapPointer(statusItem);
```
zbaylin, line 39: When returning nullable pointers like this, I usually allocate a `void *` on the stack and then wrap it at the end (using `revery_wrapPointer`). It looks a bit cleaner, IMO.
https://github.com/revery-ui/revery/blob/d737fcdaae1f292238e5ce0490d8b5ec50162b8a/src/Native/menu.c#L37-L53

zbaylin (commented):
Hey @lessp -- so happy to finally see this getting in! I had some minor comments (mostly just rebasing with new APIs like the `revery_wrapPointer`). I also wonder if it's possible to hook into the `NSMenu` stuff here too. Maybe that should be a separate PR, though.

src/Native/tray.c
```diff
@@ -0,0 +1,88 @@
...
+#endif
+
+#include "utilities.h"
+
+CAMLprim value revery_makeTrayHandle(value imagePath_v) {
+    CAMLparam1(imagePath_v);
+    CAMLlocal1(result);
+
+#ifdef USE_COCOA
+    NSStatusItem *statusItem = [NSStatusBar.systemStatusBar statusItemWithLength:NSVariableStatusItemLength];
```
zbaylin, line 27: One thing I forgot to mention is that we should garbage collect these when they're no longer in use. You can see how I did it with the NSMenu's in `Menu.re`, but you can use `NSObject.release` if you add a `retain` here.

src/Native/tray.c
lessp: Thanks! Believe I got all of them here: 240e452

src/Native/tray.c
lessp: Like that better! Updated all of them here, I hope: 240e452

src/Native/tray.c
lessp: Changed here: 240e452

src/Native/tray.c
lessp: Moved to `revery_setTrayTitle_cocoa(statusItem, title);` and created another for `revery_removeStatusItem_cocoa(statusItem);` in: 240e452

lessp:
> Hey @lessp -- so happy to finally see this getting in! I had some minor comments (mostly just rebasing with new APIs like the `revery_wrapPointer`). I also wonder if it's possible to hook into the `NSMenu` stuff here too. Maybe that should be a separate PR, though.

Thanks for the review Zach, feels like I left quite a bit for you there! 😄

> I also wonder if it's possible to hook into the `NSMenu` stuff here too. Maybe that should be a separate PR, though.

Yeah, that would have to be up for grabs, just thought I'd try to get this in!

src/Native/tray.c
lessp: Alright, might be able to have a look tomorrow! Guess I need to handle all of them? E.g. the `NSString *nsTitle = [NSString stringWithCString:titleText encoding:NSUTF8StringEncoding];` as well?

lessp: Added it to the `make` handler here: 2f2c862

src/Native/tray.c
```diff
@@ -0,0 +1,85 @@
...
+    [(NSObject *)statusItem retain];
+
+    if (vImagePath != Val_none) {
+        const char *imagePath = String_val(Some_val(vImagePath));
+
+        NSImage *nsImage = revery_makeImageFromAbsolutePath_cocoa(imagePath);
+
+        statusItem.button.image = nsImage;
+
+        UNUSED(imagePath);
```
zbaylin, line 37: Nit: is this still needed?

zbaylin (approved):
Thanks @lessp! I'm going to push up some changes to the unwrap/wrap semantics, but other than that it looks great!

src/Native/tray.c
lessp: Hmm, actually, I don't know 🤔 Most likely I followed another example, but looking at the context, I was under the impression that it is, but this is your forte!

```
    if (vImagePath != Val_none) {
        const char *imagePath = String_val(Some_val(vImagePath));

        NSImage *nsImage = revery_makeImageFromAbsolutePath_cocoa(imagePath);

        statusItem.button.image = nsImage;

        UNUSED(imagePath);
    }
```

PR #974, opened 2020-07-30
fix(shaping): Hang on non-printable ASCII characters

bryphe:
__Issue:__ When a string with a non-printable ASCII character (like `CR` - ascii 10 or `LF` - ascii code 13) was passed to be shaped, it would hit an infinite loop. 

__Defect:__ The shaping code would resolve that there was no glyph available, and then enter the fallback code path - `fallbackFor`. The problem, though, is in these cases, we'd return _the same font_ - and then we'd try and resolve the hole using the 'new' (really, the same) font - endlessly.

There also seemed to be an off-by-one issue in `resolveHole` that caused an additional glyph to be surfaced in this case.

__Fix:__ In the case where the fallback font is the same as the source font, resolving the hole isn't going to help, so detect that case and handle it the same as the error case.

__Test Impact:__
- Add cases for all ASCII characters to catch this
- Add a simple CJK fallback case to verify we're still handling fallback in the simple case correctly

### DIFF
--- a/src/Font/FontCache.re
+++ b/src/Font/FontCache.re
@@ -202,6 +202,7 @@ let matchCharacter = (fallbackCharacterCache, uchar, skiaFace) =>
 let generateShapes:
   (~features: list(Feature.t), t, string) => list(ShapeResult.shapeNode) =
   (~features, font, str) => {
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for reviewing @zbaylin !

PR #976, opened 2020-08-04
feat(native): system menus (macOS)

zbaylin:
⚠️ **Work in progress** ⚠️

This adds a bunch of APIs to `Revery.Native.Menu` for controlling menus. Currently only works on macOS, but hopefully the API I made is generalize-able to Win32/Gtk!

### DIFF
--- a/examples/MenuExample.re
+++ b/examples/MenuExample.re
@@ -22,9 +22,7 @@ module Example = {
                   Menu.Item.create(~title="Item #" ++ string_of_int(i));
                 let () =
                   item
-                  |> Menu.Item.setOnClick(item =>
-                       print_endline(Utilities.MacOS.nsObjectToString(item))
-                     );
+                  |> Menu.Item.setOnClick(_ => setLastClicked(_ => Some(i)));
                 item;
               },
             );
--- a/examples/MenuExample.re
+++ b/examples/MenuExample.re
@@ -21,7 +21,10 @@ module Example = {
                 let item =
                   Menu.Item.create(~title="Item #" ++ string_of_int(i));
                 let () =
-                  item |> Menu.Item.setOnClick(() => Window.minimize(window));
+                  item
+                  |> Menu.Item.setOnClick(item =>
+                       print_endline(Utilities.MacOS.nsObjectToString(item))
+                     );
                 item;
               },
             );
--- a/examples/MenuExample.re
+++ b/examples/MenuExample.re
@@ -1,0 +1,58 @@
+open Revery;
+open Revery.UI;
+open Revery.UI.Components;
+
+open Revery.Native;
+
+module Example = {
+  let menu = Menu.create(~title="Zach's Menu");
+
+  let%component make = (~window, ()) => {
+    let%hook (lastClicked, setLastClicked) = Hooks.state(None);
+
+    let%hook () =
+      Hooks.effect(
+        OnMount,
+        () => {
+          let menuItems =
+            Array.init(
+              10,
+              i => {
+                let item =
+                  Menu.Item.create(~title="Item #" ++ string_of_int(i));
+                let () =
+                  item |> Menu.Item.setOnClick(() => Window.minimize(window));
+                item;
+              },
+            );
+          let () =
+            menuItems |> Array.iter(item => {Menu.addItem(menu, item)});
+          None;
+        },
+      );
+
+    Gc.full_major();
+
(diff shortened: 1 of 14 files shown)


Et7f3:
Just to have same API: for windows https://github.com/revery-ui/revery/tree/Et7f3-native-windows-menu (it lack example, and small other thing)

bryphe:
Very exciting! Thanks for looking into this, @zbaylin . Native menus would be such a huge step forward. Between your work and @Et7f3 's exploration on Windows - seems like we could be close!

I tried to run the example and hit this error on OSX:
```
Moved: 559 x 389
SOURCE FILE: MenuExample.re
Switched to example: MenuExample.re
2020-08-04 10:35:03.622 Examples[87643:1169515] revery_setOnClickForMenuItem_cocoa: <NSMenuItem: 0x7f8f8ed10d70 Item #0>
2020-08-04 10:35:03.623 Examples[87643:1169515] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSDictionaryM setObject:forKeyedSubscript:]: key cannot be nil'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff2e5b99a9 __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff58cc1a17 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff2e5f89a4 -[CFPrefsConfigurationFileSource initWithConfigurationPropertyList:containingPreferences:] + 0
	3   CoreFoundation                      0x00007fff2e511019 -[__NSDictionaryM setObject:forKeyedSubscript:] + 921
	4   Examples                            0x000000010356eb53 revery_setOnClickForMenuItem + 83
	5   Examples                            0x0000000103149f24 camlMenuExample__fun_1710 + 68
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

examples/MenuExample.re
```diff
@@ -0,0 +1,61 @@
...
+                let () =
+                  item
+                  |> Menu.Item.setOnClick(item =>
+                       print_endline(Utilities.MacOS.nsObjectToString(item))
+                     );
+                item;
+              },
+            );
+          let () =
+            menuItems |> Array.iter(item => {Menu.addItem(menu, item)});
```
bryphe, line 32: I'd like to explore a more 'JSX' way to handle the menus - something like:
```
Menu.update(menu, <Label title={item1} onClick={() => print_endline ("Clicked") />)
```
It ends up being much simpler when we need to dynamically update the menu - we don't have to worry about knowing about the previous state of the menu, we can just let the reconciler handle it for us.

examples/MenuExample.re
```diff
@@ -0,0 +1,61 @@
...
+          let menuItems =
+            Array.init(
+              10,
+              i => {
+                let item =
+                  Menu.Item.create(~title="Item #" ++ string_of_int(i));
+                let () =
+                  item
+                  |> Menu.Item.setOnClick(item =>
+                       print_endline(Utilities.MacOS.nsObjectToString(item))
```
bryphe, line 26: We'll need to figure out a cross-platform abstraction for this

examples/MenuExample.re
bryphe: Have some more thoughts about this here: https://github.com/revery-ui/revery/issues/250

zbaylin: Definitely, this is just stage 1 of what I had in mind. The consumer shouldn't need to deal with this API directly, it will be managed behind the scenes by the reconciler!

bryphe: Awesome, sounds great!

examples/MenuExample.re
Et7f3: On windows the moment we add the menu is the moment where we bind the callback. And this could be imitated by the macOs API.

zbaylin: I added this for testing -- it seems like right now the way OCaml registers callbacks, if you log out `print_endline(i)`, it will always print `9`, which was the "last created" menu item. Not sure the best way to fix something like this. Obviously could lead to some pretty rough unintended behavior!

zbaylin: I created a discussion about this issue on the OCaml forum: https://discuss.ocaml.org/t/calling-impure-closures-from-c/6201 -- hopefully I can get some clarity!

PR #978, opened 2020-08-07
refactor(Linux): use Gtk's event loop on Linux

zbaylin:
This is more of a proposal. It works on Linux right now, but it won't work on macOS/Windows for linking errors (something I need to fix). 

Basically the reason to do this would be so that we can get all the niceties that Gtk provides to its widgets: things like Menu Bars, Native inputs, etc. Since Gtk does it's own processing in the event loop, unfortunately that means we have to port over our current event loop (literally just a while loop if you look in `reason-sdl2`) to the Gtk one on Linux.

If anyone has any better ideas, please let me know!


zbaylin:
I decided to pick up `ppx_optcomp` using @EduardoRFS's `config.h` to fix the symbol issues. This lets us use native features across platforms without having to have a "fake" implementation on the others.

@EduardoRFS I had to modify the config file a bit since `ppx_optcomp` doesn't support `ifdef`, just `if` and `ifndef` -- pretty simple change but just wanted to make sure I didn't miss anything.

src/Native/Revery_Gtk.c
```diff
@@ -0,0 +1,38 @@
...
+
+CAMLprim value revery_startGtkEventLoop(value camlCallback) {
+    CAMLparam1(camlCallback);
+    camlRenderLoop = malloc(sizeof(long));
+    *camlRenderLoop = camlCallback;
+
+    caml_register_generational_global_root(camlRenderLoop);
+
+    g_timeout_add(0, revery_invokeRenderLoop_gtk, NULL);
+    int status = g_application_run(G_APPLICATION(revery_gtkApp), 0, NULL);
```
glennsl, line 34: What will happen here when the timeout callback invokes `SDL_WaitTimeout(250)`? My guess is that it will block Gtk from processing events for 250 ms, which is probably not great.

zbaylin:
@glennsl totally agree, and I was actually going to close this PR and start again for the same reasons. When I first started looking into this, I couldn't seem to find a way to poll & flush like you were talking about. However, I recently found this gist that I was using to bind Gtk to our SDL/X11 window: https://gist.github.com/pastapojken/69f1e223fd926425f7ae770a95ec79a2

Towards the end the event loop does the exact same thing I think we want: poll (`gtk_events_pending()`) and, if there are any events, flush (`gtk_main_iteration()`). I still think using `ppx_optcomp` is the best way to get these Gtk functions into Revery without cross-platform compilation issues, but I also think this PR needs to be redone.

glennsl:
Yeah, that looks like exactly what need. Good call!

PR #979, opened 2020-08-08
chore(formatting): format dune files in `esy format`

zbaylin:
The inconsistent formatting of some of the dune files was slightly bugging me. I discovered `dune format-dune-file`, and thought it would be a good idea to add it to our formatting script.


Et7f3:
dune 2 already format dune file so might not be needed to roll our script.

zbaylin:
@Et7f3 very true, but didn't know what the plans were for upgrading to `dune` 2. This can be removed if/when we do that.

packages/reason-skia/src/wrapped/bindings/dune
```diff
@@ -1,16 +1,14 @@
...
  (public_name reason-skia.wrapped.bindings)
- (libraries
-    reason-skia.wrapped.types
-    reason-skia.wrapped.c
-    ctypes.stubs
-    ctypes
-    integers
-))
+ (libraries reason-skia.wrapped.types reason-skia.wrapped.c ctypes.stubs
+   ctypes integers))
```
glennsl, line 14: The lack of consistent formatting has been bugging me a bit as well, but I also find this formatting style of having list items, and in particular libraries, crammed as much as possible on a single line separated only by a single space to be significantly less readable than having each item on a separate line. Maybe there's some setting for this? I couldn't find one. Otherwise I'd honestly prefer to keep handling it manually.

zbaylin: Alright sounds good. I don't think there is a setting for it -- we could make our own s-expressions parser and formatter if we wanted to I guess!

glennsl: Bummer, but yeah S-expressions certainly aren't the most difficult format to parse and pretty-print. It could also still be a good idea to just run this once, and then fix the formatting manually in the diff afterwards. It's not like we edit these all that much anyway, other than to add or remove libraries, so they'll hopefully still remain decently well-formatted without much assistance.

PR #982, opened 2020-08-11
feat(native): add Gtk APIs for windows/widgets

zbaylin:
This should be a better version of #978. This uses Gtk's poll -> flush mechanism rather than using their event loop.

Unfortunately we can't make the Gtk window part of the window struct on Linux it seems, since the Window creation requires some initialization that isn't done by the point at which we create the window, leading to a segfault. In order to try to fix this I made a `Hashtbl` that maps Sdl windows to Gtk widgets so we don't end up with a bunch of duplicate Gtk widgets in memory.


src/Native/Gtk.re
```diff
@@ -0,0 +1,62 @@
...
+       external c_gtkWidgetGetPath: widget => string =
+         "revery_gtkWidgetGetPath";
+       external c_gtkWidgetSetOpacity: (widget, float) => unit =
+         "revery_gtkWidgetSetOpacity";
+       external c_gtkWidgetGetOpacity: widget => float =
+         "revery_gtkWidgetGetOpacity";
+     };
+
+let eventsPending = c_gtkEventsPending;
+let mainIteration = c_gtkMainIteration;
```
glennsl, line 41: Why are these rebound instead of being exposed directly as externals?

zbaylin: Sorry I've been writing C-bindings like this in a personal project -- I can change them since it doesn't match the style of the bindings we already have.

src/Native/gtk/ReveryGtk_Widget.c
```diff
@@ -0,0 +1,81 @@
...
+
+    GdkWindow *gWindow =
+        gdk_x11_window_foreign_new_for_display(gDisplay, xWindow);
+
+    GtkWidget *gWidget = gtk_widget_new(GTK_TYPE_WINDOW, NULL);
+    g_signal_connect(gWidget, "realize", G_CALLBACK(revery_gtkRealize), gWindow);
+    gtk_widget_set_has_window(gWidget, TRUE);
+    gtk_widget_realize(gWidget);
+
+    CAMLreturn((value)gWidget);
```
bryphe, line 34: Since naked pointers are being deprecated - I think we should try to avoid new introductions of naked-pointers if we can help it: https://github.com/ocaml/ocaml/pull/9610

We could use the `Abstract_tag` for the `gWidget` like @lessp used in https://github.com/revery-ui/revery/pull/972

zbaylin: Yeah I was debating if I should implement that here or if I should use raw pointers and migrate all of them in a separate PR. I guess that other PR is needed regardless so I might as well save the work here haha.

src/Native/Gtk.re
zbaylin: Fixed in 26b96bd!

src/Core/App.re
```diff
@@ -152,6 +155,21 @@ let handleKeymapChanged = () => {
...
 };
 
+%ifdef
+USE_GTK;
+
+let flushGtkEvents = () =>
+  if (Revery_Native.Gtk.eventsPending()) {
+    AppLog.debug("Flushing Gtk events");
+    Revery_Native.Gtk.mainIteration() |> (ignore: bool => unit);
+  };
```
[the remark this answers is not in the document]
zbaylin: Changed in b360114!

src/Native/gtk/ReveryGtk_Widget.c
zbaylin: In f7c033b I added `revery_wrapPointer` and `revery_extractPointer` which wraps pointers in `Abstract_tag`s. I also ported the Gtk functions to use them in 62fd8dd. Let me know what you think!

src/Native/gtk/ReveryGtk_Widget.c
```diff
@@ -0,0 +1,85 @@
...
+
+    GtkWidget *gWidget = gtk_widget_new(GTK_TYPE_WINDOW, NULL);
+    g_signal_connect(gWidget, "realize", G_CALLBACK(revery_gtkRealize), gWindow);
+    gtk_widget_set_has_window(gWidget, TRUE);
+    gtk_widget_realize(gWidget);
+
+    value camlWidget = revery_wrapPointer(gWidget);
+
+    CAMLreturn(camlWidget);
+}
```
glennsl, line 39: This function name is misleading as it doesn't just _get_ the Gtk Widget, but _creates_ it. And given this name it's hard to understand why it would be so expensive that a lookup table is needed (it's explained in the comment, but that is also somewhat confusing, or at least easy to overlook, given this name).

zbaylin: Great point! I think _this_ should be named `createGtkWidgetFromXWindow`, and the private binding on the Caml side should be the same, but the public function (the one that uses the hashmap) should be called get, since "create" is only accurate the first call -- any call after that should be a O(1) get from the hashmap. Let me know what you think!

glennsl: Yeah, that sounds good!

I'm also slightly worried that there's no cleanup of the lookup that, and that `Widget.ofWindow` doesn't indicate that there is a (small) accumulative cost to calling it for each window. I can imagine some apps will create lots of windows in the form of message boxes and such. It's certainly less likely that they'll then call `Widget.ofWindow` for each of those, but I've long since learnt to never underestimate users! I don't have a good suggestion for addressing this though, just bringing it up so it's at least not left unsaid.

zbaylin: Yeah I definitely share that concern -- If we used the Lru cache they would eventually "expire". I'll add custom finalization now so that we get that for free later.

zbaylin: @glennsl It should be fixed in 408456d -- let me know if you have any other concerns!

src/Native/Gtk.re
```diff
@@ -27,6 +27,7 @@ open {
        external c_createGtkWidgetFromXWindow:
          Sdl2.Window.nativeWindow => widget =
          "revery_createGtkWidgetFromXWindow";
+       external c_gtkWidgetDestroy: widget => unit = "revery_gtkWidgetDestory";
```
glennsl, line 36: Typo
```suggestion
       external c_gtkWidgetDestroy: widget => unit = "revery_gtkWidgetDestroy";
```

src/Native/Gtk.re
```diff
@@ -22,7 +28,7 @@ open {
           bunch of GtkWidgets for one window, so this table maps Sdl windows
           to GtkWidgets.
           */
-       let windowToWidgetTable = WindowTbl.create(1);
+       let windowWidgetCache = WindowWidgetCache.create(~initialSize=8, 64);
```
glennsl, line 31: What if you have more than 64 windows? Or, more likely, you create more than 64 temporary windows after the main window while continuing to use its associated widget?

bryphe:
Change looks good to me (pending resolving Glenn's feedback). 

Can you please test and verify Onivim 2 builds/runs with the latest changes?

src/Native/Gtk.re
zbaylin: Ah thanks! I'll fix that now.

zbaylin:
@bryphe sure! Will do.

src/Native/Gtk.re
zbaylin: That's a good point. I wonder if I can use the "depth" of the widget in the hierarchy as the weight in the cache. I'll look into that.

zbaylin: I used the depth of the widget in the hierarchy to determine the weight of the item in the cache. This means root windows wont have to be recreated while temporary child windows will be.

zbaylin:
@bryphe sorry for the delay, I've been busy moving into my new apartment. I just tested with Oni and it works, I just had to resolve `@opam/base` to `v0.14.0`.

src/Native/Gtk.re
glennsl: :+1: Seems like a good strategy to me!

src/Native/gtk/ReveryGtk_Widget.c
```diff
@@ -0,0 +1,109 @@
...
+    CAMLreturn(caml_copy_double(opacity));
+}
+
+CAMLprim value revery_gtkWidgetGetDepth(value vWidget) {
+    CAMLparam1(vWidget);
+
+    GtkWidget *gWidget = (GtkWidget *)revery_extractPointer(vWidget);
+
+    int result;
+    for (result = 0; gWidget != NULL; result++);
```
glennsl, line 102: I don't understand how this works. How is this not just an infinite loop?

glennsl: Also, `result` isn't a great variable name.

zbaylin: You were right, it was an infinite loop -- sorry about that! I fixed that and also changed `result` → `depth`

PR #983, opened 2020-08-11
feat(windows): ANGLE support

bryphe:
On Windows, Direct3D is much better supported than OpenGL - it's often the case where Direct3D is available, but OpenGL isn't (for example - on VMs). In addition, the driver support tends to be for Direct3D on Windows.

Google built a library called [ANGLE](https://github.com/google/angle) which provides an OpenGL ES API backed by Direct3D (It's used by Chrome as well to back WebGL).

This brings in ANGLE to be used as the renderer back-end for Revery and Skia. It requires bringing in `libEGL.dll` and `libGLESV2.dll`, and then setting up SDL to render to an OpenGL ES context.

Switching to ANGLE will likely fix several issues running Onivim 2 on Windows w/o up-to-date OpenGL drivers:
- https://github.com/onivim/oni2/issues/1522
- https://github.com/onivim/oni2/issues/1809
- https://github.com/onivim/oni2/issues/1842

(at least, enough to get the app visible & running - I can run Onivim 2 and the revery sample app on a Windows 10 VM w/ this change!)


Et7f3:
Skia already build angle on windows can't we copy this library ?

bryphe:
@Et7f3 - hmm, good point! Let me see if the dlls are available - we could move the build step to `esy-skia` instead of adding a new dependency.

bryphe:
@Et7f3 - I built `esy-skia` - (with and without `skia_use_angle`) - but didn't see the actual libEGL.dll or libGLESV2.dll produced in the build output. Do you know where they would be?

Et7f3:
I definitely see that it is created maybe we should need to tweak the ninja command https://github.com/google/skia/blob/52a4379f03f7cd4e1c67eb69a756abc5838a658f/third_party/angle2/BUILD.gn#L89

I would simply launch `dir /s/b | findstr dll` after launching esy.

bryphe:
Interesting, ya, I wasn't able to find it either:
```
D:\esy-skia>dir /s/b | findstr \.dll
D:\esy-skia\_esy\default\store\b\esy_skia-6e452493\third_party\externals\icu\windows\icudt.dll
D:\esy-skia\_esy\default\store\b\esy_skia-a137d801\out\Shared\skia.dll
D:\esy-skia\_esy\default\store\b\esy_skia-a137d801\third_party\externals\icu\windows\icudt.dll
```

It seems like we _should_ be able to get it from here - but I'll hold off for now (the pull-from-electron approach in [esy-angle-prebuilt](https://github.com/revery-ui/esy-angle-prebuilt) is certainly awkward... but it seems to work OK!) Open to improving this for sure, though. 👍

PR #984, opened 2020-08-12
feat(window): Add option for choosing acceleration strategy

bryphe:
This change adds an option when creating a window to determine the acceleration strategy:
- `Auto` - let the platform decide whether to create a hardware or software renderer. Usually this will default to hardware if available.
- `ForceHardware` - only allow hardware-accelerated renderers
- `ForceSoftware` - only allow software-accelerated renderers

The motivation for this change comes from investigating https://github.com/onivim/oni2/issues/2225 https://github.com/onivim/oni2/issues/1185 - it seems that, in some multiple-display cases, there may not be a hardware-accelerated renderer available. 

In that case, Onivim fails to start (`no pixel format available`) - because when it queries the available renderers, there is none matching a hardware-supported OGL context.

This can also allow Onivim to have a configuration setting - `--gpu-acceleration=hardware`/`--gpu-acceleration=software`/`--gpu-acceleration=auto` (default)

I believe this should fix #2225 and #1185 once brought into Onivim (with the caveat that the software renderer will be used, which may have perf implications - but the OpenGL software implementation on Apple actually seems pretty performant...)

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -329,6 +329,7 @@ let init = app => {
           ~icon=Some("revery-icon.png"),
           ~decorated=decorated^,
           ~forceScaleFactor=forceScaleFactor^,
+          ~acceleration=`Auto,
           (),
         ),
       app,
--- a/packages/reason-sdl2/src/sdl2.re
+++ b/packages/reason-sdl2/src/sdl2.re
@@ -159,10 +159,11 @@ module Window = {
       [ | `Undefined | `Centered | `Absolute(int)],
       [ | `Undefined | `Centered | `Absolute(int)],
       int,
-      int
+      int,
+      [ | `Auto | `ForceHardware | `ForceSoftware]
     ) =>
     t =
-    "resdl_SDL_CreateWindow";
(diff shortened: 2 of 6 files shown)


zbaylin:
Mostly unrelated to this PR, but would it ever be possible to use Metal rather than OpenGL on macOS? Considering OpenGL on macOS has been deprecated since 10.14 and Apple seemingly has no problem breaking compatibility with "old" software (re: x86 → ARM), it might not be too much longer until OpenGL is removed

bryphe:
> Mostly unrelated to this PR, but would it ever be possible to use Metal rather than OpenGL on macOS? 

@zbaylin - Yes, good question, there are two options:

1) ANGLE also has [planned/in-progress support for Metal](https://github.com/google/angle#platform-support-via-backing-renderers):
image

2) Skia also has a [Vulkan backend](https://wiki.libsdl.org/CategoryVulkan), and SDL supports creating a [Vulkan instance](https://wiki.libsdl.org/CategoryVulkan)

I think 2 - Vulkan backend - is the most likely route; and maybe can even be prototyped once we upgrade to latest Skia

zbaylin:
Awesome @bryphe -- that will be really nice to have!

PR #985, opened 2020-08-12
fix(windows): Harfbuzz build failure (remove CMake dep)

bryphe:


Et7f3:
for reduced feedback loop can't you test on harfbuzz repo ? and check also their is a change in environnement https://github.com/facebook/reason/commit/d3822b78b9a8c37594d3d5bff640f9159af0e298 this commit can help and https://github.com/revery-ui/esy-skia/commit/0b2fb9b40cc39b09a68ed00f18c8bc3a67d9d756 from http://redsymbol.net/articles/unofficial-bash-strict-mode/ could also help when debugging a remote environement

PR #986, opened 2020-08-13
fix(hooks): tick callback staleness

glennsl:
As reported by @lessp on Discord, this fixes an issue where the tick hook would always execute the initial callback instead of updating with the latest given to it. The fix is to store the callback in a ref, and call that instead of calling the closed over callback directly.

Technically this is a breaking change for code relying on the behavior that the callback goes stale. That seems highly unlikely though, at least intentionally.

### DIFF
--- a/src/UI_Hooks/Tick.re
+++ b/src/UI_Hooks/Tick.re
@@ -4,8 +4,8 @@ module Tick = Revery_Core.Tick;
 module Hooks = Revery_UI.React.Hooks;
 
 let tick = (~tickRate=Time.seconds(1), onTick) => {
-  // Because Tick.interval is only called once, intiiallly, with the initial
-  // onTick function, to execute the latest onTIck function we either have to
+  // Because Tick.interval is only called once, initiallly, with the initial
+  // onTick function, to execute the latest onTick function we either have to
   // dispose and recreate it for every call, or use a mutable variable to replace
   // the callback. Here we do the latter.
   let%hook onTickRef = Hooks.ref(onTick);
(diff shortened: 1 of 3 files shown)


src/UI_Hooks/Tick.re
```diff
@@ -1,16 +1,25 @@
 module Time = Revery_Core.Time;
+module Tick = Revery_Core.Tick;
 
 module Hooks = Revery_UI.React.Hooks;
 
 let tick = (~tickRate=Time.seconds(1), onTick) => {
+  // Because Tick.interval is only called once, intiiallly, with the initial
+  // onTick function, to execute the latest onTIck function we either have to
+  // dispose and recreate it for every call, or use a mutable variable to replace
+  // the callback. Here we do the latter.
```
bryphe, line 10: Minor typos: `intiiallly` -> `initially`, `onTIck` -> `onTick`

Thanks for the descriptive comment

test/UI/HooksTest.re
```diff
@@ -112,5 +113,31 @@ describe("Hooks", ({describe, _}) => {
...
+  describe("Tick", ({test, _}) => {
+    module Ticker = {
+      let%component make = (~f, ()) => {
+        let%hook () = Hooks.tick(~tickRate=Time.zero, f);
+
+        <View />;
+      };
+    };
+
+    test("Callback does not go stale", ({expect, _}) => {
```
bryphe, line 127: Thanks for adding a test case to cover this!

src/UI_Hooks/Tick.re
glennsl: Oops. Good catch! Fixed in https://github.com/revery-ui/revery/pull/986/commits/1f5deee1bff63cabc3be16f14ad88a10bed96935

PR #990, opened 2020-08-19
use static libSDL2.a on skia

EduardoRFS:
Currently `reason-sdl2` uses the static version while skia uses `-lSDL2` this makes the binary link to SDL2 both dynamically and statically. This is a bug on iOS


EduardoRFS:
@bryphe @glennsl I fixed this, if one of you can review it <3

bryphe:
Looks good to me! Thanks @EduardoRFS

bryphe (approved):

PR #1003, opened 2020-09-26
Updates to add special keys to mouse/filedrop events

fanantoxa:
* Adds data about: `ctrlKey`, `altKey`, `shiftKey`, `guiKey` to mouse events.
* Updates Keyboard data to have `guiKey`
* Fixes Scroll view to have horizontal scroll
* Update ScrollView to Handle horizontal scrool with Shift + MouseWheel.
* Fixed state when was no scroll bars but bounce was working

### DIFF
--- a/src/Core/Window.re
+++ b/src/Core/Window.re
@@ -403,6 +403,7 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
     let wheelEvent: Events.mouseWheelEvent = {
       deltaX: float(deltaX),
       deltaY: float(deltaY),
+      keymod: Sdl2.Keymod.getState(),
     };
     Event.dispatch(v.onMouseWheel, wheelEvent);
 
@@ -410,16 +411,26 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
     let mouseEvent: Events.mouseMoveEvent = {
       mouseX: float(x),
       mouseY: float(y),
+      keymod: Sdl2.Keymod.getState(),
     };
     Event.dispatch(v.onMouseMove, mouseEvent);
 
   | Sdl2.Event.MouseButtonUp(event) =>
-    Event.dispatch(v.onMouseUp, {button: MouseButton.convert(event.button)})
+    Event.dispatch(
+      v.onMouseUp,
+      {
+        button: MouseButton.convert(event.button),
+        keymod: Sdl2.Keymod.getState(),
+      },
+    )
 
   | Sdl2.Event.MouseButtonDown(event) =>
     Event.dispatch(
       v.onMouseDown,
-      {button: MouseButton.convert(event.button)},
+      {
+        button: MouseButton.convert(event.button),
+        keymod: Sdl2.Keymod.getState(),
+      },
     )
 
   | Sdl2.Event.KeyDown({keycode, keymod, scancode, repeat, _}) =>
@@ -492,7 +503,12 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
       v.dropState = None;
       Event.dispatch(
         v.onFileDropped,
-        {mouseX: float(x), mouseY: float(y), paths: List.rev(list)},
+        {
+          mouseX: float(x),
+          mouseY: float(y),
+          paths: List.rev(list),
+          keymod: Sdl2.Keymod.getState(),
+        },
       );
     }
   | Sdl2.Event.Quit => ()
--- a/src/UI/FileDrop.re
+++ b/src/UI/FileDrop.re
@@ -5,11 +5,22 @@ open NodeEvents;
 
 module Log = (val Log.withNamespace("Revery.UI.FileDropped"));
 
-let internalToExternalEvent = (evt: Events.internalFileDropEvents) =>
+let internalToExternalEvent = (evt: Events.internalFileDropEvents) => {
+  let keymod = Sdl2.Keymod.getState();
+
   switch (evt) {
   | InternalFileDropped(evt) =>
-    FileDropped({mouseX: evt.mouseX, mouseY: evt.mouseY, paths: evt.paths})
+    FileDropped({
+      mouseX: evt.mouseX,
+      mouseY: evt.mouseY,
+      paths: evt.paths,
+      ctrlKey: Key.Keymod.isControlDown(keymod),
+      altKey: Key.Keymod.isAltDown(keymod),
+      shiftKey: Key.Keymod.isShiftDown(keymod),
+      guiKey: Key.Keymod.isGuiDown(keymod),
+    })
   };
+};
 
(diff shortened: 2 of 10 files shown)


src/UI/FileDrop.re
```diff
@@ -5,11 +5,22 @@ open NodeEvents;
 
 module Log = (val Log.withNamespace("Revery.UI.FileDropped"));
 
-let internalToExternalEvent = (evt: Events.internalFileDropEvents) =>
+let internalToExternalEvent = (evt: Events.internalFileDropEvents) => {
+  let keymod = Sdl2.Keymod.getState();
```
bryphe, line 9: It would be preferable to add the `keymod` to the source, where we record the event:

https://github.com/revery-ui/revery/blob/0141f2b4d8a239e205ee5eedf79d40c2b565db90/src/UI/Ui.re#L79
https://github.com/revery-ui/revery/blob/0141f2b4d8a239e205ee5eedf79d40c2b565db90/src/UI/Ui.re#L88
https://github.com/revery-ui/revery/blob/0141f2b4d8a239e205ee5eedf79d40c2b565db90/src/UI/Ui.re#L117
https://github.com/revery-ui/revery/blob/0141f2b4d8a239e205ee5eedf79d40c2b565db90/src/UI/Ui.re#L127

And, add `keymod` to those internal events.

The reason is it's not obvious from the interface here - which takes an internal event model and converts to the external model - that it need to query the outside environment. The worst case would be if we for some reason held on to an internal event, and ran this function at a later time - we might get an unexpected value if the keymod state has changed in between when the filedrop event occurred, and when we actually ran `internalToExternalEvent`.

src/UI/Mouse.re
```diff
@@ -120,25 +120,86 @@ let getPositionFromMouseEvent = (c: Cursor.t, evt: Events.internalMouseEvents) =
   | InternalMouseOut(_) => Cursor.get(c)
   };
 
-let internalToExternalEvent = (c: Cursor.t, evt: Events.internalMouseEvents) =>
+let internalToExternalEvent = (c: Cursor.t, evt: Events.internalMouseEvents) => {
+  let keymod = Sdl2.Keymod.getState();
```
bryphe, line 124: Same as above - we should push this closer to the source of when we get the actual event, so that `internalToExternalEvent` can be a pure function of an internal event -> external event.

src/UI/NodeEvents.re
```diff
@@ -21,19 +21,31 @@ module Actions = {
 type mouseMoveEventParams = {
   mouseX: float,
   mouseY: float,
+  ctrlKey: bool,
+  altKey: bool,
+  shiftKey: bool,
+  guiKey: bool,
```
bryphe, line 27: Awesome! Thanks for adding these 💯

src/UI_Components/ScrollView.re
```diff
@@ -85,12 +131,30 @@ let%component make =
     switch (outerRef) {
     | Some(outer) =>
       let inner = outer#firstChild();
-      let childMeasurements = inner#measurements();
+
+      // TODO: #287 For some reasone `inner` component doesn't get expanded by it's childrens
+      // Could be but with https://github.com/jordwalke/flex or how we use it
```
bryphe, line 136: Minor typo: `reasone` -> `reason`

This is a bummer, I wish we knew why we weren't getting the full expanded width here.

src/UI_Components/ScrollView.re
```diff
@@ -85,12 +131,30 @@ let%component make =
...
+
+      let maxChildWidth =
```
bryphe, line 138: This workaround seems OK for now! Thanks for taking a look at this.

src/UI_Components/ScrollView.re
```diff
@@ -156,28 +215,34 @@ let%component make =
...
-          setBouncingState(_ => Idle)
-        | Bouncing(_) => ()
-        | Idle when !bounce && (isAtTop || isAtBottom) =>
-          let clampedScrollTop = isAtTop ? 0 : maxHeight;
-          dispatch(ScrollUpdated(clampedScrollTop));
-        | Idle when bounce && (isAtTop || isAtBottom) =>
-          setBouncingState(_ => Bouncing(- delta * 2));
-          dispatch(ScrollUpdated(isAtTop ? 0 : maxHeight));
-        | Idle => dispatch(ScrollUpdated(newScrollTop))
+        let horizontalScroll = wheelEvent.shiftKey || wheelEvent.deltaX > 0.;
```
bryphe, line 218: On OSX - it seems the `deltaX` is reversed - so the `wheelEvent.deltaX > 0.` check here and below might actually need to be `abs(wheelEvent.deltaX)` to check if scrolling is active.

src/UI_Components/ScrollView.re
```diff
@@ -156,28 +215,34 @@ let%component make =
...
+
+        if (horizontalScroll) {
+          if (isHorizontalScrollBarVisible) {
+            handeScroll(
+              ~deltaValue=
+                wheelEvent.deltaX > 0. ? wheelEvent.deltaX : wheelEvent.deltaY,
```
bryphe, line 224: It looks like on my OSX machine that this is reversed - so we'll need to check if it is Mac (`let horizontalScrollMultiplier = Environment.os == Mac ? -1 .: 1.`) and normalize this multiplier.

bryphe (changes requested):
Thanks for your help with this, @fanantoxa ! Really appreciate it

Overall approach looks great to me. I had some feedback about pushing the get-keymod-state closer to the source of the events - that will help us keep the `internalToExternalEvent` functions pure.

I also noticed one issue with the horizontal scrolling on mac - I can help test that piece.

One other note - the controls we use in Onivim are actually special (not based on this scrollable component, unfortunately) -  they use a virtualized scrollable component: https://github.com/onivim/oni2/blob/master/src/Components/FlatList.re

But this fix makes sense to make the `ScrollView` working here in Revery.

src/UI/FileDrop.re
fanantoxa: Done

fanantoxa:
@bryphe Updated with your comments!

> One other note - the controls we use in Onivim are actually special (not based on this scrollable component, unfortunately)
I know that, but I needed somehow to test and showcase that special keys actually pass through, ScrollView was the easiest one :smiley:

src/Core/Window.re
```diff
@@ -403,23 +403,34 @@ let handleEvent = (sdlEvent: Sdl2.Event.t, v: t) => {
     let wheelEvent: Events.mouseWheelEvent = {
       deltaX: float(deltaX),
       deltaY: float(deltaY),
+      keymod: Sdl2.Keymod.getState(),
```
bryphe, line 406: Perfect! Thanks for moving these 👍

bryphe (approved):
Thanks @fanantoxa for wiring up the modifier state the mouse events - looks great! 👍

PR #1017, opened 2020-11-15
fix(docs): update sidebar links

paulshen:
This updates the doc sidebar links so they all point a valid URL.

I noticed the [deployed docs](https://www.outrunlabs.com/revery/api/revery/index.html) seem to be a bit behind and some sidebar links point to 404s.

## Test Plan
`esy '@doc' run` and click all sidebar links


bryphe:
Thanks @paulshen ! Appreciate the fix :tada:

PR #1019, opened 2020-11-19
feat(macOS): add API to get titlebar size

zbaylin:
Since macOS Big Sur changes the size of the titlebar by default, it can lead to things looking kind of strange when you roll your own titlebar (like in Oni). This adds an API to get the size of the titlebar from Cocoa itself, so we don't have to make guesses from the version of macOS or something like that.

### DIFF
--- a/packages/reason-sdl2/src/sdl2_wrapper.cpp
+++ b/packages/reason-sdl2/src/sdl2_wrapper.cpp
@@ -353,6 +353,25 @@ extern "C" {
         CAMLreturn(Val_unit);
     }
 
+    CAMLprim value resdl_SDL_GetMacTitlebarHeight(value vWin) {
+        CAMLparam1(vWin);
+        double titlebarHeight = 0.0;
+
+#ifdef SDL_VIDEO_DRIVER_COCOA
+        SDL_Window *win = (SDL_Window *)vWin;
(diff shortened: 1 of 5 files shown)


packages/reason-sdl2/src/sdl2_wrapper.cpp
```diff
@@ -353,6 +353,25 @@ extern "C" {
...
+        SDL_Window *win = (SDL_Window *)vWin;
+        SDL_SysWMinfo wmInfo;
+        SDL_VERSION(&wmInfo.version);
+        SDL_GetWindowWMInfo(win, &wmInfo);
+        NSWindow *nWindow = wmInfo.info.cocoa.window;
+
+        // Sourced from: https://stackoverflow.com/a/59323932/12701512
+        CGFloat windowFrameHeight = CGRectGetHeight([nWindow contentView].frame);
+        CGFloat contentLayoutRectHeight = CGRectGetHeight([nWindow contentLayoutRect]);
+        titlebarHeight = (double)(windowFrameHeight - contentLayoutRectHeight);
```
bryphe, line 370: Thanks for finding this, @zbaylin !

bryphe (approved):

bryphe (approved):
Looks good to me @zbaylin - thanks for the fix!

PR #1020, opened 2020-11-19
fix(notifications): remove startup prompt on macOS

zbaylin:
This addresses onivim/oni2#2705. Before we were needlessly setting the notification delegate on start, which would prompt the user if they wanted to allow notifications. Now the prompt is shown on first attempt at showing notifications.


bryphe (approved):
Nice catch @zbaylin - that's great that the prompt will now only show on first notification. 

Will be a huge improvement in the first-run experience on OSX to not have this pop up!

bryphe:
Just tried w/ Onivim too - fixes https://github.com/onivim/oni2/issues/2705 :tada:

PR #992, opened 2020-08-25
refactor(native): use wrapped pointers

zbaylin:
This is currently breaking very early on -- I get a segfault with the Window in SDL2 and I can't seem to figure out why.

Here's a stack trace from LLDB:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x1400)
  * frame #0: 0x00000001000a2d2d Examples`SDL_GetWindowID_REAL(window=0x0000000000001400) at SDL_video.c:1719:5 [opt]
    frame #1: 0x0000000100b0a325 Examples`resdl_SDL_GetWindowId(vWindow=4432065872) at sdl2_wrapper.cpp:1539:18
    frame #2: 0x00000001002707f0 Examples`camlRevery_Core__Window__create_1627 + 768
    frame #3: 0x0000000100273656 Examples`camlRevery_Core__App__createWindow_inner_1628 + 22
    frame #4: 0x00000001000f02a1 Examples`camlExamples__init_1444 + 1361
    frame #5: 0x000000010027386c Examples`camlRevery_Core__App__start_951 + 284
    frame #6: 0x0000000100b599b8 Examples`caml_start_program + 72
```

If anybody has any ideas, let me know! I'll keep looking at it.

PR #1023, opened 2020-11-24
feat(native): wrap common NSObject functions in an OCaml context

zbaylin:
This creates a basic wrapper of various `NSObject` selectors, such as `isEqual`, `hash`, `description`, etc.

This will be helpful in the creation of native menus on macOS, since we'll have generic functions for comparing values, which will be useful in data structures like `Hashtbl.t`


src/Native/NSObject.re
```diff
@@ -0,0 +1,27 @@
...
+%ifdef
+USE_COCOA;
+
+type t('a) = 'a;
+
+let wrap: 'a => t('a) = nsObj => nsObj;
+
+/* equal
+   Uses the NSObject isEqual selector */
+external equal: (t('a), t('b)) => bool = "revery_NSObject_equal";
```
bryphe, line 13: Nice! Interface looks great.

src/Native/ReveryNSObject.c
```diff
@@ -0,0 +1,52 @@
...
+#import <objc/objc-runtime.h>
+
+#include <caml/alloc.h>
+#include <caml/callback.h>
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
+
+CAMLprim value revery_NSObject_equal(value vNSObjA, value vNSObjB) {
+    CAMLparam2(vNSObjA, vNSObjB);
+    NSObject *nsObjA = (NSObject *)vNSObjA;
```
bryphe, line 12: Should we wrap these pointers here? Just so we don't add another case we need to handle here: https://github.com/revery-ui/revery/pull/992

zbaylin: My plan was to get #992 merged into master then merge master into this. That PR includes fixes to the pointer wrap functions so I think it makes most sense to do it in that order.

bryphe: Makes sense!

bryphe (approved):
Looks great :+1:

PR #1024, opened 2020-11-28
feat(macOS): native menus

zbaylin:
This wraps the `NSMenu*` API in the FFI, and tries to give a generic API from the Caml-end.

Example code:
```reason
  let menubar = Native.Menu.getMenuBarHandle();
  let submenu = Native.Menu.create("Zach's Menu!");
  let subsubmenu = Native.Menu.create("Zach's Submenu!");
  let subsubmenuItem =
    Native.Menu.Item.create(~title="Zach's Submenu Item!", ~onClick=() => ());
  let subsubmenuItem2 =
    Native.Menu.Item.create(~title="Zach's Submenu Item?", ~onClick=() => ());
  let submenuItem =
    Native.Menu.Item.create(~title="Zach's Menu Item!", ~onClick=() =>
      Native.Menu.removeSubmenu(~parent=submenu, ~child=subsubmenu)
    );

  Native.Menu.addItem(submenu, submenuItem);
  Native.Menu.addItem(subsubmenu, subsubmenuItem);
  Native.Menu.insertItemAt(subsubmenu, subsubmenuItem2, 0);

  Native.Menu.addSubmenu(~parent=menubar, ~child=submenu);
  Native.Menu.insertSubmenuAt(~parent=submenu, ~child=subsubmenu, ~idx=0);
```


src/Native/utilities.c
```diff
@@ -33,3 +33,15 @@ value revery_wrapPointer(void *p) {
...
     return *((void **) Data_abstract_val(v));
 }
+
+/* Create an OCaml value encapsulating the pointer p
+    None when p == NULL, Some(p) when p != NULL
+*/
+value revery_wrapOptionalPointer(void *p) {
+    if (p != NULL) {
+        value v = revery_wrapPointer(p);
+        return Val_some(v);
```
bryphe, line 43: I believe we'll need to wrap with `CAMLparam0`, `CAMLlocal1(vSome)`, and `CAMLreturn(...)` - `Val_some` may allocate so we could get a crash if we're not tracking the value here.

src/Native/ReveryNSObject.c
```diff
@@ -50,4 +52,13 @@ CAMLprim value revery_NSObject_className(value vNSObj) {
...
 }
 
+CAMLprim value revery_NSObject_release(value vNSObj) {
+    CAMLparam1(vNSObj);
+
+    NSObject *nsObj = (NSObject *)revery_unwrapPointer(vNSObj);
+    [nsObj release];
+
+    CAMLreturn(Val_unit);
+}
```
bryphe, line 62: Thanks for thinking about the object lifecycle here!

Where do we actually call `release`? I was thinking it might follow the pattern we use for objects we allocate in `skia`, like: https://github.com/revery-ui/revery/blob/1d095616594289183b83ccc669ca62929caf33c0/packages/reason-skia/src/Skia.re#L119

Where, once we're done with the `NSObject` on the OCaml side - we let the finalizer call `release` for us, so we don't have to manually manage memory from OCaml/Reason.

src/Native/Menu.re
```diff
@@ -0,0 +1,85 @@
...
+    switch (callback) {
+    | Some(cb) => cb()
+    | None => ()
+    };
+  };
+
+  Callback.register("revery_callbackForMenuItem", callbackForMenuItem);
+
+  let create = (~title, ~onClick) => {
+    let menu = c_create(title);
```
bryphe, line 49: I guess this may be one place where we'd want to hook up a finalizer to `release` the menu item when we're done?

src/Native/Menu.rei
```diff
@@ -0,0 +1,25 @@
...
+let create: string => t;
+let toString: t => string;
+let nth: (t, int) => option(Item.t);
+let addItem: (t, Item.t) => unit;
+let insertItemAt: (t, Item.t, int) => unit;
+let removeItem: (t, Item.t) => unit;
+let addSubmenu: (~parent: t, ~child: t) => unit;
+let removeSubmenu: (~parent: t, ~child: t) => unit;
+let insertSubmenuAt: (~parent: t, ~child: t, ~idx: int) => unit;
+let clear: t => unit;
```
bryphe, line 25: This is really exciting to have this! :heart_eyes:

bryphe:
This is awesome, @zbaylin ! Will be a huge step forward for Revery / Onivim to have an application menu on OSX.

I had a few questions / comments around memory management - I'm always nervous about that since it can give us surprise crashes (or memory leaks).

bryphe:
Just tried this API out! It's looking great @zbaylin - nice work!

I added a Native menu example that lets me clear the menu bar and add a full set (like we'll do for Onivim when the menu state changes):



I think the only thing we're missing is a way to specify the `keyEquivalent` - right now, all the menu items are hard-coded to `Command+P`. A quick and easy way would be to just pass an optional/nullable string to the item creation - this would let us implement `Command+` and `Command+Shift+` shortcuts.

Longer term - we might want to add a `KeyEquivalent` module - which would let us add other marks (like the option key): https://stackoverflow.com/questions/6229848/cocoa-setting-the-key-equivalent. Under the hood, it seems like it's just a string with some helpers for appending special characters to represent option and other interesting keys (backspace, arrow keys, etc).

zbaylin:
@bryphe I just added a `KeyEquivalent` module that uses an API like almost exactly what you described. Let me know what you think!

examples/NativeMenuExample.re
```diff
@@ -0,0 +1,88 @@
...
+      let subMenu31 = Menu.create("Submenu 1");
+      Menu.addSubmenu(~parent=menu3, ~child=subMenu31);
+
+      let item311 =
+        Menu.Item.create(
+          ~title="Item 3.1.1",
+          ~onClick=menuCallback("Item 1.3"),
+          ~keyEquivalent=
+            Menu.KeyEquivalent.ofString("c") |> Menu.KeyEquivalent.enableAlt,
+          (),
```
bryphe, line 61: API looks great @zbaylin ! Thank you!

bryphe (approved):
Looks good to me - the new `KeyEquivalent` API is exactly what we need. And the functionality works great - can't wait to have it in Onivim 💯

PR #1029, opened 2020-12-23
feat(macOS): native buttons (NSButton)

zbaylin:
This does a couple main things:
- creates a wrapper for the `NSView` class similar to that created for `NSObject` in #1023
- creates a `Native.Input.Button` module that houses the basic functions to render a button
- wraps those functions in a standardized `Node` (similar to an `ImageNode`)
- wraps the `NativeButtonNode` in a `NativeButton` `nativeComponent`
- adds a `cleanup` function to the `Node` class
  - this is to deal with removing the `NSButtons` from the view once they are removed. Before, when a child was removed, `setParent(None)` was called on the child. However, if you have a basic view hierarchy like `Root -> Example -> [Buttons]` and the `Example` is removed, we have to notify all the children that they have been removed too.


zbaylin:
After further investigation, here are the things left that don't (and won't, on macOS at least) work:
- background color -- Cocoa explicitly disallows setting the background color of a "bordered" button, i.e. one that has any styling: https://developer.apple.com/documentation/appkit/nsbuttoncell/1529743-backgroundcolor?language=objc
- opacity -- buttons can either be fully opaque or fully transparent, no in-between: https://developer.apple.com/documentation/appkit/nsbuttoncell/1533667-opaque?language=objc
- height -- the height of the button itself cannot be changed, only the frame which it occupies. Again, this is a styling issue (I assume it's because they use bitmaps to construct the buttons and it only scales well horizontally)
- rotational transformations -- I *might* be able to get this to work but it would take some complex math. If anything, I think it would make sense to be in a separate PR.

@bryphe let me know if I need to change anything else, or if you have any questions about anything I have already!

src/Native/Input.rei
```diff
@@ -0,0 +1,27 @@
+module Button: {
```
bryphe, line 1: Thanks for coming up with a common, cross-platform API for this :+1:

src/Native/Input.rei
```diff
@@ -0,0 +1,27 @@
...
+  type t;
+
+  let hash: t => int;
+  let equal: (t, t) => bool;
+  let toString: t => string;
+
+  let create: (~title: string, ~onClick: unit => unit) => t;
+
+  let setX: (t, int) => unit;
+  let setY: (t, int) => unit;
```
bryphe, line 11: From exploring the windows implementation - I wonder if it makes sense to coalesce the setting x/y/width/height into a single call?

src/Native/NSView.re
```diff
@@ -0,0 +1,37 @@
...
+"config.h";
+
+%if
+defined(USE_COCOA);
+
+type t('a) = 'a;
+
+open {
+       external c_displayIn: (t('a), Sdl2.Window.nativeWindow) => unit =
+         "revery_NSView_displayIn";
```
bryphe, line 11: This is really exciting! Opens the door for integrating several native components:
- Native input control 
- Native HTML renderer

src/Native/NSView.re
```diff
@@ -0,0 +1,37 @@
+%import
+"config.h";
+
+%if
+defined(USE_COCOA);
+
+type t('a) = 'a;
```
bryphe, line 7: I wonder - what is the type parameter here for? Can we just remove it and use `t`?

src/UI/NativeButtonNode.re
```diff
@@ -0,0 +1,59 @@
...
+  val button = Button.create(~title, ~onClick);
+  inherit (class node)() as _super;
+  pub! draw = (parentContext: NodeDrawContext.t) => {
+    _super#draw(parentContext);
+    let dimensions = _this#measurements();
+    let worldTransform = _this#getWorldTransform();
+    Button.setX(button, int(Skia.Matrix.getTranslateX(worldTransform)));
+    Button.setY(button, int(Skia.Matrix.getTranslateY(worldTransform)));
+    Button.setWidth(button, dimensions.width);
+    Button.setHeight(button, dimensions.height);
```
bryphe, line 23: Nice job figuring out how to synchronize the revery-dimensions to the native button dimensions!

src/UI/NativeButtonNode.re
```diff
@@ -0,0 +1,59 @@
...
+    | Some(color) =>
+      let (red, green, blue, alpha) = Revery_Core.Color.toRgba(color);
+      Button.setColor(~red, ~green, ~blue, ~alpha, button);
+    | None => ()
+    };
+
+    _super#setStyle(adjustedStyle);
+  };
+  pub! cleanup = () => {
+    Button.remove(button);
```
bryphe, line 56: Glad you thought about this! Thanks for adding the `cleanup` method

bryphe (approved):
Just had some minor questions / feedback - but overall looks good! Thanks for exploring the Win32 API in the separate branch as well :+1: 

This is really exciting because it means potentially better native integration with input (ie, using or merging platform input with Revery text input) - would be a huge improvement over our manual emulation.

One concern I had was ensuring Onivim doesn't regress, but I tested with this change and didn't see any issues :+1: 

Thanks for all the work and investigation here @zbaylin !

src/Native/NSView.re
zbaylin: I do this so we can loosely model inheritance, i.e. a button can be wrapped as `NSView.t(NSButton.t)`.

PR #1030, opened 2020-12-24
feat(environment): macOS version

zbaylin:
This restructures the `Revery_Native.Environment.getOS` function and the `os` type to include the operating system version on macOS. The main change is that the `Mac` variant now contains three constructors: `Mac(major, minor, bugfix)`

I plan on extending this to other OS's as well (i.e. Linux could contain the kernel version (same format: major, minor, bugfix), Windows could use [`RtlGetVersion`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion?redirectedfrom=MSDN), etc.)

This is immediately to fix the problem with differing design constraints on different macOS versions in onivim/oni2#2876

### DIFF
--- a/examples/Examples.re
+++ b/examples/Examples.re
@@ -374,6 +374,12 @@ let init = app => {
       Console.log(Printf.sprintf("Moved: %d x %d", x, y))
     );
 
+  switch (Environment.os) {
+  | Mac(major, minor, bugfix) =>
+    Printf.printf("MacOS: %d.%d.%d\n", major, minor, bugfix)
+  | _ => ()
+  };
+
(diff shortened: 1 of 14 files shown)


bryphe (approved):
Awesome, thanks @zbaylin ! Great idea to bubble up the version directly instead of just using the titlebar size as a proxy.

I think there's some places where we check `Revery.Environment.os == Mac` in Onivim, so it might help to have some helper methods like `Revery.Environment.isMac` / `Revery.Environment.isWin` / `Revery.Environment.isLinux` to make it easy to switch those.

PR #1031, opened 2020-12-28
feat(environment): add kernel version info on Linux

zbaylin:
This extends #1030 to add the same sort of information on Linux. Now the constructor has four arguments: `Linux(kernel, major, minor, patch)`.

### DIFF
--- a/src/Core/Environment.re
+++ b/src/Core/Environment.re
@@ -22,8 +22,17 @@ type os =
     | IOS
     | Windows
     | Browser
-    | Mac(int, int, int)
-    | Linux(int, int, int, int);
+    | Mac({
+        major: int,
+        minor: int,
+        bugfix: int,
+      })
+    | Linux({
+        kernel: int,
+        major: int,
+        minor: int,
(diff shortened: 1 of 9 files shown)


bryphe (approved):
Awesome! Thanks @zbaylin

src/Core/Environment.re
```diff
@@ -20,10 +20,10 @@ type os =
     | Unknown
     | Android
     | IOS
-    | Linux
     | Windows
     | Browser
-    | Mac(int, int, int);
+    | Mac(int, int, int)
+    | Linux(int, int, int, int);
```
bryphe, line 26: One thought I just had - we might be able to name these paramters:
``` 
| Mac({ major: int, minor: int, bugfix: int })
| Linux({ kernel: int, major: int, minor: int, patch: int })
```

I believe that should have the same underlying FFI representation - just makes the intention of each `int` clearer

zbaylin: Great idea! I just changed it

PR #1032, opened 2020-12-29
feat(environment): add version information on Windows

zbaylin:
Extends both #1030 #1031 to include the version information on Windows. Now the constructor looks like `Windows({major: int, minor: int, build: int})`

### DIFF
--- a/src/Native/environment_windows.c
+++ b/src/Native/environment_windows.c
@@ -1,0 +1,33 @@
+#include "config.h"
+#ifdef IS_WINDOWS
+#include <windows.h>
+#include <stdio.h>
+
+void getOperatingSystemVersion_windows(int *major, int *minor, int *build) {
+    OSVERSIONINFOW osVersion;
+    ZeroMemory(&osVersion, sizeof(OSVERSIONINFOW));
+    osVersion.dwOSVersionInfoSize = sizeof(OSVERSIONINFOW);
+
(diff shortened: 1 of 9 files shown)


src/Native/environment_windows.c
```diff
@@ -0,0 +1,33 @@
...
+        if (rtlGetVersion != NULL) {
+            rtlGetVersion(&osVersion);
+        }
+
+        FreeLibrary(hNtDll);
+    }
+
+    if (rtlGetVersion == NULL) {
+        GetVersionEx((OSVERSIONINFO*)&osVersion);
+    }
```
bryphe, line 26: Just curious - is there a reason we can't always use `GetVersionEx`? Are there cases where it wouldn't return the correct info?

bryphe (commented):
Thanks for wiring this up across all our desktop OS's, @zbaylin !

Change overall looks good, I was just curious why we need both `RtlGetVersion` and `GetVersionEx`?

src/Native/environment_windows.c
zbaylin: Yeah, after Windows 8.1 `GetVersionEx` is deprecated and will only return Windows 8 information: https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa -- the kernel version still works, though!

bryphe: Ah, thanks for the info! Sounds good 👍

bryphe (approved):

PR #1036, opened 2020-12-31
feat(window): add function to set unsaved work

zbaylin:
This adds a function in a newly created `Revery_Native.Window` and in `Revery_Core.Window` to mark a window as having unsaved work. All this means as of now is that on macOS, the exit button has a little dot in it, i.e.:
image
This would be a nice little thing to add to Oni, I think 😃


bryphe (approved):
Very cool! This would indeed be awesome to have for Onivim. 

Thanks for thinking of this, @zbaylin ! Code change looks great.

PR #1042, opened 2021-01-09
feat(menu): add more menu item features

zbaylin:
I totally forgot this in the original menu PR, but @bryphe mentioned it in onivim/oni2#2969.

This just adds `Menu.Item.createSeparator`, which creates a separator item.

On macOS, it looks like this:
image

### DIFF
--- a/examples/NativeMenuExample.re
+++ b/examples/NativeMenuExample.re
@@ -70,6 +70,8 @@ module View = {
           (),
         );
 
+      Menu.Item.setEnabled(item312, false);
+
       Menu.addItem(subMenu31, item311);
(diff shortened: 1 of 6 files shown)


zbaylin:
I realized there are other things we could use (like enabling/disabling items), so I decided to broaden the scope of the PR

bryphe (approved):
Sweet, thanks for your help adding this @zbaylin !

Will definitely help https://github.com/onivim/oni2/pull/2969

bryphe:
I'll go ahead and merge this so I can pick it up in Onivim :sunglasses:

PR #1043, opened 2021-01-09
feat(audio): Add Wav File Playback to SDL

jozanza:
After seeing it mentioned in the roadmap, [this issue](https://github.com/revery-ui/revery/pull/555), and wanting it for my own personal Revery project, I added some WAV file playback support.

The interface exposed is a fairly straightforward mapping to a subset of SDL's Audio API. If we want to expose more of the SDL Audio API or would like to work on a higher-level Revery Audio API, I'd be happy to help out there.

Example Usage:
```re
open Revery.Sdl2.Audio;

/* 1. Load the WAV file spec and audio buffer */
let (desiredSpec, buf, len) =
  switch (Wav.load("audio.wav")) {
  | Ok(data) => data
  | Error(err) => failwith(err)
  };

/* 2. Open the audio device */
let (device, _obtainedSpec) =
  switch(Device.open_(None, false, desiredSpec, Device.AllowedChanges.none)) {
  | Ok(data) => data
  | Error(err) => failwith(err)
  };

/* 3. Play the audio buffer */
switch (queue(device, buf, len)) {
| Ok(_) => Device.pause(device, false)
| Error(err) => failwith(err)
};
```

I wasn't exactly sure what the best approach for testing and web support was, so I'd appreciate some direction.
Cheers! 🎉

### DIFF
--- a/examples/WavFilePlaybackExample.re
+++ b/examples/WavFilePlaybackExample.re
@@ -135,7 +135,7 @@ module WavFilePlaybackExamples = {
 
     let play = () => {
       switch (audioData, audioDevice) {
-      | (Some((spec, buf, len)), Some(device)) =>
+      | (Some((_spec, buf, len)), Some(device)) =>
         if (audioQueuedBytes > 0.) {
           Sdl2.Audio.Device.pause(device, false);
         } else {
--- a/examples/WavFilePlaybackExample.re
+++ b/examples/WavFilePlaybackExample.re
@@ -1,0 +1,238 @@
+open Revery;
+open Revery.UI;
+open Revery.UI.Components;
+open Revery.Native;
+
+module AudioHelpers = {
+  let getBitsPerSample =
+    fun
+    | Sdl2.Audio.Format.U8
+    | S8 => 8
+    | S16LSB
+    | S16MSB
+    | U16LSB
+    | U16MSB => 16
+    | S32LSB
+    | S32MSB
+    | F32LSB
+    | F32MSB => 32;
+
+  let getAudioDuration = (spec: Sdl2.Audio.Spec.t, bytes) => {
+    bytes / (spec.freq * spec.channels * (getBitsPerSample(spec.format) / 8));
+  };
+};
+
+module Styles = {
+  let container =
+    Style.[
+      position(`Absolute),
+      top(0),
+      bottom(0),
+      left(0),
+      right(0),
+      alignItems(`Center),
+      justifyContent(`Center),
+      flexDirection(`Column),
+    ];
+  let text = Style.[color(Colors.white), margin(14)];
+  let audioPlayerTrack =
+    Style.[
+      width(200),
+      marginBottom(14),
+      padding(2),
+      borderRadius(8.),
+      backgroundColor(Colors.dimGrey),
+      alignItems(`Center),
+      flexDirection(`Row),
+      justifyContent(`Center),
+    ];
+};
+
+module WavFilePlaybackExamples = {
+  let%component make = () => {
+    let%hook (filename, setFilename) = Hooks.state("");
+    let%hook (audioData, setAudioData) = Hooks.state(None);
+    let%hook (audioDevice, setAudioDevice) = Hooks.state(None);
+    let%hook (audioDeviceStatus, setAudioDeviceStatus) =
+      Hooks.state(Sdl2.Audio.Device.Status.Stopped);
+    let%hook (audioQueuedBytes, setAudioQueuedBytes) = Hooks.state(0);
+    let%hook () =
+      Hooks.tick(~name="", ~tickRate=Time.milliseconds(50), _ => {
+        switch (audioDevice) {
+        | Some(device) =>
+          let status = Sdl2.Audio.Device.getStatus(device);
+          setAudioDeviceStatus(_ => status);
+          let bytes = Sdl2.Audio.getQueuedSize(device);
+          setAudioQueuedBytes(_ => bytes);
+        | _ => ()
+        }
+      });
+
+    let audioQueuedBytes = float(audioQueuedBytes);
+    let audioTotalBytes =
+      switch (audioData) {
+      | Some((_, _, len)) => float(len)
+      | _ => 0.
+      };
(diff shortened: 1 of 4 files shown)


zbaylin (commented):
Looks great to me @jozanza! I only had one minor naming thing, but other than that it I think the code looks good.

Looks like you might need to run `esy format` on your code to pass CI, however.

Do you think you would be able to write an example to show off the functionality? I can help in the process if you want! You could combine the filepicker with the new functionality!

jozanza:
@zbaylin Sure! Do you want the example as part of this PR or another one after this is merged?

zbaylin:
@jozanza I think here, if you don't mind! Let me know if you need any assistance! (you seem to know what you're doing though haha)

bryphe:
Thanks for the awesome contribution, @jozanza - this looks great! Wanted an Audio API for a while.

packages/reason-sdl2/src/sdl2_wrapper.cpp
```diff
@@ -1252,6 +1253,83 @@ extern "C" {
...
+        SDL_AudioSpec spec = {0};
+        spec.freq = Int_val(Field(vSpec, 0));
+        spec.format = Int_val(Field(vSpec, 1));
+        spec.channels = Int_val(Field(vSpec, 2));
+        spec.silence = Int_val(Field(vSpec, 3));
+        spec.samples = Int_val(Field(vSpec, 4));
+        spec.padding = Int_val(Field(vSpec, 5));
+        spec.size = Int_val(Field(vSpec, 6));
+        SDL_AudioDeviceID dev = SDL_OpenAudioDevice(NULL, 0, &spec, NULL, 0);
+        CAMLreturn(Val_int(dev));
```
bryphe, line 1294: I believe that this could potentially return a negative value on error based on the documentation here: [SDL_OpenAudoDevice](https://wiki.libsdl.org/SDL_OpenAudioDevice), so we should wrap in an `option` like for other API.

packages/reason-sdl2/src/sdl2_wrapper.cpp
```diff
@@ -1252,6 +1253,83 @@ extern "C" {
...
+    CAMLprim value resdl_SDL_LoadWAV(value vPath) {
+        CAMLparam1(vPath);
+        CAMLlocal3(ret, vSpec, vTup3);
+        SDL_AudioSpec spec;
+        Uint8 *audioBuf;
+        Uint32 audioLen;
+        SDL_LoadWAV(String_val(vPath), &spec, &audioBuf, &audioLen);
+        if (audioBuf == NULL) {
+            ret = Val_error(caml_copy_string(SDL_GetError()));
+            CAMLreturn(ret);
```
bryphe, line 1265: Nice job with the C FFI / interop in these new APIs - it can be tricky to get the `CAMLparam`/`CAMLlocal`/`CAMLreturn` right... looks like you're a pro!

jozanza: Thanks. I like to mess around with ocaml/c bindings for fun sometimes :)

jozanza:
@bryphe while working on the example for wav playback, I wound up adding more SDL2 Audio API bindings. So this PR might need a second pass for review. @zbaylin the example is fairly basic, so if you've got any cool ideas, feel free to jump in and add/change whatever you want. It'd probably be good to at least test it out to make sure it also works properly for other people besides me 😄

bryphe:
Example looks great @jozanza - thanks for adding the progress bar / start / end times - very nice 😎 

Too bad the gif doesn't have sound... but it works well:

packages/reason-sdl2/src/sdl2_wrapper.cpp
```diff
@@ -1252,6 +1253,194 @@ extern "C" {
...
+    }
+
+    CAMLprim value resdl_SDL_CloseAudioDevice(value vDevice) {
+        CAMLparam1(vDevice);
+        SDL_AudioDeviceID device = Int_val(vDevice);
+        SDL_CloseAudioDevice(device);
+        CAMLreturn(Val_unit);
+    }
+
+    CAMLprim value resdl_SDL_GetAudioDeviceStatus(value vDevice) {
```
bryphe, line 1401: Thanks for the additional APIs you added - they look good to me 👍

jozanza: No prob!

examples/WavFilePlaybackExample.re
```diff
@@ -0,0 +1,238 @@
...
+module WavFilePlaybackExamples = {
+  let%component make = () => {
+    let%hook (filename, setFilename) = Hooks.state("");
+    let%hook (audioData, setAudioData) = Hooks.state(None);
+    let%hook (audioDevice, setAudioDevice) = Hooks.state(None);
+    let%hook (audioDeviceStatus, setAudioDeviceStatus) =
+      Hooks.state(Sdl2.Audio.Device.Status.Stopped);
+    let%hook (audioQueuedBytes, setAudioQueuedBytes) = Hooks.state(0);
+    let%hook () =
+      Hooks.tick(~name="", ~tickRate=Time.milliseconds(50), _ => {
```
bryphe, line 60: One issue I noticed - even when not playing, the hook will be active (which will cause continual re-render / CPU usage). Not a big deal for a game of course, since it'll be rendering every frame anyway - but something to watch out for with an app. 

One way to address this might be to set the `enabled` parameter based on the `audioDeviceStatus` (once it flips back to `Stopped`, we no longer run the timer, until it is playing again)

I'm OK with it as-is for the example, though - just something I thought about while I was playing with the example.

bryphe (approved):
Latest changes look great to me! Time to add some 'beeps' to Onivim 😄

zbaylin:
The example is *awesome*! Thanks @jozanza! Looks like @bryphe approved it so I can merge this in!

examples/WavFilePlaybackExample.re
jozanza: Ah yeah good point! No reason to poll at 50ms for progress when nothing's happening 🤪

PR #1051, opened 2021-03-05
Allow '`Undefined' WindowCreateOption for initial window position

Timbus:
Passing the `Undefined` variant to SDL allows the OS to choose the window position, which seems preferable to picking a location (`Absolute`) or spawning right in the middle of the desktop, across my two monitors (`Centered`).
Works fine from my testing.

Cool project btw.


glennsl (approved):
LGTM. Thanks!

PR #1053, opened 2021-03-20
onBeforeQuit: allow subscribers to prevent immediate quit

timbertson:
Implements the API change discussed in https://github.com/onivim/oni2/issues/3270

I decided to make a `Fanout` submodule of Event to represent "events with return values", but am very open to alternative structures. It seemed unnecessary to add another type paramater for `Event.t` when it's almost always going to be `unit`.

Since this changes the onBeforeQuit signature anyway, I included the exit code since it seems like that might be useful, but I don't have a specific use case in mind.

I didn't make it backwards compatible. It would be possible by instead adding a new `onBeforeQuitFanout` (or something) function that exposes the ability to suppress the response, but I don't know what your plans are in general for backwards compat, so I went with the simplest version first.

### DIFF
--- a/src/Core/App.re
+++ b/src/Core/App.re
@@ -14,9 +14,9 @@ type quitResponse =
   | PreventQuit;
 
 let mergeQuitResponse = (a: quitResponse, b: quitResponse) => {
-  switch (a,b) {
-    | (AllowQuit, AllowQuit) => AllowQuit
-    | _ => PreventQuit
+  switch (a, b) {
+  | (AllowQuit, AllowQuit) => AllowQuit
+  | _ => PreventQuit
   };
 };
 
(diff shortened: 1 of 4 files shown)


src/Core/App.re
```diff
@@ -9,13 +9,24 @@ module Log = AppLog;
 type delegatedFunc = unit => unit;
 type unsubscribe = unit => unit;
 
+type quitResponse =
+  | AllowQuit
+  | PreventQuit;
```
bryphe, line 14: Thanks for adding this 👍  This looks very clear to me

src/Core/Event.re
```diff
@@ -1,17 +1,34 @@
 /* Event.re */
 
+module Fanout = {
+  /* An event where subsccribers can respond with a value */
```
bryphe, line 4: Nice! I like the way you thought through this API, and the way you handle multiple subscribers with the `merge` function.

I'm also glad you found a name for it - I wasn't sure the best way to describe this, and `Fanout` sounds good to me 👍

bryphe:
> Since this changes the onBeforeQuit signature anyway, I included the exit code since it seems like that might be useful, but I don't have a specific use case in mind.

Sounds reasonable to me. I could potentially see cases where, whether the exit code is success or failure, we may want to handle it differently.

> It would be possible by instead adding a new onBeforeQuitFanout (or something) function that exposes the ability to suppress the response, but I don't know what your plans are in general for backwards compat, so I went with the simplest version first.

For this API, I'm OK breaking backwards compat - AFAIK, Onivim is the only consumer of this API. So the simpler approach sounds good - thank you for thinking about this and checking @timbertson !

bryphe (approved):
Looks like just what we need to fix https://github.com/onivim/oni2/issues/3270 - will merge in once the build is green. 

Thanks @timbertson for thinking through and implementing this API!

timbertson:
yay! 🥳 

I've raised a proper PR in Oni2: https://github.com/onivim/oni2/pull/3326

bryphe:
> I've raised a proper PR in Oni2: onivim/oni2#3326

Nice! Thank you @timbertson  - merging now 😄

PR #1058, opened 2021-04-17
feat(skia): bind to Skia's SVGDOM

zbaylin:
This uses Skia's SVGDOM API to render SVGs onto Skia canvas's, which is exactly what we need for SVG rendering!

The basic process to get it working is:
- Create an `SkStream` of some sort from the SVG you want to render
  - You can either construct an `SkMemoryStream` from an SVG string in memory, or an `SkFileStream` from an SVG file
- Create an SVG with the stream
- Render it onto a canvas

The results are surprisingly good, considering the SVG API is marked as "experimental" in Skia:
image


packages/reason-skia/src/wrapped/bindings/SkiaWrappedBindings.re
```diff
@@ -1121,4 +1143,27 @@ module M = (F: FOREIGN) => {
     let getProps =
       foreign("sk_surface_get_props", t @-> returning(SurfaceProps.t));
   };
+
+  module SVG = {
+    type t = ptr(structure(SkiaTypes.SVG.t));
+    let t = ptr(SkiaTypes.SVG.t);
+
+    let makeFromStream =
+      foreign("sk_svgdom_create_from_stream", Stream.t @-> returning(t));
```
bryphe, line 1152: We'll need the corresponding `unref` call available, so that we can clean this up when it is no longer referenced, in order to avoid a memory leak

packages/reason-skia/src/wrapped/bindings/SkiaWrappedBindings.re
```diff
@@ -1121,4 +1143,27 @@ module M = (F: FOREIGN) => {
...
+
+    let render =
+      foreign("sk_svgdom_render", t @-> Canvas.t @-> returning(void));
```
bryphe, line 1155: :tada:

packages/reason-skia/src/Skia.re
```diff
@@ -657,16 +657,24 @@ module ColorSpace = {
...
 
   let hasLength = SkiaWrapped.Stream.hasLength;
   let getLength = SkiaWrapped.Stream.getLength;
+
+  let makeFileStream = SkiaWrapped.Stream.makeFileStream;
+
+  let makeMemoryStreamWithData = (str, length) =>
+    SkiaWrapped.Stream.makeMemoryStreamWithData(str, length, true);
+
+  let makeFromData = SkiaWrapped.Stream.makeFromData;
```
bryphe, line 673: For these, we'll need to make sure we have the corresponding `delete` call when they are no longer referenced by the GC. Similar pattern to things like:
```
  let make = () => {
    let paint = SkiaWrapped.Paint.allocate();
    Gc.finalise(SkiaWrapped.Paint.delete, paint);
    paint;
  };
```
where we hook up a GC finaliser to delete/unref the allocated item when it is no longer used

packages/reason-skia/src/Skia.rei
```diff
@@ -544,3 +546,13 @@ module Surface: {
   let getHeight: t => int;
   let getProps: t => SurfaceProps.t;
 };
+
+module SVG: {
+  type t;
+
+  let makeFromStream: Stream.t => t;
```
bryphe, line 553: I believe this should be `Stream.t => option(t)`, as I would imagine an invalid SVG stream would return null?

packages/reason-skia/src/Skia.re
zbaylin: I actually tried that and had to remove it because I was getting segfaults on GC runs. I looked into it and my guess is that the unref call is trying to free memory from the stream that has already been consumed by the `SVGDOM` creation call. I'll see if I can get it to work again!

packages/reason-skia/src/Skia.rei
zbaylin: That's a good point. I haven't played around with invalid SVG streams, but IIRC there's some C++ exception handling in the Skia code so we might have to just return NULL on error.

packages/reason-skia/src/Skia.re
bryphe: Ah, interesting! This is a good point - if the SVGDOM relies on the stream sticking around, we might need to do something special, like create a wrapper type that holds a reference to both the SkSVGDOM instance _and_ the stream used to create it.

zbaylin: Resolved with revery-ui/esy-skia#31 :)

packages/reason-skia/src/Skia.re
```diff
@@ -657,16 +657,35 @@ module ColorSpace = {
...
+  };
+
+  let makeMemoryStreamWithData = (str, length) => {
+    let stream =
+      SkiaWrapped.Stream.makeMemoryStreamWithData(str, length, true);
+    Gc.finalise(SkiaWrapped.Stream.deleteMemoryStream, stream);
+    stream;
+  };
+
+  let makeFromData = SkiaWrapped.Stream.makeFromData;
```
bryphe, line 684: Does this need a `Gc.finalise` as well?

bryphe (approved):
Thanks for all the work on this, @zbaylin !

Just had one minor piece of feedback (I believe there is one more `Gc.finalise` needed) - otherwise looks great!

packages/reason-skia/src/Skia.re
zbaylin: Resolved in 2979572!

PR #1060, opened 2021-04-27
experiment: Update runtime SVG example

bryphe:
Experiment to update the SVG example to use the new `SkSVGDOM` API that @zbaylin implemented in #1058 

Unfortunately - it's not quite working to render the SVG. 

If the SVG is invalid, it crashes (likely due to a `null` return we don't handle in the FFI at the moment), and if it is valid, it loads and returns a width/height, but is empty:

image

Maybe something obvious I'm missing?

Also - we'll be able to completely remove our manual implementation of the `Revery_SVG` module here: https://github.com/revery-ui/revery/tree/master/src/SVG with #1058


bryphe:
Updated - working now with the layout of the canvas fixed

zbaylin:
This should be made obsolete by #1061 (I think)

bryphe:
For sure - thanks @zbaylin !

PR #1064, opened 2021-05-10
fix(build): shell.c depends utilities.h on mobile

EduardoRFS:
This is an error that appears if you don't have `GTK`, `COCOA` or `WIN32` available. Like if you don't have gtk-dev installed or you're on mobile like Android and iOS


zbaylin (approved):
Thanks @EduardoRFS!

PR #1065, opened 2021-05-10
fix(android): build of reason-harfbuzz on Android

EduardoRFS:
This fix reason-harfbuzz build when doing Linux -> Android / iOS, this allows revery to cross compile with less patches, almost zero.

Same setup of the other files, we should definitely abstract this, but I'm lazy.

### DIFF
--- a/packages/reason-harfbuzz/src/config/discover.re
+++ b/packages/reason-harfbuzz/src/config/discover.re
@@ -1,52 +1,80 @@
+module Configurator = Configurator.V1;
+
 type os =
-  | Windows
-  | Mac
+  | Android
+  | IOS
   | Linux
-  | Unknown;
+  | Mac
+  | Windows;
 
-let uname = () => {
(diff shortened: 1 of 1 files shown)


packages/reason-harfbuzz/src/config/discover.re
```diff
@@ -1,52 +1,80 @@
...
+  | Mac
+  | Windows;
 
-let uname = () => {
-  let ic = Unix.open_process_in("uname");
-  let uname = input_line(ic);
-  let () = close_in(ic);
-  uname;
-};
+let detect_system_header = {|
```
zbaylin, line 10: I wonder if we can make this a library or something so we don't have to keep repeating it -- it would probably be useful in Onivim as well. Outside of the scope of this PR though!

zbaylin (approved):
Looks good to me! CI will be the judge.

PR #1071, opened 2021-05-20
feat(native): add `Input.openEmojiPanel` on macOS

zbaylin:
One thing I've missed using Oni as my daily driver is the Emoji picker that is usually triggered by <kbd>Edit</kbd> -> <kbd>Emoji & Symbols</kbd>. This binds to that in `Revery_Native.Input`.

https://user-images.githubusercontent.com/4527949/119004996-0b713e00-b95d-11eb-8535-5f4be97e505f.mov

### DIFF
--- a/examples/InputExample.re
+++ b/examples/InputExample.re
@@ -70,6 +70,13 @@ module Example = {
           title="Set value"
           onClick={() => setValue(state => {...state, first: "New value"})}
         />
+        <Button
+          height=50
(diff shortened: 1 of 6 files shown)


bryphe (approved):
Neat, thanks for hooking this up @zbaylin ! Good way to showcase the emoji font fallback support too 😄 

Changes look good to me!

PR #1077, opened 2021-05-26
feat(sdl2): bring back continuous resize patch

zbaylin:
This seems to consistently work now (after #1072), and here is why I think that's the case:

Before, we would release the runtime lock before flushing events. These events would include things like mouse clicks, keyboard input, **window resizes**, etc. This meant that in the old patch, *most* resizes would require us to acquire the runtime lock since they were being "committed" in an event flush. However, last time we bumped up against a problematic resize: *synchronous* resizes, like `Window.maximize`. In this case, we already had the runtime lock when the event handler is called, so trying to acquire it again caused a deadlock.

*Now* we no longer release the runtime lock on event flushes 🎉. This means that in every case that we resize the window, the runtime lock is already acquired, so we don't have to acquire it.

@bryphe please let me know if you can think of another scenario where resizes would be triggered without the runtime lock, or if any of my reasoning sounds funny/wrong.

### DIFF
--- a/packages/reason-sdl2/src/sdl2.re
+++ b/packages/reason-sdl2/src/sdl2.re
@@ -905,6 +905,7 @@ let _nativeLoop = renderFn => {
 };
 
 let renderLoop = (renderFunction: renderFunction) => {
+  Callback.register("__sdl2_caml_resize__", renderFunction);
   switch (Sys.backend_type) {
   | Native => _nativeLoop(renderFunction)
   | Bytecode => _nativeLoop(renderFunction)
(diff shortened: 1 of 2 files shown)


bryphe:
Good idea to revisit this @zbaylin - 100% agree with your assessment, it should be safe now w/o the release/acquire dance around the PollEvent/WaitEvent.

Testing it (resizing the window, maximizing, etc) - seemed to work great!

I noticed one bug, though - the 'layer' in the canvas example seems to get stuck once the window is dragged below a minimum threshold:


I wasn't able to reproduce this w/o the change here - although I'm guessing this is a bug with the layer/canvas and not the resize, but is being exacerbated by the higher frequency rendering.

bryphe:
To follow up on my issue - I realized this isn't due to this change. In fact, it reproduces on Linux too which does smooth resizing even without this patch. So we shouldn't block this on that layer bug - seems like a separate issue.

Thanks for bringing this back @zbaylin - having the continuous resize is a big fit and finish improvement on OSX!

bryphe (approved):

PR #1078, opened 2021-05-27
feat(menu): add `Native.Menu.displayIn`

zbaylin:
#1072 allows us to do this safely since we never have to acquire the runtime lock 😀 

https://user-images.githubusercontent.com/4527949/119751161-a17afc00-be68-11eb-8738-6ad33858efeb.mov


bryphe (approved):
Awesome! Thanks for adding this @zbaylin 💯

PR #1079, opened 2021-05-28
feat(menu): better key equivalent API

zbaylin:
This adds a <kbd>CTRL</kbd> modifier and conversion of special keys to the right strings.


bryphe (approved):
Nice - thanks for investigating this 👍

PR #1086, opened 2021-06-19
chore(deps): upgrade Skia to 1c81aac

zbaylin:
This brings in the SVG rendering fix in different locales


bryphe (approved):
Thanks @zbaylin !

PR #1090, opened 2021-08-27
Delete .discover.re.swp

Et7f3:
I think this is the only swap file that exist.


zbaylin (approved):
Thanks for cleaning this up @Et7f3!

